Issue #2, opened 2019-02-16
Opt_einsum-based optimizer cannot handle Normal variable elimination


fritzo:
One thing blocking the use of the opt_einsum-based optimizer for continuous variable elimination is that `Normal` currently implements only a limited contraction interface:
- `.reduce(ops.logaddexp, self.value)` which just returns `0 = log(1)` because it's normalized, and
- `.binary(ops.add, x)` where x is ground wrt the current contraction.

As a workaround we're using a `tree_engine` that performs Birch-style variable elimination on trees, starting from observations (leaves) and moving backwards. While this will work to demo the Kalman filter example, it is not a long-term solution.

## Possible fixes
- try to support more general `x` in `Normal.binary()`
- support direction hinting in the opt_einsum optimizer

eb8680:
Thinking about this, our use case is different enough that it's probably easier to implement our own optimizer directly to begin with rather than attempting to use `opt_einsum.paths.greedy` or reuse some of their code.

eb8680:
Mostly addressed by #37

PR #13, opened 2019-02-17
Refactor constructor delegation order

### LINKED ISSUE #7: Move methods .unary, .binary, etc. into a handler
Currently weak simplification logic is baked into constructor methods of various Funsor types, e.g. `.binary()`, `.__call__()`. We'll want to move these to a handler, either `Eager` or `Simplify`.

fritzo:
Resolves #7 

This PR changes our delegation order from
```
x.__add__(y)  -->  x.binary(y)  -->  Binary(x,y)
```
to
```
x.__add__(y)  -->  Binary(x,y)  -->  _eager_add(x, y)  # under Eager handler
```
and moves code from `.binary()` into `._eager_add()`.

This PR also declares `Eager` and `Simplify` handlers in a new `funsor.interpretations`.

## Tasks
- [x] `Substitution`: rename `.__call__()` to `._eager_subs()` method
- [x] `Unary`: rename `.unary()` to `._eager_unary()` method
- [x] `Binary`: rename `.binary()` to `_eager_binary()` function
- [x] move unit laws from `Eager` to `Simplify` handler.
- [x] `Reduction`: rename `.reduce()` to `._eager_reduce()` method
- [ ] update tests to use appropriate handlers


fritzo:
@eb8680 This is what we discussed Friday. Can you review what I've done so far to make sure it's sane? I've only converted `.__call__()`, and others will be similar. I'll use `Eager` and keep simplification logic relegated to an optional `Simplify()` that does very little then delegates to `Eager`.

I think by combining this with a slightly more general `_eager_binary(Normal, Normal)` we might be able to switch to your opt_einsum optimizer #2 .

funsor/distributions.py
```diff
@@ -113,11 +114,14 @@ def __repr__(self):
         return '{}({})'.format(type(self).__name__,
                                ', '.join('{}={}'.format(*kv) for kv in self.params.items()))
 
-    def __call__(self, *args, **kwargs):
-        kwargs = {d: to_funsor(v) for d, v in kwargs.items() if d in self.dims}
-        kwargs.update(zip(self.dims, map(to_funsor, args)))
-        if not kwargs:
-            return self
+    def _eager_subs(self, **kwargs):
+        # DEBUG
```
fritzo: @eb8680 I'm having some trouble using the `@Eager.register(Substitution)` decorator. I've added some print statements and the following code fails with:
```
$ pytest -vx test/ -k normal_density --pdb
...
DEBUG Funsor.__call__ Normal(loc=-2.25884914398, scale=0.941187381744, value=value) (Tensor((), 1.13507556915),) {}
DEBUG _eager_subs Normal(loc=-2.25884914398, scale=0.941187381744, value=value) (('value', Tensor((), 1.13507556915)),)
DEBUG Distribution._eager_subs Normal(loc=-2.25884914398, scale=0.941187381744, value=value) {'value': Tensor((), 1.13507556915)}
DEBUG Funsor.__call__ value () {'value': Tensor((), 1.13507556915)}
FunsorMeta._ast_call <class 'funsor.terms.Substitution'> (Variable('value', 'real'), (('value', Tensor((), 1.13507556915)),))
```
I don't understand why `Variable._eager_subs()` is not being called.

eb8680: I modified `Handler` so that, by default, effectful operations that happen inside a `Handler.process` are considered out of scope for that handler and are processed only by the handlers outside it on the stack.  In this case `var(**kwargs)` calls `Substitution`, which is considered out of scope by `Eager` because it happens inside `Eager.process`.  Here you would need to call `Variable._eager_subs` directly within `Distribution._eager_subs`.

The motivation for this experimental change was so that effects could be fused - for example, in `opteinsum_engine` the optimized interpreter is implemented as a single base interpreter pass via `EagerEval(Optimize(Deoptimize(Desugar(eval))))`

fritzo: Hmm, I'll have to think about this more. I guess I'll have to change e.g.
```diff
  def _eager_subs(self, **kwargs):
-     return Binary(self.op, self.lhs(**kwargs), self.rhs(**kwargs))
+     return eager_binary(self.op, self.lhs._eager_subs(**kwargs), self.rhs._eager_subs(**kwargs))
```

fritzo: Hmm I think the new non-reentrant behavior will prevent eager evaluation of terms like
```py
(Variable('x', 'real') + 1 + 1 + 1)(x=1)
```

fritzo: Ok, I'm making those changes now.

fritzo:
After more thought, I think we'll want to instead:
- remove handlers from `funsor.terms` (if `engine.interpeter.eval` is still used, it will need to be wrapped in `effectful`)
- move `Memoize` logic back into `FunsorMeta.__call__()`
- make eager+memoize the default evaluation behavior (optimizability is achieved by avoiding materializing)
- move the `self.binary(op, other)` method to a pattern matching function `binary_op(op, lhs, rhs)`

The only part of this PR that actually helps unblock the kalman filter example is the `binary_op(op, lhs, rhs)`, and it's easier to do that in separate smaller PRs.

eb8680:
> After more thought, I think we'll want to instead: ...

Ok.  It seems like the issue is really that `Substitution` needs to be handled before the rest of eager evaluation, like broadcasting.  I don't really follow the alternative, so I'll leave any work on it to you unless you want PRs reviewed.

Issue #19, opened 2019-02-22
Add discrete HMM example


eb8680:
A simple discrete eager HMM without plates seems like the simplest end-to-end example we could get working that would exercise lots of funsor machinery at once.

Tasks
- [x] Write a complete but xfailing example script that we can develop against
- [ ] Figure out what's missing/broken, especially in `distributions`

PR #20, opened 2019-02-22
Add xfailing discrete hmm example

eb8680:
Part of #19


fritzo:
LGTM after fixing lint errors

PR #21, opened 2019-02-22
Fix Function, Substitute, Reduce; add more tests

fritzo:


funsor/torch.py
```diff
@@ -276,13 +267,16 @@ def __init__(self, inputs, output, fn):
...
     def eager_subs(self, subs):
+        assert isinstance(subs, tuple)
+        subs = dict(subs)
         if not all(isinstance(subs.get(key), (Number, Tensor)) for key in self.inputs):
             return None  # defer to default implementation
+
         funsors = tuple(subs[key] for key in self.inputs)
         inputs, tensors = align_tensors(*funsors)
         data = self.fn(*tensors)
         inputs = None  # FIXME
```
eb8680: Is this correct now or is something still missing?

fritzo: Good catch, I believe this is fixed now, I'll remove this comment in a follow-up PR

Issue #10, opened 2019-02-16
Add plated einsum example


fritzo:
Add an example implementing plated einsum via funsors. Funsors are arguably a cleaner DSL for expressing these kinds of dynamic programs, and the example could demonstrate both how to implement general plated einsum, and how to express various specific problems using funsor syntax rather than einsum syntax.

## Outline
- [x] add a non-plated einsum example
- [x] add plated einsum implementation

eb8680:
Even a completely naive working version of this will be really useful for testing

eb8680:
Addressed by #46

PR #23, opened 2019-02-22
Add naive einsum implementation and tests

eb8680:
Part of #10

### DIFF
--- a/test/test_einsum.py
+++ b/test/test_einsum.py
@@ -71,7 +71,7 @@ def test_einsum(equation):
             assert actual.inputs[output_dim].dtype == sizes[output_dim]
 
 
-PLATED_EINSUM_EXAMPLES = [
+PLATED_EINSUM_EXAMPLES = [(ex, '') for ex in EINSUM_EXAMPLES] + [
     ('i->', 'i'),
     ('i->i', 'i'),
     (',i->', 'i'),
--- a/test/test_einsum.py
+++ b/test/test_einsum.py
@@ -12,6 +12,10 @@ import funsor
 import funsor.ops as ops
 
 
+def xfail_param(*args, **kwargs):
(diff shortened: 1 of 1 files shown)


test/test_einsum.py
```diff
@@ -0,0 +1,101 @@
...
+    # ...
+    raise NotImplementedError("TODO implement naive plated einsum")
+
+
+EINSUM_EXAMPLES = [
+    "a,b->",
+    "ab,a->",
+    "a,a->",
+    "a,a->a",
+    "a,a,a,ab->ab",
```
fritzo: We should add an xfailing example
```
"ab->ba"
```
To make this pass we should resurrect [.align()](https://github.com/pyro-ppl/funsor/blob/da322019d355bd43d564a6b5b42c65402b2df064/funsor/terms.py#L105) and [Align](https://github.com/pyro-ppl/funsor/blob/da322019d355bd43d564a6b5b42c65402b2df064/funsor/terms.py#L388). I'll add this as a separate issue.

fritzo: I've added issue #26

test/test_einsum.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def naive_einsum(eqn, *terms):
+    assert isinstance(eqn, str)
+    assert all(isinstance(term, funsor.Funsor) for term in terms)
+    inputs, output = eqn.split('->')
+    input_dims = frozenset(d for inp in inputs.split(',') for d in inp)
+    output_dims = frozenset(d for d in output)
+    reduce_dims = tuple(d for d in input_dims - output_dims)
+    prod = terms[0]
```
fritzo: I guess you could even
```py
return reduce(ops.mul, terms).sum(reduce_dims)
```

Issue #27, opened 2019-02-23
Remove Substute; make all Funsors closed under substitution


fritzo:
## Tasks

- [x] ~~#28~~ #29 Make `Function` closed under substitution
- [x] #29 Remove `Substitute` (requires refactoring `.eager_subs()` methods)
- [ ] #33 Make `Tensor` closed under substitution (and disallow real `inputs` #29)
- [ ] Make `Stack`/`Branch`/`Switch` closed under substitution

fritzo:
It appears that lazy substitution can simplify distribution code and allow e.g. transformed distributions and factorial distributions.

PR #28, opened 2019-02-23
Make funsor.torch.Function closed under substitution

fritzo:
Addresses #27 

This makes `Function` closed under substitution, following the pattern of `Distribution` objects.

@eb8680 It is a little more subtle that I expected to make `Function` generic and let `Unary` and `Binary` inherit from `Function`. I suggest we prioritize #27 and then revisit our new class hierarchy.

## Tested

- updated an existing unit test


fritzo:
Closing in favor of #29

PR #29, opened 2019-02-24
Drop Substitute funsor; close Function under substitution

fritzo:
Addresses #27 
Supersedes #28 

Thanks @eb8680 for suggesting this simplifying change!

1. Makes `Function` closed under substitution, following the pattern of `Distribution` objects.
2. Drops lazy `Substitution` objects and the global `eager_subs` function.
3. Updates `Tensor` and `Stack` to raise `NotImplementedError('TODO implement advanced indexing')`
4. Adds a check in `Tensor` to ensure tensors cannot depend on real free variables. (We may later create something like a `Branch(Tensor(...))` when substituting a real-dependency into a `Tensor`, but that requires more thinking).

## Tested

- updated an existing unit test

### DIFF
--- a/funsor/distributions.py
+++ b/funsor/distributions.py
@@ -7,7 +7,7 @@ import pyro.distributions as dist
 import funsor.ops as ops
 from funsor.domains import ints, reals
 from funsor.pattern import simplify_sum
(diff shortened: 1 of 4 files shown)


fritzo:
Thanks for reviewing!

PR #30, opened 2019-02-25
add cuda tests

jpchen:


jpchen:
> 2. That avoids overhead when creating each new test file.

yeah that was intentional in case you didnt want to test certain files on cuda, but i guess you can individually x-fail tests as in pyro

PR #37, opened 2019-02-27
Implement general Gaussian funsor

fritzo:
This implements a minimal `Gaussian` class with interface similar to `Tensor`.

The goal is to generalize Tensor Variable Elimination from discrete random variables to multivariate Gaussian variables. This PR follows the current Funsor pattern of compiling a `Categorical` distribution to a `Tensor`, and making `Tensor` closed under a number of variable elimination operations. I.e. this PR aims to compile `Normal` (and later `MultivariateNormal`) to a `Gaussian` which is closed under operations used in variable elimination.

@eb8680 points out that `Gaussian` is still not closed under all substitutions, specifically nonlinear substitutions. However this problem also affects `Tensor`, which is not closed under real-valued substitutions. @fritzo suggested we try to get `Gaussian` working as well as `Tensor` currently works, and then explore how we can close them under general substitutions.

## Future work

This PR does not attempt to discover affine relationships among variables. That question is for future research, and may be solved by one of: operator fusion, autograd, or general symbolic algebra. Early work on this has been moved to the [gaussian-affine](https://github.com/pyro-ppl/funsor/tree/gaussian-affine) branch.

This PR does not implement an efficient `eager_contract` function; instead we simulate this with the less efficient `Reduce(ops.logaddexp, Binary(ops.add, -, -), reduced_dims)` combination.

## Tasks

- [x] implement `Gaussian.__init__()`
- [x] implement `Gaussian.eager_subs()` (only a very limited version)
- [x] implement `Gaussian.eager_reduce(ops.logaddexp, ...)`
- [x] implement `eager_binary(ops.add, Gaussian, Number)`
- [x] implement `eager_binary(ops.add, Number, Gaussian)`
- [x] implement `eager_binary(ops.add, Gaussian, Tensor)`
- [x] implement `eager_binary(ops.add, Tensor, Gaussian)`
- [x] implement `eager_binary(ops.add, Gaussian, Gaussian)`
- [x] implement `eager_normal(...)` -> `Gaussian` conversion
- [x] update kalman_filter.py
- [x] add tests for `Gaussian`
- [x] add tests for `Normal` -> `Gaussian`
- [x] add `kalman_filter.py` test to Makefile

**Triaged tasks:**
- implement `Gaussian.eager_reduce(ops.add, ...)`
- implement `Affine` class
- implement `to_affine()` function
- implement partial substitution in `eager_subs`


funsor/gaussian.py
```diff
@@ -0,0 +1,182 @@
...
+def eager_binary_gaussian_gaussian(op, lhs, rhs):
+    if op is ops.add:
+
+        if lhs.inputs != rhs.inputs:
+            raise NotImplementedError('TODO align vectors and matrices')
+
+        inputs = lhs.inputs
+        lhs_precision = _scale_tril_to_precision(lhs.scale_tril)
+        rhs_precision = _scale_tril_to_precision(rhs.scale_tril)
+        precision = lhs_precision + rhs_precision
```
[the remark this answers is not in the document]
fritzo: This implements the pointwise sum of two log-density distributions, which corresponds to Gaussian "update" or "fusion". For this op the precisions do add.

funsor/gaussian.py
```diff
@@ -0,0 +1,263 @@
...
+        rhs_log_density, rhs_loc, rhs_precision = rhs_moments
+
+        # Fuse aligned Gaussians.
+        precision = lhs_precision + rhs_precision
+        scale_tril_inv = torch.cholesky(precision)
+        scale_tril = torch.inverse(scale_tril_inv)
+        precision_loc = _mv(lhs_precision, lhs_loc) + _mv(rhs_precision, rhs_loc)
+        loc = _mv(scale_tril, _mv(scale_tril.transpose(-1, 2), precision_loc))
+        log_density = (lhs_log_density + rhs_log_density +  # FIXME add missing terms
+                       _batch_mahalanobis(scale_tril, lhs_loc - rhs_loc))
```
fehiepsi: It seems to me (see [my derivation](https://gist.github.com/fehiepsi/5ef8e09e61604f10607380467eb82006)) that the required term is
```
_batch_mahalanobis(lhs.scale_tril, lhs_loc) + _batch_mahalanobis(rhs.scale_tril, rhs.scale_tril) - _batch_mahalanobis(scale_tril, loc) + (logdet_terms if it is involved??)
```
I am not sure I have derived the terms in a correct way. As a side node, we can use Woobury matrix identity to compute `scale_tril` (to avoid working with precision matrix which involves `inverse` at many places). But we can do it later (I'll update the above gist in the mean time).

fritzo: Thanks @fehiepsi! Yes I'd like to improve the math after this initial PR.

fritzo: @fehiepsi I believe the formula in your derivation is correct, but I believe a more numerically stable version can be arrived at by restricting to a delta term `lhs_loc - rhs_loc`, I believe something like
```
(lhs_loc - rhs_loc)' * ((lhs_precision rhs_precision) / precision) * (lhs_loc - rhs_loc)
```
with suitable realignments. I had derived this for the univariate case, but I'm not sure how to adapt it to multivariate.

fehiepsi: Thanks @fritzo ! I agree that we can cancel some terms involving `c` in the notebook to arrive at your formula with `p_1 p_2 / (p_1 + p_2) = `(cov_1 + cov_2)^{-1}`. I'll work that out tomorrow, including the logdet terms.

fehiepsi: @fritzo I just updated the notebook with come up with the multivariate form of your formula. It is interesting that we can derive a nice form for log density adjustment (it is log density at lhs_loc - rhs_loc of a Gaussian with covariance is the sum of two covariances). Though I still think that using the formula involving Mahalanobis terms is better, we should address it later because this PR is already big!

Please ping me when this PR is ready so I'll take another pass for other places too. :)

fritzo:
@eb8680 This is ready for review. I'm happy to walk you through this on Monday.

@fehiepsi It turns out we need to use the `precision` representation since we sometimes deal with incomplete information (zero eigenvalues in the precision matrix). The only effort I've made to keep things numerically stable in this PR is to ensure all `torch.inverse()` calls happen on an Cholesky factor.

fehiepsi:
@fritzo Agree, using precision matrix also seems natural with the computation. You can use `torch.potrs` with cholesky factor to avoid calling (`torch.inverse` + matmul + matmul) but I guess it does the same thing behind the scene so no need to concern about it. The `quadratic_term` seems new to me. I'll walk through your PR tomorrow to understand what is going on here. :)

fehiepsi:
So no need for me to take a look... +1

eb8680:
@fehiepsi this was a pretty large PR blocking other work, we just wanted to get this merged so we could fix things in smaller followup PRs even if everything isn't perfect.  We'd still really like your feedback (or even PRs fixing things), especially on making the Gaussian math more numerically stable.  Feel free to go through the diff and ask questions here or on Slack!

fehiepsi:
Thanks! Indeed, I'm still not able to understand the logic behind functions `eager_normal(Variable, Tensor, Variable)`, `eager_subs`, `eager_reduce` so it is hard for me to have more comments on this PR. Instead, most of them will just be questions. Because I'm stuck into the traditional viewpoint, it will take me lot of effort to be able to keep up with this line of work. I'll take another chance in the future to get a sense of what the team are doing here. :) It looks interesting by the way!

Issue #9, opened 2019-02-16
Support plates in optimizer


fritzo:
Currently the optimizer handles only non-plated variable elimination, but we'd like to additionally support [tensor variable elimination](https://arxiv.org/abs/1902.03210).

PR #40, opened 2019-03-01
Reinstate opt_einsum-based optimizer

eb8680:
This PR adds an updated version of the old opt_einsum optimizer and tests it against the existing `test_einsum` tests.

Tasks:
- [x] Rewrite `Binary` ops to `Finitary`
- [x] Exploit associativity of `Finitary` ops
- [x] Exploit associativity of `Reduce` ops
- [x] Exploit distributivity for `Reduce` and `Finitary`
- [x] Support multiple semirings
- [x] Support distributivity in mixed terms like `Finitary(Reduce, Finitary)`
- [x] Add HMM-style test cases that only work with optimization enabled
- [x] Add more tests that exercise edge cases

Triaged:
- Distribute `Finitary` ops over one another (no optimizer for this currently)
- Implement tensor variable elimination #9 
- Handle variable name collisons correctly throughout the optimizer instead of assuming global scope

### DIFF
--- a/funsor/optimizer.py
+++ b/funsor/optimizer.py
@@ -178,8 +178,8 @@ def optimize_reduction(op, arg, reduced_vars):
 
         # don't reduce a dimension too early - keep a collections.Counter
         # and only reduce when the dimension is removed from all lhs terms in path
-        reduce_dim_counter.subtract({d: 1 for d in reduced_vars & set(ta.inputs.keys())})
-        reduce_dim_counter.subtract({d: 1 for d in reduced_vars & set(tb.inputs.keys())})
+        reduce_dim_counter.subtract({d: 1 for d in reduced_vars & frozenset(ta.inputs.keys())})
+        reduce_dim_counter.subtract({d: 1 for d in reduced_vars & frozenset(tb.inputs.keys())})
 
         # reduce variables that don't appear in other terms
         both_vars = frozenset(ta.inputs.keys()) | frozenset(tb.inputs.keys())
--- a/funsor/optimizer.py
+++ b/funsor/optimizer.py
@@ -151,11 +151,12 @@ def optimize_reduction(op, arg, reduced_vars):
     size_dict = {}
     for operand in arg.operands:
         inputs.append(frozenset(d for d in operand.inputs.keys()))
-        size_dict.update({k: (REAL_SIZE if v.dtype == 'real' else v.dtype)
+        size_dict.update({k: ((REAL_SIZE * v.num_elements) if v.dtype == 'real' else v.dtype)
                           for k, v in operand.inputs.items()})
     outputs = frozenset().union(*inputs) - reduced_vars
 
     # optimize path with greedy opt_einsum optimizer
+    # TODO switch to new 'auto' strategy when it's released
     path = greedy(inputs, outputs, size_dict)
 
     # convert path IR back to sequence of Reduce(Finitary(...))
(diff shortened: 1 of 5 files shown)


funsor/optimizer.py
```diff
@@ -0,0 +1,205 @@
...
+    r"""
+    Recursively convert large Reduce(Finitary) ops to many smaller versions
+    by reordering execution with a modified opt_einsum optimizer
+    """
+    # build opt_einsum optimizer IR
+    inputs = []
+    size_dict = {}
+    for operand in arg.operands:
+        inputs.append(frozenset(d for d in operand.inputs.keys()))
+        size_dict.update({k: (REAL_SIZE if v.dtype == 'real' else v.dtype)
```
fritzo: Consider setting this to `REAL_SIZE * v.num_elements` so as to account for increased expense of multivariate tensors.

funsor/optimizer.py
```diff
@@ -0,0 +1,205 @@
...
+                          for k, v in operand.inputs.items()})
+    outputs = frozenset().union(*inputs) - reduced_vars
+
+    # optimize path with greedy opt_einsum optimizer
+    path = greedy(inputs, outputs, size_dict)
```
fritzo: FYI we should switch to the new [auto strategy[(https://github.com/dgasmith/opt_einsum/commit/f9afb0b4ba440a247ff55312e8cc20ec38752b71) once it is released.

test/test_einsum.py
```diff
@@ -56,6 +53,9 @@ def naive_plated_einsum(eqn, *terms, **kwargs):
     "a,a->a",
     "a,a,a,ab->ab",
     "a,ab,bc,cd->",
+    make_hmm_einsum(10),
+    # make_hmm_einsum(15),  # slows down tests
+    # make_hmm_einsum(20),  # slows down tests
```
eb8680: I don't think the optimizer is working correctly yet, it's failing to produce optimized expressions for these examples.

funsor/optimizer.py
```diff
@@ -0,0 +1,224 @@
...
+def associate_finitary(op, operands):
+    # Finitary(Finitary) -> Finitary
+    new_operands = []
+    for term in operands:
+        if isinstance(term, Finitary) and term.op is op and op in ASSOCIATIVE_OPS:
+            new_operands.extend(term.operands)
+        else:
+            new_operands.append(term)
+
+    with interpretation(reflect):
```
fritzo: So far I've been able to avoid changing interpretation inside constructors by using logic like
```py
def associate_finitary(op, operands):
    if any(isinstance(x, Finitary) for x in operands):
        # Eagerly fuse Finitary terms.
        new_operands = []
        ...
        return Finitary(op, new_operands)

```
This seems a little safer, but I don't feel strongly either way.

I'll merge now, but feel free to send a quick follow-up PR if you prefer the other style.

funsor/optimizer.py
```diff
@@ -0,0 +1,224 @@
...
+        result = reflect(cls, *args)
+    return result
+
+
+_optimize = KeyedRegistry(default=lambda *args: None)
+optimize.register = _optimize.register
+
+
+# TODO set a better value for this
+REAL_SIZE = 3  # the "size" of a real-valued dimension passed to the path optimizer
```
fritzo: I've started a discussion about this in #45

PR #47, opened 2019-03-06
Fix path evaluation order in optimizer

eb8680:
Addresses #9 

This PR fixes the evaluation of pairwise contractions in the optimizer.  Previously the path produced by `opt_einsum` was being interpreted incorrectly, causing optimized evaluation to evaluate the wrong pairs of contractions and produce exponentially large intermediate results.  As a demonstration, I added some large plated HMM contractions to `test_optimizer.py` that would run out of memory without this fix.


eb8680:
Test failures are just numerical instability

PR #49, opened 2019-03-06
Add a simple Delta distribution

fritzo:
Addresses #51

This PR
1. adds a simple `Delta` distribution to `funsor.distributions`
2. refactors all distributions to use a `._fill_defaults()` method, reducing metaclass boilerplate
3. removes `Number` handling from distribution pattern matching, since numbers should already be converted by `numbers_to_tensors()` in `DistributionMeta`.

This PR will be followed up by a general `Delta` class similar to `Tensor` and `Gaussian`.

## Tested
- added new unit tests for `dist.Delta`

### DIFF
--- a/funsor/distributions.py
+++ b/funsor/distributions.py
@@ -30,6 +30,17 @@ def numbers_to_tensors(*args):
     return args
 
 
+class DistributionMeta(FunsorMeta):
+    """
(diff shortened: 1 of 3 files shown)


fritzo:
I'm going to quick add some tests for nontrivial event shape.

PR #54, opened 2019-03-07
Implement Monte Carlo interpretation of Integrate

### LINKED ISSUE #52: Implement Monte Carlo interpretation of .logaddexp()
It appears that the natural way to express Monte Carlo sampling in funsor is with a nonstandard interpretation of `.logaddexp()` reductions. (Note that for numerical stability, we represent densities in log-space; hence expectation is represented by `.logaddexp()` reductions rather than `.sum()` reductions).

To perform sampling on factor graphs, we will orthogonally need #50.

## Tasks

- [ ] #75 Implement monte carlo `.sample()` methods
- [ ] Implement a funsor `Integrate(log_measure, integrand, reduced_vars)`
- [ ] Use `Integrate()` in test_samplers.py to test `Gaussian` moments
- [ ] Implement a `monte_carlo` interpretation that uses samplers in `Integrate()`

fritzo:
Resolves #52 

This PR:
1. Adds an `Integrate(log_measure, integrand, reduced_vars)` funsor.
2. Adds a `monte_carlo` interpretation of the `Integrate` funsor.
3. Changes `.sample()` to include the normalization factor (i.e. `1/product(sample_shape)`).
4. Fixes logs of little flaws:
    - supporting broadcasting of output shape in `Tensor`
    - fixed & clarified some Cholesky manipulations `Gaussian`
    - fixed handling of `torch.nn.Module` in `funsor.torch.function`
    - fixed `GetitemOp.__name__`
    - made `funsor.testing` a little more robust
    - improved `FUNSOR_DEBUG` logging
    - ...

## Tests
- [x] test Tensor-Tensor
- [x] test Gaussian-Gaussian
- [x] regression test for broadcasting bug

## Triaged
- add examples/vae.py to Makefile (moved to #95)


funsor/expectation.py
```diff
@@ -0,0 +1,97 @@
...
+    def eager_subs(self, subs):
+        raise NotImplementedError('TODO')
+
+
+@eager.register(Expectation, Funsor, Funsor, frozenset)
+def eager_expectation(measure, integrand, reduced_vars):
+    return (measure * integrand).reduce(ops.add, reduced_vars)
+
+
+@monte_carlo.register(Expectation, Funsor, Funsor, frozenset)
```
fritzo: @eb8680 @martinjankowiak Here is a sketch of a monte carlo `Expectation` implementation that matches probability factors to the terms they depend on, as in `TraceGraph_ELBO` or `Dice`. This approach isn't quite right, but it is a step towards a feasible design. Thoughts?

eb8680: I spent this afternoon working through implementing a similar `Integrate` class to help myself understand all the issues involved.  I think something like this is basically what we want - let's talk tomorrow.

examples/vae.py
```diff
@@ -0,0 +1,100 @@
...
+        q = q.reduce(ops.add, frozenset(['i']))
+
+        probs = decode(z)
+        x = funsor.Variable('x', bint(28))
+        y = funsor.Variable('y', bint(28))
+        p = dist.Bernoulli(probs[x, y], value=data[x, y])
+        p = p.reduce(ops.add, frozenset(['x', 'y']))
+
+        elbo = funsor.Integrate(q, scale * (p - q), 'z')
+        elbo = elbo.reduce(ops.add, frozenset(['batch']))
```
neerajprad: I see that the `batch` dim is only being referenced here the first time. Do all the funsors above, like `encode`, `dist.Normal` work with arbitrary batch dims on the left?

fritzo: Yes, we've taken care that all funsors "broadcast" to allow arbitrary batch dims.

fritzo:
@eb8680 This is now ready for review. It's probably easier for me to walk you through the changes, rather than split this into smaller PRs. I've split examples/vae.py out as #95 since that involves shape issues.

funsor/gaussian.py
```diff
@@ -26,10 +28,14 @@ def _issubshape(subshape, supershape):
     return True
 
 
-def _log_det_tril(x):
+def _log_det_tri(x):
```
fritzo: It turns out some of our matrices are upper triangular, but that's ok since these functions work with both upper- and lower- triangular matrices.

funsor/gaussian.py
```diff
@@ -317,13 +335,71 @@ def eager_add_gaussian_gaussian(op, lhs, rhs):
...
-    scale_tril = torch.inverse(torch.cholesky(precision))
-    loc = _mv(scale_tril.transpose(-1, -2), _mv(scale_tril, precision_loc))
+    scale_tri = torch.inverse(torch.cholesky(precision)).transpose(-1, -2)
+    loc = _mv(scale_tri, _mv(scale_tri.transpose(-1, -2), precision_loc))
     quadratic_term = _vmv(lhs_precision, loc - lhs_loc) + _vmv(rhs_precision, loc - rhs_loc)
     likelihood = Tensor(-0.5 * quadratic_term, int_inputs)
     return likelihood + Gaussian(loc, precision, inputs)
 
 
+@eager.register(Integrate, Gaussian, Variable, frozenset)
```
fritzo: It would be nice to generalize this to arbitrary polynomials.

funsor/interpreter.py
```diff
@@ -24,12 +25,18 @@ def interpret(cls, *args):
...
         print(indent + ' '.join(typenames))
+
         _STACK_SIZE += 1
         try:
             result = _INTERPRETATION(cls, *args)
         finally:
             _STACK_SIZE -= 1
-        print(indent + '-> ' + type(result).__name__)
+
+        if _DEBUG > 1:
```
fritzo: You can specify `FUNSOR_DEBUG=2` for even more verbose output.

funsor/integrate.py
```diff
@@ -0,0 +1,85 @@
...
+@eager.register(Integrate, Reduce, Funsor, frozenset)
+@integrator
+def eager_integrate(log_measure, integrand, reduced_vars):
+    if log_measure.op is ops.logaddexp:
+        if not log_measure.reduced_vars.isdisjoint(reduced_vars):
+            raise NotImplementedError('TODO alpha convert')
+        arg = Integrate(log_measure.arg, integrand, reduced_vars)
+        return arg.reduce(ops.add, log_measure.reduced_vars)
+
+    return None  # defer to default implementation
```
fritzo: TODO: `return Contract(log_measure.exp(), ...)`

PR #55, opened 2019-03-07
Implement Op wrapper class to enable pattern matching

fritzo:
This implements an `Op` wrapper class around operators to enable:
1. pattern matching (e.g. `Op` versus `AssociativeOp`), and
2. slightly cleaner pretty printing of expressions.

After this PR we might further create additional singleton classes to clean up some of our pattern matching logic, e.g.
```py
# OLD
@eager.register(Binary, Op, Gaussian, Gaussian)
def binary_gaussian_gaussian(op, lhs, rhs):
    if op is ops.add:
        ...do stuff...

# NEW
@eager.register(Binary, AddOp, Gaussian, Gaussian)
def add_gaussian_gaussian(op, lhs, rhs):
    assert op is ops.add
    ...do stuff...
```
We might also simplify registration of pairs of functions by defining a special `CommutativeOp` registration mechanism.

## Tested
- exercised by existing unit tests
- I verified that this has no measurable performance impact

### DIFF
--- a/funsor/optimizer.py
+++ b/funsor/optimizer.py
@@ -1,15 +1,15 @@
 from __future__ import absolute_import, division, print_function
 
 import collections
-from six.moves import reduce
 
 from opt_einsum.paths import greedy
+from six.moves import reduce
 
 from funsor.domains import find_domain
 from funsor.interpreter import interpretation, reinterpret
-from funsor.ops import ASSOCIATIVE_OPS, DISTRIBUTIVE_OPS
+from funsor.ops import DISTRIBUTIVE_OPS, AssociativeOp
 from funsor.registry import KeyedRegistry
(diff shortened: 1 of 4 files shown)


funsor/optimizer.py
```diff
@@ -64,12 +64,12 @@ def binary_to_finitary(op, lhs, rhs):
     return Finitary(op, (lhs, rhs))
 
 
-@associate.register(Finitary, object, tuple)
+@associate.register(Finitary, AssociativeOp, tuple)
```
fritzo: @eb8680 can you review the change here carefully? I believe it is correct, but it is a bit subtle since this pattern will be matched less often after this PR.

eb8680: Seems fine, since the default behavior in `associate` is to `reflect`, preserving the previous behavior in all cases.

PR #56, opened 2019-03-08
Update minipyro storyboard to correctly handle plates

fritzo:
This updates minipyro.py to handle plates, based on a future `funsor.contract` module.

Minipyro.py is currently only a storyboard: it does not yet function but does demonstrate the kind of computations we want to write in funsor. This PR updates our `pyro.plate` and `log_joint` data structures to clarify how we might move plate logic into `Domain` or another lower-level place. In particular, this PR adds use of partial tensor variable elimination to minipyro.

### DIFF
--- a/funsor/einsum.py
+++ b/funsor/einsum.py
@@ -59,7 +59,7 @@ def _partition(terms, sum_vars):
     return components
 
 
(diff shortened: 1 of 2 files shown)


fritzo:
@martinjankowiak you might be interested in this PR, let me know if you want a tour of the code some morning.

PR #58, opened 2019-03-08
Add numpy backend for funsor

neerajprad:
This adds a `funsor.numpy` backend in correspondence with the `funsor.torch` backend. I am surprised how easy it was to do this, and loved the design! 😄 
 - I'll comment on the places where I had to make some (very minor) changes.
 - The biggest difference in the API is that `np.ndarray` doesn't have attributes like `abs`, `log`, etc. and we need to dispatch to `np.abs`, `np.log` instead. I am still trying to get a better handle on the changes needed to support this in a clean way.
 - Currently, the code in `funsor.numpy` is very similar to `funsor.torch`, and we can discuss whether it makes sense to share more code and also tests (the only flipside is tighter coupling since the API of numpy and torch while largely very similar has some subtle differences).
 - I did not want to clutter this by using jax with its own dependencies, but we should be able to swap that out later with some minor changes.

### DIFF
--- a/funsor/terms.py
+++ b/funsor/terms.py
@@ -550,6 +550,7 @@ class Number(Funsor):
             data = type(dtype)(data)
         else:
             assert isinstance(dtype, str) and dtype == "real"
+            data = float(data)
         inputs = OrderedDict()
         output = Domain((), dtype)
         super(Number, self).__init__(inputs, output)
--- a/funsor/terms.py
+++ b/funsor/terms.py
@@ -31,15 +31,11 @@ from funsor.registry import KeyedRegistry
 from funsor.six import getargspec, singledispatch
 
 
-def _cons_cache_key(*args):
-    return tuple(id(arg) if not isinstance(arg, Hashable) else arg for arg in args)
-
-
 def reflect(cls, *args):
     """
     Construct a funsor, populate ``._ast_values``, and cons hash.
     """
-    cache_key = _cons_cache_key(*args)
+    cache_key = tuple(id(arg) if not isinstance(arg, Hashable) else arg for arg in args)
     if cache_key in cls._cons_cache:
         return cls._cons_cache[cache_key]
     result = super(FunsorMeta, cls).__call__(*args)
--- a/funsor/terms.py
+++ b/funsor/terms.py
@@ -16,7 +16,7 @@ import functools
 import itertools
 import numbers
 from abc import ABCMeta, abstractmethod
-from collections import OrderedDict
+from collections import OrderedDict, Hashable
 from weakref import WeakValueDictionary
 
 from six import add_metaclass, integer_types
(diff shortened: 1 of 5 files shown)


funsor/terms.py
```diff
@@ -30,15 +30,20 @@
 from funsor.six import getargspec, singledispatch
 
 
+def _cons_cache_key(*args):
+    return tuple(id(arg) if not isinstance(arg, Hashable) else arg for arg in args)
```
neerajprad: I added this because `np.ndarray`s are not hashable, and should also work in case `Tensor.__hash__` is deprecated with https://github.com/pytorch/pytorch/issues/7733. Currently this would be a no-op for Tensor.

funsor/terms.py
```diff
@@ -545,7 +550,6 @@ def __init__(self, data, dtype=None):
             data = type(dtype)(data)
         else:
             assert isinstance(dtype, str) and dtype == "real"
-            data = float(data)
```
neerajprad: This took me quite a long while to track down. The issue here is that due to the behavior below, if we have a funsor expression like `u = 1 - v`, where v is a `Variable`, this would always give us an array of dtype `np.float64` instead of `np.int64` when we call `numpy.materialize`. The reason why this works for tensors is because PyTorch does not do dtype promotion. 

This change doesn't seem to break any existing tests, and seemed intuitive, but let me know if this is likely to cause other unintended consequences.

```python
>>> 1 - np.array([0, 1])
 array([1, 0])  # since 1 is int we get an long array

>>> 1. - np.array([0, 1])
 array([1., 0.])  # since 1. is float we get a double array 

>>> 1 - torch.tensor([0, 1])
 tensor([1, 0])  # long tensor

>>> 1. - torch.tensor([0, 1])
 tensor([1, 0])  # tensor type isn't promoted; still long tensor
```

fritzo: I'd rather keep these as Python floats here and do the downcasting to np.float32 in `binary_number_array`. The original intent was to avoid `int`s being stored as real `Number`s, because arithmetic would break in Python 2 user code. In funsor, there is no unbounded integer type, only bounded integers `bint(n)`. If you want to add a number to a long tensor, you should know the bound beforehand, like:
```py
1 + Tensor(torch.tensor(1, dtype=torch.long))  # should be a double tensor
Number(1) + Tensor(torch.tensor(1, dtype=torch.long))  # should be a real tensor
Number(1, 10) + Tensor(torch.tensor(1, dtype=torch.long), dtype=10)  # should be a long tensor dtype is bint(10)
```
Let's discuss subtleties though.

neerajprad: That makes sense. I suppose one thing that we may need to be careful with then are methods that operate on the raw data type. e.g. in the first example, the underlying `.data` attribute will be holding a `torch.LongTensor`. e.g. I had to recast all the `.data` for indices to be `np.int64` in `Array.eager_subs` because the underlying numpy arrays were promoted to double even though the Array's dtype wasn't "real".

funsor/terms.py
```diff
@@ -30,15 +30,20 @@
 from funsor.six import getargspec, singledispatch
 
 
+def _cons_cache_key(*args):
```
fritzo: nit: I believe it is slightly cheaper to pass `args` rather than `*args`. I might even inline this one-line function.

neerajprad:
This should be good for review. I have removed the changes to the `ops.py` file, and will address that in a separate PR.

test/test_numpy.py
```diff
@@ -0,0 +1,192 @@
...
+import funsor
+from funsor import bint, reals, Variable, Number
+from funsor.numpy import Array
+import numpy as np
+
+from funsor.testing import check_funsor, assert_equiv, random_array
+
+
+@pytest.mark.parametrize('shape', [(), (4,), (3, 2)])
+@pytest.mark.parametrize('dtype', [np.float32, np.float64, np.int32, np.int64, np.uint8])
```
fritzo: :+1:

fritzo:
It would also be nice to fork `funsor.torch.Function` so we can easily wrap numpy functions.

PR #59, opened 2019-03-08
 Assert that args to Tensor.eager_subs have bounded int type

neerajprad:
Refer to the discussion in #58. 

This adds a check in `eager_subs` to assert that the dtype of any substitution value is an int and not "real". This is more important for the numpy backend where `np.ndarray`s have their dtype promoted due to various binary operations, and therefore results of expressions such as `1 + Array(np.array([1])`  (which would have dtype="real" and `.data` backed by a np.float64 array) shouldn't be allowed as argument to `eager_subs`. This adds a little bit of verbosity in the call, so just opening this for discussion.

### DIFF
--- a/funsor/torch.py
+++ b/funsor/torch.py
@@ -139,10 +139,10 @@ class Tensor(Funsor):
         data = self.data.permute(tuple(old_dims.index(d) for d in new_dims))
         return Tensor(data, inputs, self.dtype)
 
-    def eager_subs(self, subs_idxs):
-        assert isinstance(subs_idxs, tuple)
+    def eager_subs(self, dim_subs):
+        assert isinstance(dim_subs, tuple)
         subs = {}
-        for k, v in subs_idxs:
+        for k, v in dim_subs:
             if k in self.inputs:
                 if not isinstance(v, Number):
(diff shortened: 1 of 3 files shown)


funsor/torch.py
```diff
@@ -139,9 +139,14 @@ def align(self, names):
...
-    def eager_subs(self, subs):
-        assert isinstance(subs, tuple)
-        subs = {k: materialize(v) for k, v in subs if k in self.inputs}
+    def eager_subs(self, dim_subs):
+        assert isinstance(dim_subs, tuple)
+        subs = {}
+        for k, v in dim_subs:
+            if k in self.inputs:
+                if not isinstance(v, Number):
+                    assert v.output != reals(), "subs for dim {} must be of bounded integer (bint) type.".format(k)
```
fritzo: Hmm, I think this would be better to do in the top level `Funsor.__call__()` to catch more errors. Let me try to put up a quick PR...

fritzo: Ok, I think this is a little safer: https://github.com/pyro-ppl/funsor/pull/62

neerajprad: That looks great! Let me close this PR.

Issue #50, opened 2019-03-06
Implement a generic funsor.adjoint algorithm


fritzo:
See [design doc](https://docs.google.com/document/d/1NVlfQnNQ0Aebg8vfIGcJKsnSqAhB4bbClQrb5dwm2OM/edit#heading=h.nbajrdyksttw) for details.

PR #64, opened 2019-03-09
Add prototype funsor.adjoint module

eb8680:
Addresses #50.  Pair coded with @fritzo 

This PR adds a `funsor.adjoint` module for forward-backward tensor variable elimination algorithms.  Currently it only supports plated forward-filter backward-marginal computations.

### DIFF
--- a/funsor/adjoint.py
+++ b/funsor/adjoint.py
@@ -7,7 +7,7 @@ import funsor.ops as ops
 from funsor.interpreter import interpretation, reinterpret
 from funsor.ops import AssociativeOp
 from funsor.registry import KeyedRegistry
-from funsor.terms import Binary, Funsor, Number, Reduce, Variable, reflect
+from funsor.terms import Binary, Funsor, Number, Reduce, Variable, eager
 from funsor.torch import Tensor
 
 
@@ -17,7 +17,7 @@ class AdjointTape(object):
         self.tape = []
 
     def __call__(self, cls, *args):
-        result = reflect(cls, *args)
+        result = eager(cls, *args)
         if cls in (Reduce, Binary, Tensor):
             self.tape.append((result, cls, args))
         return result
@@ -80,8 +80,9 @@ def adjoint_binary(out_adj, out, op, lhs, rhs):
 
 @adjoint_ops.register(Reduce, Funsor, Funsor, AssociativeOp, Funsor, frozenset)
 def adjoint_reduce(out_adj, out, op, arg, reduced_vars):
+    assert op in (ops.logaddexp, ops.add)
+
     if op is ops.logaddexp:
-        in_adj = out_adj + (arg * 0.)  # XXX hack to simulate "expand"
+        return {arg: out_adj + (arg * 0.)}  # XXX hack to simulate "expand"
     elif op is ops.add:  # plate!
-        in_adj = out_adj + (out - arg)
-    return {arg: in_adj}
+        return {arg: out_adj + (out - arg)}
--- a/funsor/adjoint.py
+++ b/funsor/adjoint.py
@@ -1,0 +1,87 @@
+from __future__ import absolute_import, division, print_function
+
+from collections import defaultdict
+import torch
+
+import funsor.ops as ops
+from funsor.interpreter import interpretation, reinterpret
+from funsor.ops import AssociativeOp
+from funsor.registry import KeyedRegistry
+from funsor.terms import Binary, Funsor, Number, Reduce, Variable, reflect
+from funsor.torch import Tensor
+
+
+class AdjointTape(object):
+
+    def __init__(self):
+        self.tape = []
+
+    def __call__(self, cls, *args):
+        result = reflect(cls, *args)
+        if cls in (Reduce, Binary, Tensor):
+            self.tape.append((result, cls, args))
+        return result
+
+
+def adjoint(expr, targets, start=Number(0.)):
+
+    adjoint_values = defaultdict(lambda: Number(0.))  # 1 in logspace
+    multiplicities = defaultdict(lambda: 0)
+
+    tape_recorder = AdjointTape()
+    with interpretation(tape_recorder):
+        adjoint_values[reinterpret(expr)] = start
+
+    while tape_recorder.tape:
+        output, fn, inputs = tape_recorder.tape.pop()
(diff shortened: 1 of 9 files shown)


test/test_adjoint.py
```diff
@@ -0,0 +1,126 @@
...
+
+@pytest.mark.parametrize('einsum_impl', [naive_plated_einsum, einsum])
+@pytest.mark.parametrize('equation,plates', PLATED_EINSUM_EXAMPLES)
+@pytest.mark.parametrize('backend', ['pyro.ops.einsum.torch_marginal'])
+def test_plated_einsum_adjoint(einsum_impl, equation, plates, backend):
+    inputs, outputs, sizes, operands, funsor_operands = make_einsum_example(equation)
+
+    with interpretation(reflect):
+        fwd_expr = einsum_impl(equation, *funsor_operands, plates=plates, backend=backend)
+        actuals = adjoint(fwd_expr, funsor_operands)
```
eb8680: Note that in all these tests the call to `adjoint` is inside a `reflect` interpreter.  This is because mixing eager mode and lazy expressions appears to create cycles in the substitution graphs at the points noted above.

funsor/adjoint.py
```diff
@@ -0,0 +1,89 @@
...
+
+    return {lhs: lhs_adj, rhs: rhs_adj}
+
+
+@adjoint_ops.register(Reduce, Funsor, Funsor, AssociativeOp, Funsor, frozenset)
+def adjoint_reduce(out_adj, out, op, arg, reduced_vars):
+    if op is ops.logaddexp:
+        in_adj = out_adj + (arg * 0.)  # XXX hack to simulate "expand"
+    elif op is ops.add:  # plate!
+        in_adj = out_adj + (out + -arg)
```
eb8680: Reusing `arg` here when evaluating `adjoint_reduce` in eager mode sometimes produces cyclical term graphs, causing memory use to explode.

funsor/adjoint.py
```diff
@@ -0,0 +1,89 @@
...
+
+@adjoint_ops.register(Binary, Funsor, Funsor, AssociativeOp, Funsor, Funsor)
+def adjoint_binary(out_adj, out, op, lhs, rhs):
+    assert op is ops.add
+
+    lhs_reduced_vars = frozenset(rhs.inputs) - frozenset(lhs.inputs)
+    lhs_adj = (out_adj + rhs).reduce(ops.logaddexp, lhs_reduced_vars)
+
+    rhs_reduced_vars = frozenset(lhs.inputs) - frozenset(rhs.inputs)
+    rhs_adj = (out_adj + lhs).reduce(ops.logaddexp, rhs_reduced_vars)
```
eb8680: Reusing `lhs` and `rhs` in eager mode here leads to cyclic term graphs

test/test_adjoint.py
```diff
@@ -0,0 +1,176 @@
...
+OPTIMIZED_PLATED_EINSUM_EXAMPLES = [
+    make_plated_hmm_einsum(num_steps, num_obs_plates=b, num_hidden_plates=a)
+    for num_steps in [40, 50]  # range(20, 50, 6)
+    for (a, b) in [(0, 0), (0, 1), (0, 2), (1, 1), (1, 2)]
+]
+
+
+@pytest.mark.parametrize('equation,plates', OPTIMIZED_PLATED_EINSUM_EXAMPLES)
+@pytest.mark.parametrize('backend', ['pyro.ops.einsum.torch_marginal'])
+def test_optimized_plated_einsum_adjoint(equation, plates, backend):
```
eb8680: This test is still quite slow.  The reason seems to be related to substitution as above.

eb8680:
> Can you explain why multiplicities are needed for Variables?

They're needed under the current definition of `adjoint` for targets with multiple parents, I think.  Right now the output of `adjoint` for `Variable` is actually the product (or sum, in log-space) of one copy of the marginal for that variable for each of the ground terms it appears in.  We should refine the semantics of `adjoint` so that this issue is dealt with differently.

> Can you explain why eager mode causes exponential size?

I'm not entirely sure what's going on here - you can see some related failures in Travis.  I think maybe expressions are having subexpressions substituted into themselves in the `adjoint_ops` functions.  Still figuring it out.

funsor/adjoint.py
```diff
@@ -0,0 +1,89 @@
...
+
+    return {lhs: lhs_adj, rhs: rhs_adj}
+
+
+@adjoint_ops.register(Reduce, Funsor, Funsor, AssociativeOp, Funsor, frozenset)
+def adjoint_reduce(out_adj, out, op, arg, reduced_vars):
+    if op is ops.logaddexp:
+        in_adj = out_adj + (arg * 0.)  # XXX hack to simulate "expand"
```
fritzo: It would be nice if we could omit the `+ (arg * 0.)` hack for `.expand()`. I tried to implement a quick `Expand` funsor and it seems gross. My current perspective is thus:

We can either fastidiously track inputs by wrapping funsors in `Constant`/`Expand` funsors, or we can ensure reduction ops have enough metadata to implicitly expand their args. The `.expand()` form will introduce complexity into our pattern matching code, whereas the logic for reduction over absent dims should be isolated to a few `.eager_reduce()` implementations. WDYT?

funsor/ops.py
```diff
@@ -143,6 +143,17 @@ def reciprocal(x):
     raise ValueError("No reciprocal for type {}".format(type(x)))
 
 
+@Op
+def _neg_logprobs(x):
```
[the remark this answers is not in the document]
eb8680: I'll move this logic into `eager_binary_tensor_tensor` and remove these functions, which currently aren't used anyway

funsor/adjoint.py
```diff
@@ -0,0 +1,88 @@
...
+from funsor.torch import Tensor
+
+
+class AdjointTape(object):
+
+    def __init__(self):
+        self.tape = []
+
+    def __call__(self, cls, *args):
+        result = eager(cls, *args)
```
eb8680: Changing this from `reflect` to `eager` seems to have solved the issues I was seeing before.

funsor/adjoint.py
eb8680: Yeah, seems like logic that naturally should live in `eager_reduce`.

PR #65, opened 2019-03-09
Implement a general Delta funsor

### LINKED ISSUE #51: Implement a Delta distribution and generalized Delta funsor
It appears we may be able to represent joint samples as Delta distributions. This will require two new components:
- [x] #49 a `dist.Delta` distribution wrapping a univariate PyTorch-style distribution, and 
- [ ] #65 a general `Delta` funsor similar to `Tensor` and `Gaussian`.

fritzo:
Resolves #51 

`Delta` funsors are the basic data structure to hold adjoint samples. `Delta` is intended to serve as a component in a future `Joint` funsor (which will have multiple `Delta`s and either a `Gaussian` or a `Tensor`).

This `Delta` is a simpler representation than `dist.Delta` in that this `Delta` assumes the bound variable is known, whereas `dist.Delta` may set `value` to an arbitrary Funsor (i.e. not a Variable).

This PR does not implement substitution through bijections with Jacobian.

## Tasks
- [x] implement `Delta` class
- [x] implement `dist.Delta` -> `Delta` conversions
- [x] #66 fix `(x==y).all().log()` and `.expand()` issues
- [x] test substitution
- [x] test reduction
- [x] test binary addition
- [x] test conversion from `dist.Delta`


fritzo:
@eb8680 I think this is ready now. I got confused about `log_density`, but I believe we need to keep track of this in the `Delta` funsors because it scales `eager_subs` but not `eager_reduce`.

funsor/delta.py
```diff
@@ -0,0 +1,109 @@
...
+        log_density = self.log_density.eager_subs(index_part)
+        if value is not None:
+            if isinstance(value, Variable):
+                name = value.name
+            elif isinstance(value, (Number, Tensor)) and isinstance(point, (Number, Tensor)):
+                return (value == point).all().log() + log_density
+            else:
+                # TODO Compute a jacobian, update log_prob, and emit another Delta.
+                raise ValueError('Cannot substitute a {} into a Delta'
+                                 .format(type(value).__name__))
```
eb8680: Would we always want to emit a `Delta` here?

fritzo: Hmm I haven't thought about this, but I guess we could emit `Delta` for bijective functions, or in case of a non-bijective function a mixture of `Deltas`. For example
```py
y = Variable('y', reals())
Delta('x',1)(x=y**2/2) == (Delta('y', -1/2) + Delta(y,+1/2)) / 2
```

PR #66, opened 2019-03-09
Make reduction methods operate over events

fritzo:
This repurposes the reduction ops (`.sum()`, `.all()`, etc.) to reduce over output events, rather than reduce over input variables. This is needed e.g. to implement delta via `(x == y).all()`. For example:
```py
x = Tensor(torch.randn(5,3,3), inputs=OrderedDict(i=bint(5)))
x.inputs  # OrderedDict(i=bint(5))
x.output  # reals(3,3)

# OLD behavior: reduce over the free variable i.
x.sum().inputs  # OrderedDict()
x.sum().output  # reals(3, 3)

# NEW behavior: reduce over the event shape.
x.sum().inputs  # OrderedDict(i=bint(5))
x.sum().output  # reals()
```
## Questions
- [x] Is this a good idea? It seems to me that we should fully expose the array interface for ground values, and use `.reduce(...)` for funsor-specific operations.
- [x] Should we add shorthand for the funsor-specific ops, e.g. `x.f.sum()` where `.f` exposes the funsor-specific interface? Or maybe `x.funsor_sum()`?

## Tasks
- [x] change method behavior
- [x] update old tests
- [x] update examples
- [x] update readme
- [x] add new tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,8 +59,8 @@ def pyro_sample(name, dist, obs=None):
     return value
 
 # ...later during inference...
-log_prob = trace_log_prob.logsumexp()  # collapses delayed variables
-loss = -funsor.eval(log_prob)          # performs variable elimination
+log_prob = trace_log_prob.reduce(logsaddexp)  # collapses delayed variables
+loss = -funsor.eval(log_prob)                 # performs variable elimination
 ```
 See [examples/minipyro.py](examples/minipyro.py) for a more complete example.
 
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ def pyro_sample(name, dist, obs=None):
     return value
 
 # ...later during inference...
-log_prob = trace_log_prob.reduce(logsaddexp)  # collapses delayed variables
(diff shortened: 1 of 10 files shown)


eb8680:
> Is this a good idea?

Yeah, I think it definitely makes sense to draw this distinction.

> Should we add shorthand for the funsor-specific ops

It's a little verbose, but I kind of like having `reduce` as a single entry point to the funsor contraction API to distinguish it from the tensor/array methods.

fritzo:
OK done :smile: @neerajprad note that this changes the reduction tests you may have forked in #58

README.md
```diff
@@ -59,8 +59,8 @@ def pyro_sample(name, dist, obs=None):
     return value
 
 # ...later during inference...
-log_prob = trace_log_prob.logsumexp()  # collapses delayed variables
-loss = -funsor.eval(log_prob)          # performs variable elimination
+log_prob = trace_log_prob.reduce(logsaddexp)  # collapses delayed variables
```
eb8680: typo: `logaddexp`

PR #69, opened 2019-03-10
Implement a Joint normal form funsor

### LINKED ISSUE #68: Implement a Joint funsor as a normal form / state machine
~~Blocked by #51 ~~

It appears that the simplest way to implement an inference strategy is via a `Joint` funsor that keeps track of various `Tensor`, `Gaussian`, and `Delta` factors in a canonical, searchable data structure. This `Joint` object is really a state machine / interpreter disguised as a Funsor. Its denotational semantics are defined by its `.eager_subs()` method, and its operational semantics are defined by the transition operations we call on `Joint` objects during inference:
- `+=` a log prob term with either a lazy or eager value
- `.reduce(ops.logaddexp, -)` out variables when they are no longer live
- `.reduce(ops.sample, -)` variables when they need a ground value (e.g. for control flow)
- `.reduce(ops.add, -)` out plates when they are complete (see log_joint in minipyro.py)

Note that this can be sensibly implemented even before we implement sampling #52 

## Tasks

- [ ] Implement a `Joint` funsor with `Tensor`, `Gaussian`, and various `Delta` terms
- [ ] Implement `+= Tensor` transition
- [ ] Implement `+= Gaussian` transition
- [ ] Implement `+= Delta` transition
- [ ] Implement `.reduce(ops.logaddexp, -)` to eliminate random variables
- [ ] Implement `.reduce(ops.sample, -)` to sample random variables
- [ ] Implement `.reduce(ops.add, -)` to eliminate plates variables
- [ ] Implement `+= Joint` to fuse two state machines
- [ ] Add tests
- [ ] Use `Joint` in examples

fritzo:
Resolves #68 
Based on #65 

This implements a `Joint` funsor incorporating a single `Number`/`Tensor`, a single `Gaussian` (or `Number(0)` if not needed), and zero or more `Delta`s. See #68 for motivation.

Pyro handles distributions of a few different forms. With `num_particles=1` Pyro handles a product-of-deltas (i.e. a single sample at each site):
```py
Finitary(ops.add, (Delta(...), ..., Delta(...)))  # product in log-space
```
The same representation works for vectorized particles, since the reduction doesn't happen until the end, so each `Delta` can carry an extra `particles` input dim.

`TraceEnum_ELBO` in Pyro supports lazy discrete samples, which would be represented as a product of a single Tensor (for all enumerated sites) and finitely many Deltas for sampled sites.
```py
Finitary(ops.add, (Tensor(...), Delta(...), Delta(...), ..., Delta(...)))
```
Note 1. `JointNormalForm` is intended as the interpreter during an optimized variable elimination schedule, therefore we store a single treewidth-sized `Tensor` rather than separate factors.
Note 2. unless we support direct products #70, plates are handled outside of `JointNormalForm`.

This PR builds on the generalized Gaussian #37 which combines a joint distribution among multiple Categorical, Normal, and MultivariateNormal random variables, which can all be sampled lazily. Thus the initial representation in Funsor is
```py
Finitary(ops.add, (Delta(...), ..., Delta(...), Tensor(...), Gaussian(...)))
```

## Tasks
- [x] add tests for `Joint` ops
- [x] add tests for promotion to `Joint`


funsor/gaussian.py
```diff
@@ -254,101 +251,27 @@ def eager_reduce(self, op, reduced_vars):
         return None  # defer to default implementation
 
 
-@eager.register(Binary, Op, Gaussian, Number)
```
fritzo: I've moved `log_density` tracking from `Gaussian` to `Joint`, which simplifies the code here.

test/test_joint.py
```diff
@@ -0,0 +1,152 @@
+from __future__ import absolute_import, division, print_function
```
fritzo: There aren't many tests here, but since this PR moves much of the Gaussian functionality into `Joint`, we can rely on the `Gaussian` unit tests for many of the tests.

funsor/joint.py
```diff
@@ -0,0 +1,239 @@
...
+    if subs:
+        other = other.eager_subs(subs)
+    if joint.gaussian is not Number(0):
+        other = joint.gaussian + other
+    if not isinstance(other, Gaussian):
+        return Joint(joint.deltas, joint.discrete) + other
+    return Joint(joint.deltas, joint.discrete, other)
+
+
+@eager.register(Binary, AddOp, Joint, Binary)
```
eb8680: How would this case come up in `eager`?

funsor/gaussian.py
eb8680: This looks great now!

funsor/joint.py
fritzo: Oh good point, it wouldn't come up. I've removed this case.

fritzo:
Thanks for reviewing @eb8680 !

Issue #61, opened 2019-03-08
Add backend dispatchers for operations in  funsor.ops


neerajprad:
It seems to me that the simplest way to support multiple backends in `funsor.ops` would be to use multiple dispatch as follows:

```python
@singledispatch
def abs(x):
    return _builtin_abs(x)

@abs.register(torch.Tensor)
def abs_torch(x):
    return x.abs()

@abs.register(np.ndarray)
def abs_np(x):
    return np.abs(x)
```

What do you think? Will that cause any inadvertent complications otherwise?

eb8680:
This seems OK as long as `ops` remains restricted to the subset of functionality of each backend for which the APIs already overlap *exactly* - we shouldn't be in the business of codifying a generic tensor API, that's Numpy's job now.

neerajprad:
This only gives us flexibility to do things like clamping or have some custom defined behavior, when it might be needed. When the API is exactly the same e.g.  in the case of `abs` or `sqrt` we can just return `np.abs(x)` where x could be a tensor, `ndarray` or a python numeric type and numpy would dispatch to the correct implementation.

fritzo:
This seems like a good idea, also using `multipledispatch` for binaries and an `(object, object)` default.

Note that #55 and #63 implement an `Op` wrapper, and you may need to add a `.register` and/or `.dispatch` property to get this to work correctly with singledispatch and multipledispatch. Sorry for the added complexity!

fritzo:
Another thing I like about this approach is that it would allow us to avoid importing `torch` and `numpy` in the low-level `ops` module, since we could move backend registrations to the respective backend modules. More generally I think we should be able to avoid importing both numpy and torch in any one module, so they would eventually be optional dependencies, as in opt_einsum.

eb8680:
Closed by #74

PR #74, opened 2019-03-12
 Separate out ops implementations based on backend

neerajprad:
As discussed in #61, this refactors `funsor.ops` to move any torch specific ops to the `torch` module. Once this is merged, we can implement the corresponding operations for the `numpy` backend. The Op class now inherits from `Dispatcher`.

Some questions:
 - The default dispatcher for object types currently defaults to numpy for dispatching. I believe that we can save a small amount of time if we were to dispatch directly via `obj.sqrt()` instead of `np.sqrt(obj)`. Is this worth doing, i.e. separating out the dispatchers for Number and object type?
 - In `torch.py` file, for binary operators with distinct types, I had to keep the other type as `object` (consistent with our earlier implementation). Am I correct in understanding that for all operations - we expect raw (i.e. non Funsor wrapped) types except for `funsor.Number` where we can either get a wrapper `Number` or a python numeric type? My guess is yes based on the tests, but just wondering if this is a conscious decision.

### DIFF
--- a/funsor/ops.py
+++ b/funsor/ops.py
@@ -128,22 +128,12 @@ def logaddexp(x, y):
 def safesub(x, y):
     if isinstance(y, Number):
         return sub(x, y)
-    assert isinstance(y, torch.Tensor)
-    try:
-        return x + -y.clamp(max=torch.finfo(y.dtype).max)
-    except TypeError:
-        return x + -y.clamp(max=torch.iinfo(y.dtype).max)
 
 
 @Op
 def safediv(x, y):
     if isinstance(y, Number):
         return truediv(x, y)
-    assert isinstance(y, torch.Tensor)
-    try:
-        return x * y.reciprocal().clamp(max=torch.finfo(y.dtype).max)
-    except TypeError:
-        return x * y.reciprocal().clamp(max=torch.iinfo(y.dtype).max)
 
 
 # just a placeholder
--- a/funsor/ops.py
+++ b/funsor/ops.py
@@ -1,14 +1,11 @@
 from __future__ import absolute_import, division, print_function
 
-import functools
-from collections import namedtuple
 from numbers import Number
 import operator
 
 import numpy as np
+from multipledispatch import Dispatcher
 
-from funsor.registry import KeyedRegistry
-from funsor.six import getargspec
 
 _builtin_abs = abs
 _builtin_max = max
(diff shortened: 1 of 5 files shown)


neerajprad:
The tests pass on my machine, but its failing on travis. I'm looking into that.

funsor/ops.py
```diff
@@ -15,17 +17,30 @@
...
+
     def __new__(cls, fn):
         result = super(Op, cls).__new__(cls, fn)
+        # register as default op impl for object args
         functools.update_wrapper(result, fn)
+        args, vargs, kwargs, default = getargspec(fn)
+        assert not vargs
+        assert not kwargs
+        default_signature = (object,) * len(args)
+        cls.keyed_registry.register_impl(result, lambda *args: fn(*args), *default_signature)
```
fritzo: It seems simpler to use a `Dispatcher` object rather than a `KeyedRegistry`, something like
```py
def __new__(cls, fn):
    if not isintance(fn, multipledispatch.Dispatcher):
        old_fn, fn = fn, multipledispatch.Dispatcher(fn.__name__)
        args, vargs, kwargs, default = getargspec(fn)
        default_signature = (object,) * len(args)
        fn.register(*default_signature)(old_fn)
    result = super(Op, cls).__new__(cls, fn)
    functools.update_wrapper(result, fn)
    return result

@property
def register(self):
    return self.fn.register
```

fritzo: Actually, why don't we simply let `Op` inherit from `Dispatcher` and bake your nice `default_signature` logic into `Op.__init__()`? I don't think there was any particular reason to keep `Op` a namedtuple.

neerajprad: That really cleaned things up, thanks for the suggestion!

test/test_torch.py
```diff
@@ -190,7 +190,7 @@ def unary_eval(symbol, x):
 
 @pytest.mark.parametrize('dims', [(), ('a',), ('a', 'b')])
 @pytest.mark.parametrize('symbol', [
-    '~', '-', 'abs', 'sqrt', 'exp', 'log', 'log1p',
+    'abs', 'sqrt', 'exp', 'log', 'log1p',
```
fritzo: Why aren't we testing `~` and `-` any more?

funsor/torch.py
```diff
@@ -414,7 +414,114 @@ def torch_einsum(equation, *operands):
...
+
+@ops.pow.register(object, torch.Tensor)
+def _pow(x, y):
+    result = x ** y
+    # work around shape bug https://github.com/pytorch/pytorch/issues/16685
+    return result.reshape(y.shape)
+
+
+@ops.pow.register(torch.Tensor, torch.Tensor)
+@ops.pow.register(torch.Tensor, object)
```
fritzo: Note you can use notation
```py
@ops.pow.register(torch.Tensor, (object, torch.Tensor))
```

funsor/registry.py
```diff
@@ -12,21 +12,23 @@ def __init__(self, default=None):
...
         # supported by multipledipatch (which returns a Dispatch object rather
         # than the original function).
         def decorator(fn):
-            register(*types)(fn)
+            self.register_impl(key, fn, *types)
             return fn
 
         return decorator
 
+    def register_impl(self, key, impl, *types):
```
fritzo: Are changes to this file still needed?

neerajprad: Will remove, this shouldn't be needed.

test/test_torch.py
neerajprad: My bad, I was debugging something, and forgot to reinstate this.

PR #75, opened 2019-03-14
Implement monte carlo .sample() methods

fritzo:
Addresses #52 

This implements a low-level `Funsor.sample()` method that will be used in Monte Carlo approximation. This method returns a funsor of the same shape as the original funsor, indeed an unbiased estimator of the original funsor, in value and all derivatives. Samples are encoded as `Delta` distributions, and differentiability is achieved by either reparametrization (for `Gaussian`) or multiplying by Dice factors (for discrete samples of `Tensor`). Sampling from a `Delta` is a no-op. Multiple samples can be drawn by passing in a `sample_inputs`, analogous to the `sample_shape` of torch.distributions.

Note that `Distribution` objects do not implement `.sample()` because the only way we could draw samples would be with materialized tensors, and all current `Distribution` objects eagerly convert to other funsors (`Categorical` -> `Tensor`, `Normal` -> `Gaussian`, ...), and it is those other funsors that implement `.sample()` methods.

This PR also deletes the unused `ops.sample` and `ops.marginal` because we seem to be implementing that functionality elsewhere.

## Questions

- [x] I would have preferred the multiple sample result to be wrapped in a lazy `Reduce()` so as to maintain the property of unbiased estimation, however our current eager `Reduce()` semantics would unbind the resulting funsor and convert to a deeply nested `Binary()`. Suggestions welcome.

## Tasks
- [x] Implement `Tensor.sample()`
- [x] Implement `Gaussian.sample()`
- [x] Implement `Delta.sample()`
- [x] Implement `Joint.sample()`
- [x] test `Tensor.sample()` shapes
- [x] test `Gaussian.sample()` shapes
- [x] test `Joint.sample()` shapes
- [x] test `Tensor.sample()` distribution
- [x] test `Tensor.sample()` gradients
- ~~test `Gaussian.sample()` distribution~~ blocked by `Integrate`
- ~~test `Gaussian.sample()` gradient~~ blocked by `Integrate`

### DIFF
--- a/funsor/torch.py
+++ b/funsor/torch.py
@@ -5,8 +5,10 @@ from collections import OrderedDict
 
 import torch
 from six import add_metaclass, integer_types
+from six.moves import reduce
 
 import funsor.ops as ops
+from funsor.delta import Delta
 from funsor.domains import Domain, bint, find_domain, reals
 from funsor.ops import Op
 from funsor.six import getargspec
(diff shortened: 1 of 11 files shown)


test/test_samplers.py
```diff
@@ -0,0 +1,191 @@
...
+    sampled_vars = frozenset(event_inputs)
+    p = random_tensor(be_inputs)
+
+    q = p.sample(sampled_vars, sample_inputs) - ops.log(num_samples)
+    mq = materialize(q).reduce(ops.logaddexp, 'n')
+    mq = mq.align(tuple(p.inputs))
+    assert_close(mq, p, atol=0.1, rtol=None)
+
+
+@pytest.mark.skip(reason='infinite loop')
```
fritzo: @eb8680 Could we chat about this test, which is the final task of this PR. The test exercises the `(p.exp() * f).sum()` operation we discussed yesterday, but I haven't figured out how to eagerly execute that correctly.

funsor/torch.py
```diff
@@ -234,6 +237,66 @@ def eager_reduce(self, op, reduced_vars):
             return Tensor(data, inputs, self.dtype)
         return super(Tensor, self).eager_reduce(op, reduced_vars)
 
+    def sample(self, sampled_vars, sample_inputs=None):
```
fritzo: I'm happy with how this function turned out.

test/test_samplers.py
eb8680: Yeah let's go through this tomorrow

funsor/torch.py
```diff
@@ -234,6 +237,67 @@ def eager_reduce(self, op, reduced_vars):
...
+        # Derivation: Let f be a nonnormalized distribution (a funsor), and
+        #   consider operations in linear space (source code is in log space).
+        #   Let x0 ~ f/|f| be a monte carlo sample from a normalized f/|f|.
+        #                              f(x0) / |f|      # dice numerator
+        #   Let g = delta(x=x0) |f| -----------------
+        #                           detach(f(x0)/|f|)   # dice denominator
+        #                       |detach(f)| f(x0)
+        #         = delta(x=x0) -----------------  be a dice approximation of f.
+        #                         detach(f(x0))
+        #   Then g is an unbiased estimator of f in value and all derivatives.
```
eb8680: Does this automatically do the maximum amount of gradient estimator Rao-Blackwellizing, like `TraceGraph_ELBO`?

test/test_samplers.py
```diff
@@ -0,0 +1,207 @@
...
+        assert p_data.shape == mq_data.shape
+        probe = torch.randn(p_data.shape)
+        expected = grad((p_data.exp() * probe).sum(), [p.data])[0]
+        actual = grad((mq_data.exp() * probe).sum(), [p.data])[0]
+        assert_close(actual, expected, atol=0.1, rtol=None)
+
+
+# This is a stub for a future PR.
+def Integrate(log_measure, integrand, reduced_vars):
+    pytest.xfail(reason='Integrate is not implemented')
```
eb8680: For #77

funsor/torch.py
fritzo: Hmm, I would expect Rao-Blackwellization to try to avoid calling `.sample()` at all and instead enumerate. Can you explain offline?

PR #77, opened 2019-03-15
Implement Contract term

### LINKED ISSUE #71: Implement a Contract op as a fused Reduce(Binary or Finitary)
Funsor currently implements variable elimination entirely with sum and product ops, but for efficiency these should be fused into a GEMM-like Contract op for at least binary contractions.

We could support finitary contractions, but binary contractions are very basic: they are GEMM-like ops, inner products in Hilbert spaces, the basic tensor contraction of cutensor, and the basic op in an optimized variable elimination schedule.

eb8680:
Resolves #71 and will replace #54 's [Expectation](https://github.com/pyro-ppl/funsor/blob/4e75094d56/funsor/expectation.py) class.

This PR adds a `Contract(lhs, rhs, reduced_vars)` term for representing expectations of complicated functions wrt joint measures like guide programs.

Tasks:
- [x] Fix product measure tests against einsum
- [x] Turn current `Integrate` into `Contract`
- [x] Add `reduced_vars` argument for partial integration
- [x] Split `funsor.contract` module renaming changes into separate PR to simplify the diff
- [x] Rename `Contract` args to `lhs`, `rhs`
- [x] Fix general tests against einsum
- [x] implement `eager.register(Contract, Tensor, Tensor, frozenset)` using einsum (probably using [interleaved input](https://optimized-einsum.readthedocs.io/en/latest/input_format.html#interleaved-input))
- [x] Implement `eager_subs`

Triaged (will handle in followup PRs so we can get this merged):
- Make new `Integrate` that assumes the `measure` is in log-space and `integrand` is in linear space
- Handle `Joint` either in `optimize(Integrate)` or by decomposing into a `Finitary(ops.add, ...)`
- Look into making iterated expectation less conservative by exploiting directionality
- Make repeated connected components analysis in `contract_finitary_ground` cheaper


funsor/integrate.py
```diff
@@ -0,0 +1,152 @@
...
+
+import funsor.ops as ops
+from funsor.distributions import Gaussian, Delta
+from funsor.interpreter import interpretation, reinterpret
+from funsor.optimizer import Finitary, optimize
+from funsor.terms import Funsor, Number, Reduce, Variable, eager, reflect
+from funsor.torch import Tensor
+
+
+GROUND_TERMS = (Tensor, Gaussian, Delta, Number, Variable)
```
fritzo: We will want to either include `Joint` here, or split `Joint` into a `Finitary(ops.add, ...)` whose terms are ground.

funsor/integrate.py
```diff
@@ -0,0 +1,152 @@
...
+        if frozenset(measure.inputs) & frozenset(operand.inputs):
+            new_operands.append(operand)
+        else:
+            constants.append(operand)
+    return tuple(constants), tuple(new_operands)
+
+
+class Integrate(Funsor):
+
+    def __init__(self, measure, integrand):
```
fritzo: I'm pretty sure we'll want to specify `reduced_vars` so we can compute a batch of integrals. I think this will even be needed for iterated expectation with overlapping variables.

funsor/contract.py
```diff
@@ -0,0 +1,242 @@
...
+    result = _simplify_contract(lhs, rhs, reduced_vars)
+    if result is not None:
+        return result
+
+    if reduced_vars:
+        out_vars = (frozenset(lhs.inputs) | frozenset(rhs.inputs)) - reduced_vars
+    else:
+        out_vars = {}
+
+    out_inputs = OrderedDict([
```
fritzo: This is nondeterministic (the ordering depends on frozenset details). Elsewhere in Funsor I've tried to preserve the ordering of inputs. Would you mind changing the computation here to something like:
```py
otu_inputs = OrderedDict((k, d) for t in (lhs, rhs)
                         for k, d in t.inputs.items() if k not in reduced_vars)
out_vars = frozenset(inputs)
```

funsor/contract.py
```diff
@@ -0,0 +1,242 @@
...
+        subs = tuple((k, v) for k, v in subs if k not in self.reduced_vars)
+        if not any(k in self.inputs for k, v in subs):
+            return self
+        if not all(self.reduced_vars.isdisjoint(v.inputs) for k, v in subs):
+            raise NotImplementedError('TODO alpha-convert to avoid conflict')
+        return Contract(self.lhs.eager_subs(subs), self.rhs.eager_subs(subs),
+                        self.reduced_vars)
+
+
+@optimize.register(Contract, ATOMS[1:], ATOMS[1:], frozenset)
```
fritzo: Do you also need to handle (`Tensor`, non-`Tensor`) cases?

funsor/contract.py
```diff
@@ -0,0 +1,242 @@
...
+        remaining_lhs = None
+
+    return root_lhs, remaining_lhs
+
+
+def _simplify_contract(lhs, rhs, reduced_vars):
+    """
+    Reduce free variables that do not appear explicitly in the lhs
+    """
+    meas_vars = frozenset(lhs.inputs)
```
fritzo: nit: could we rename these to `lhs_vars` and `rhs_vars`

funsor/contract.py
```diff
@@ -0,0 +1,242 @@
...
+from funsor.optimizer import Finitary, optimize
+from funsor.sum_product import _partition
+from funsor.terms import Funsor, Number, Reduce, Variable, eager
+from funsor.torch import Tensor
+
+
+ATOMS = (Tensor, Gaussian, Delta, Number, Variable)
+
+
+def _make_base_lhs(arg, reduced_vars, normalized=False):
```
fritzo: This should no longer be used, right?

funsor/contract.py
```diff
@@ -0,0 +1,242 @@
...
+    )
+
+
+@optimize.register(Contract, Funsor, Reduce, frozenset)
+def contract_reduce(lhs, rhs, reduced_vars):
+    result = _simplify_contract(lhs, rhs, reduced_vars)
+    if result is not None:
+        return result
+
+    # XXX should we be doing this conversion at all given that Reduce
```
fritzo: The use of `_make_base_lhs()` here seems weird to me, I'll have to think about it...

funsor/contract.py
```diff
@@ -0,0 +1,242 @@
...
+import funsor.ops as ops
+from funsor.distributions import Gaussian, Delta
+from funsor.interpreter import interpretation, reinterpret
+from funsor.optimizer import Finitary, optimize
+from funsor.sum_product import _partition
+from funsor.terms import Funsor, Number, Reduce, Variable, eager
+from funsor.torch import Tensor
+
+
+ATOMS = (Tensor, Gaussian, Delta, Number, Variable)
```
fritzo: Should we add `Joint` to this list, or are you decomposing that into a sum?

fritzo: Ah I see this is triaged. Would you mind adding a #TODO in the code to help me remember?

funsor/contract.py
```diff
@@ -0,0 +1,242 @@
...
+    return None
+
+
+@optimize.register(Contract, ATOMS, Finitary, frozenset)
+def contract_ground_finitary(lhs, rhs, reduced_vars):
+    result = _simplify_contract(lhs, rhs, reduced_vars)
+    if result is not None:
+        return result
+
+    # exploit linearity of integration
```
fritzo: Should we handle lhs similarly, in case `lhs.op is ops.add`?

funsor/contract.py
```diff
@@ -0,0 +1,242 @@
...
+    if result is not None:
+        return result
+
+    # recursively apply law of iterated expectation
+    assert len(lhs.operands) > 1, "Finitary with one operand should have been passed through"
+    if lhs.op is ops.mul:
+        root_lhs, remaining_lhs = _order_lhss(lhs, reduced_vars)
+        if remaining_lhs is not None:
+            inner = Contract(remaining_lhs, rhs,
+                             reduced_vars & frozenset(remaining_lhs.inputs))
```
fritzo: Hmm, should this be
```py
reduced_vars.intersection(remaining.lhs.inputs).difference(root_lhs.inputs)
```
I would have guessed that a test like `"a,a,a->"` would have caught this 😕

funsor/contract.py
```diff
@@ -0,0 +1,242 @@
...
+    return None
+
+
+@optimize.register(Contract, Finitary, (Finitary,) + ATOMS, frozenset)
+def contract_finitary_ground(lhs, rhs, reduced_vars):
+    result = _simplify_contract(lhs, rhs, reduced_vars)
+    if result is not None:
+        return result
+
+    # recursively apply law of iterated expectation
```
fritzo: Should we handle rhs similarly since `Contract` is symmetric in lhs, rhs?

funsor/contract.py
eb8680: `_order_lhss` guarantees that this intersection is empty

funsor/contract.py
eb8680: I removed this rule

funsor/contract.py
eb8680: Moved to `einsum` since I'm only using it for testing

fritzo:
Thanks for simplifying! I look forward to seeing the `Integrate` implementation.

funsor/contract.py
```diff
@@ -0,0 +1,141 @@
...
+ATOMS = (Tensor, Gaussian, Delta, Number, Variable)
+
+
+def _order_lhss(lhs, reduced_vars):
+    assert isinstance(lhs, Finitary)
+
+    components = _partition(lhs.operands, reduced_vars)
+    root_lhs = Finitary(ops.mul, tuple(components[0][0]))
+    if len(components) > 1:
+        remaining_lhs = Finitary(ops.mul, tuple(t for c in components[1:] for t in c[0]))
```
eb8680: Note that since these are undirected, the optimizer should be using cost to decide which term is the root

Issue #72, opened 2019-03-10
Implement recognition of affine transforms


fritzo:
To perform full exact inference of Gaussian distributions, we will need some ability to recognize affine transforms. This might either rely on autograd, or could simply pattern match `+`, `*`, and einsum patterns.

## Tasks

- [x] Handle `Binary` and `Variable` patterns #116 
- [x] Rewrite `Normal(Affine, ...)` to an appropriate `Gaussian` #119 
- [x] Handle `Contract` and `einsum` #157 
- [x] Recognize affine funsors including `Contraction` and `Einsum` #249 
- [x] Recognize `GetitemOp` and `ReshapeOp` as a linear transformations #249 
- [x] Rewrite `MultivariateNormal(Affine, ...)` to an appropriate `Gaussian` #245
- [x] Fix `funsor.affine.is_affine()` to be sound but incomplete #282
- [ ] Rewrite `Gaussian(x=affine)` as a `Gaussian`. In three steps: #284, #285, #286

fehiepsi:
Thanks for keeping tracks of those tasks! It is convenient for me to understand what is going on.

fritzo:
@fehiepsi Do you want to try working on the task
> Rewrite `Gaussian(x=affine)` as a `Gaussian`

following the ideas in #245 ? If not, I'm happy to do this part to unblock our first release. If so, it might be a nice intro to Funsor and give us insight into how to incorporate structured Gamma priors in Funsor, generalizing your Pyro PR https://github.com/pyro-ppl/pyro/pull/2025 .

fehiepsi:
Hi @fritzo, could you do it instead? It is better for me to follow-up (though I want to try), given my lacking knowledge of Funsor. I am happy to review and learn from you.

fritzo:
@fehiepsi sounds good, I'll add you as reviewer.

PR #78, opened 2019-03-18
Support for pattern matching with the unification library

eb8680:
I was looking into options for pattern matching and came across the [`unification`](https://github.com/mrocklin/unification/) library, and it turns out it's super easy to add support for Funsor expressions thanks to the magic of `multipledispatch`.  We should be able to use this as a substrate to implement #72.

Tasks:
- [x] Add a couple tests

### DIFF
--- a/funsor/pattern.py
+++ b/funsor/pattern.py
@@ -1,0 +1,96 @@
+from __future__ import absolute_import, division, print_function
+
+# Design notes:
+# Component 1: Recognition interpreter generator
+# Component 2: Rewrites of recognized patterns
+# Component 3: Constant folding/CSE? Just use eager?
+
+import funsor.ops as ops
+from funsor.interpreter import interpretation, reinterpret
+from funsor.registry import KeyedRegistry
+from funsor.terms import Binary, Funsor, reflect
+
+
+class Pattern(Funsor):
+    """
+    Recognized pattern term
+    """
+    pass
+
+
+class Free(Pattern):
+    pass
+
+
+class Choice(Pattern):
+    pass
+
+
+def make_recognizer(pattern):
+    # make a recognizer for a Funsor pattern, rewriting the first match
+
+    # semiring matcher: sum = any_match, product = all_match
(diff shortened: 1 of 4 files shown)


fritzo:
I wonder if this can help some of our logic to reason about `Delta` funsors in `Joint`.

funsor/pattern.py
```diff
@@ -0,0 +1,36 @@
...
+
+from unification import unify
+from unification.variable import isvar
+
+import funsor.ops as ops
+from funsor.interpreter import interpretation
+from funsor.terms import Binary, Funsor, Variable, lazy
+
+
+@lazy.register(Binary, ops.Op, Funsor, Funsor)
```
fritzo: As we discussed, it would be safest to implement this as a standalone interpretation. Even further, it would be good to prevent construction of new funsors under this interpretation:
```py
def compare(cls, *args):
    if cls is not Binary:
        raise ValueError
    op, lhs, rhs = args
    if op is ops.eq:
        return lhs is rhs
    if op is ops.neq:
        return lhs is not rhs
    raise ValueError
```

test/test_pattern.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def test_match_binary():
+    with interpretation(lazy):
+        pattern = Variable('a', reals()) + Number(2.) * Variable('b', reals())
+        expr = Number(1.) + Number(2.) * (Number(3.) - Number(4.))
+
+    @match_vars(pattern)
+    def expand_2_vars(a, b):
+        return a + b + b
```
eb8680: This shows how `match_vars` will be helpful for interpreters that use fancier patterns

funsor/pattern.py
```diff
@@ -0,0 +1,69 @@
+from __future__ import absolute_import, division, print_function
+
+import functools
+
+from unification import unify
+from unification.variable import isvar
+from unification.match import match, Dispatcher
```
fritzo: nit: avoid overwriting the global variable `match`
```py
import unification

match_vars = functools.partial(unification.match, ...)
match = functools.partial(unification.match, ...)
```

PR #91, opened 2019-03-21
Refactor contract dependencies

fritzo:
This refactors `@...register(Contract, ...)` so that:
1. Registration relies on the funsor class hierarchy rather than an explicit `ATOMS` list.
2. Custom registrations (e.g. for `Tensor`) live in the custom files, rather than `contract.py`.

## Motivation

`Contract` (and later `Integrate`) appear to be more basic than custom implementations like `Tensor`, `Array`, `Delta`, `Gaussian`, `Joint`. Therefore it seems cleaner to import `Contract` in custom code than to import all custom objects in contract.py.

### DIFF
--- a/funsor/contract.py
+++ b/funsor/contract.py
@@ -88,13 +88,13 @@ def contract_funsor_funsor(lhs, rhs, reduced_vars):
 
 @optimize.register(Contract, Funsor, Finitary, frozenset)
 @contractor
-def contract_ground_finitary(lhs, rhs, reduced_vars):
+def contract_funsor_finitary(lhs, rhs, reduced_vars):
     return Contract(rhs, lhs, reduced_vars)
 
 
 @optimize.register(Contract, Finitary, (Finitary, Funsor), frozenset)
 @contractor
-def contract_finitary_ground(lhs, rhs, reduced_vars):
+def contract_finitary_funsor(lhs, rhs, reduced_vars):
     # exploit linearity of contraction
     if lhs.op is ops.add:
         return Finitary(
--- a/funsor/contract.py
+++ b/funsor/contract.py
@@ -79,22 +79,27 @@ class Contract(Funsor):
                         self.reduced_vars)
 
 
-@optimize.register(Contract, Funsor, Funsor, frozenset)
 @eager.register(Contract, Funsor, Funsor, frozenset)
 @contractor
-def contract_funsor_funsor(lhs, rhs, reduced_vars):
+def eager_contract(lhs, rhs, reduced_vars):
     return (lhs * rhs).reduce(ops.add, reduced_vars)
+
+
+@optimize.register(Contract, Funsor, Funsor, frozenset)
+@contractor
+def optmize_contract(lhs, rhs, reduced_vars):
+    return None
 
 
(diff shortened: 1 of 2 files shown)


funsor/torch.py
```diff
@@ -352,6 +355,22 @@ def eager_binary_tensor_tensor(op, lhs, rhs):
...
+def eager_contract(lhs, rhs, reduced_vars):
+    inputs = OrderedDict((k, d) for t in (lhs, rhs)
+                         for k, d in t.inputs.items() if k not in reduced_vars)
+    data = opt_einsum.contract(lhs.data, list(lhs.inputs),
+                               rhs.data, list(rhs.inputs),
+                               list(inputs), backend="torch")
+    return Tensor(data, inputs, rhs.dtype)
+
+
+@optimize.register(Contract, Tensor, Tensor, frozenset)
```
fritzo: @eb8680 I believe this accomplishes what was previously accomplished via `ATOMS[1:]`.

funsor/contract.py
```diff
@@ -77,50 +79,22 @@ def eager_subs(self, subs):
...
-@optimize.register(Contract, ATOMS[1:], ATOMS, frozenset)
-@optimize.register(Contract, ATOMS, ATOMS[1:], frozenset)
-@eager.register(Contract, ATOMS[1:], ATOMS, frozenset)
-@eager.register(Contract, ATOMS, ATOMS[1:], frozenset)
-def contract_ground_ground(lhs, rhs, reduced_vars):
-    result = _simplify_contract(lhs, rhs, reduced_vars)
-    if result is not None:
-        return result
-
+@optimize.register(Contract, Funsor, Funsor, frozenset)
```
fritzo: This now applies to `Binary`, `Unary`, `Reduce` etc. Should I add a `COMPOUNDS` list with a reflect op as I've done for `Tensor` below?

funsor/contract.py
```diff
@@ -45,11 +41,17 @@ def _simplify_contract(lhs, rhs, reduced_vars):
...
         progress = True
-
     if progress:
         return Contract(lhs, rhs, reduced_vars)
 
-    return None
+    return fn(lhs, rhs, reduced_vars)
+
+
+def contractor(fn):
```
fritzo: I've made this a little easier to use, since we're now importing it in other files.

funsor/contract.py
eb8680: I would say to just remove that line since `optimize` defaults to reflection

funsor/contract.py
```diff
@@ -77,50 +79,27 @@ def eager_subs(self, subs):
...
-    return Tensor(
-        opt_einsum.contract(lhs.data, list(lhs.inputs.keys()),
-                            rhs.data, list(rhs.inputs.keys()),
-                            list(out_inputs.keys()), backend="torch"),
-        out_inputs
-    )
-
+@optimize.register(Contract, Funsor, Funsor, frozenset)
+@contractor
+def optmize_contract(lhs, rhs, reduced_vars):
```
eb8680: nit: `optmize` -> `optimize`

PR #95, opened 2019-03-24
Add a VAE example using the monte_carlo interpretation

### LINKED ISSUE #11: Add simple svi example
The elbo implementation in [funsor/minipyro.py](https://github.com/pyro-ppl/funsor/blob/master/funsor/minipyro.py) is currently broken.

## Outline
- [ ] add unit tests for elbo
- [ ] add example of variational inference

fritzo:
Resolves #11 
Blocked by #97 (`Lambda` and `Independent`)

This adds a VAE example similar to PyTorch's example.

I've added a few new reduction rules to make this example work.

## Tested

- [x] added to Makefile


examples/vae.py
```diff
@@ -0,0 +1,110 @@
...
+    @funsor.interpreter.interpretation(funsor.montecarlo.monte_carlo)
+    def loss_function(data, subsample_scale):
+        # Lazily sample from the guide.
+        loc, scale = encode(data)
+        q = funsor.Independent(
+            dist.Normal(loc['i'], scale['i'], value='z'),
+            'z', 'i')
+
+        # Evaluate the model likelihood at the lazy value z.
+        probs = decode('z')
```
eb8680: nit: we might want to use a `Variable` object here (and in any similar situations) as it's not immediately clear that `decode` is a `Funsor`.  We can fix example idioms in followup PRs though.

PR #97, opened 2019-03-25
Add Lambda and Independent funsors

fritzo:
This implements two new funsors to convert between "event dims" and "batch dims":
```
              _[i] :        (i->j->R)  -->  i -> (j -> R)
       Lambda(i,_) :      i -> (j->R)  -->  (i->j->R)
Independent(_,_,i) : i -> (j->R) -> x  -->  (i->j->R) -> x
```
In funsor.distributions, `Independent` serves the purpose of `.to_event()`, together with a rule to convert independent `Delta`s; this rule uses `Lambda` internally.

This is motivated by the diagonal normal sample statement in examples/vae.py working #95 

## Tested
- [x] unit tests for `Lambda`
- [x] unit tests for `Independent`
- [x] sampler tests for `Independent` used as `.to_event()`


test/test_torch.py
```diff
@@ -379,6 +379,26 @@ def test_getitem_tensor():
...
+    assert Lambda(i, y) is x
+
+
+def test_lambda_subs():
+    x = Tensor(torch.randn(2))
+    y = Tensor(torch.randn(2))
+    z = Variable('z', reals())
+    i = Variable('i', bint(2))
+
+    actual = (x + Lambda(i, z))(z=y[i])  # FIXME this doesn't work
```
fritzo: This is what the VAE example needs, but it's not quite right.

PR #98, opened 2019-03-25
Resurrect lazy Subs funsor (again)

fritzo:
This implements a lazy `Subs` funsor.

## Motivation

It appears this is the cleanest way to implement transformed distributions and factorial distributions. For example, it appears this is needed to implement an `Independent` distribution for examples/vae.py in #95 .

## Tests

### DIFF
--- a/funsor/__init__.py
+++ b/funsor/__init__.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import, division, print_function
 
 from funsor.domains import Domain, bint, find_domain, reals
+from funsor.integrate import Integrate
 from funsor.interpreter import reinterpret
 from funsor.terms import Funsor, Number, Variable, of_shape, to_data, to_funsor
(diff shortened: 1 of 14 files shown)


funsor/optimizer.py
```diff
@@ -8,7 +8,7 @@
 from funsor.domains import find_domain
 from funsor.interpreter import dispatched_interpretation, interpretation, reinterpret
 from funsor.ops import DISTRIBUTIVE_OPS, AssociativeOp
-from funsor.terms import Binary, Funsor, Reduce, eager, reflect
+from funsor.terms import Binary, Funsor, Reduce, Subs, eager, lazy
```
fritzo: @eb8680 I've changed this from `reflect` to `lazy` in this file since that preserves current behavior. Let me know if you want me to change to `reflect`.

PR #99, opened 2019-03-26
Resurrect lazy Subs funsor (again)

fritzo:
Replaces #98 

This implements a lazy `Subs` funsor.

## Motivation

It appears this is the cleanest way to implement transformed distributions and factorial distributions. For example, it appears this is needed to implement an `Independent` distribution for examples/vae.py in #95 .

## Tests


fritzo:
@eb8680 could you take another look after my git snafu in #98.

PR #100, opened 2019-03-26
Add a working minipyro with tests

fritzo:
Resolves #57 
Partly pair coded with @eb8680 

This implements:
- minimal version of minipyro
- tests adapted from Pyro's test_valid_models.py

The minipyro in this PR uses a stub for `Expectation` #109, and remaining issues are blocked either by expectation or by missing patterns for `Joint`. Therefore I think it's safe to merge this frontend and move on to more backend work.

## Tested
- [x] added a test_minipyro.py that runs with multiple backends
- [x] ran `pyro/examples/minipyro.py --backend=funsor` (until blocked by `Expectation`)

PR #103, opened 2019-03-27
Implement basic transformed distributions

fritzo:
Addresses issues with #78, #100

This PR starts to implement transformed distributions as lazy substitution.
1. `Gaussian` now supports lazy substitution, e.g. `LogNormal(m,s)=Normal(m,s)(value=exp(value))`. This is implemented as new logic in `Gaussian.eager_subs`.
2. Sampling from transformed distributions is enabled by a new method `Subs.unscaled_sample()` and new logic in `Delta.eager_subs()`. The `Delta.eager_subs()` logic relies on a new `funsor.delta.solve()` function that computes inverses. Currently inverses are implemented only for `ops.exp()` and `ops.log()` analogous to `ExpTransform` and `ExpTransform().inv`; future inverse can be registered via `@solve.register(...)`.

This PR does not make changes to `Joint` that will be needed to perform inference with transformed distributions. We defer that restructuring of `Joint` to a future PR.

## Questions

- [x] Is `Delta.eager_subs()` the right place to insert `log_abs_det_jacobian`?

## Tests
- [x] tests for `Delta` substitution
- [x] tests for `Gaussian` substitution
- [x] tests for sampling from a transformed `Gaussian`
- [x] test for a `LogNormal` density

## Triaged
- ~~test  for `LogNormal` sampler~~
    I've added an xfailing test. `LogNormal(...).sample()` is blocked on `Joint` being able to incorporate transformed distributions, which I'll implement in a follow-up PR.

### DIFF
--- a/funsor/delta.py
+++ b/funsor/delta.py
@@ -5,10 +5,12 @@ from collections import OrderedDict
 from six import add_metaclass
 
 import funsor.ops as ops
-from funsor.domains import reals
+from funsor.domains import Domain, reals
 from funsor.integrate import Integrate, integrator
-from funsor.ops import Op
-from funsor.terms import Align, Binary, Funsor, FunsorMeta, Number, Subs, Variable, eager, to_funsor
+from funsor.interpreter import debug_logged
+from funsor.ops import Op, TransformOp
+from funsor.registry import KeyedRegistry
+from funsor.terms import Align, Binary, Funsor, FunsorMeta, Number, Subs, Unary, Variable, eager, to_funsor
 
 
 class DeltaMeta(FunsorMeta):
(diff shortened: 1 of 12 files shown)


fritzo:
cc @neerajprad @martinjankowiak

funsor/delta.py
```diff
@@ -118,6 +126,49 @@ def eager_integrate(delta, integrand, reduced_vars):
     return Integrate(log_measure, integrand, reduced_vars)
 
 
+def solve(expr, value):
+    """
+    Tries to solve for free inputs of an ``expr`` such that ``expr == value``,
+    and computes the log-abs-det-Jacobian of the resulting substitution.
```
eb8680: If we merge #78 I think we should be able to factor this into a unification step and a log-abs-det-Jacobian step, where unification is handled by `unification.unify`.

fritzo: Sounds like a good follow-up PR.

funsor/distributions.py
```diff
@@ -164,6 +164,14 @@ def eager_delta(v, log_density, value):
     return funsor.delta.Delta(v.name, value, log_density)
 
 
+def LogNormal(loc, scale, value=None):
```
fritzo: We should be able to follow this pattern to create more general transformed distributions.

funsor/ops.py
```diff
@@ -28,6 +28,34 @@ def __str__(self):
...
+        self.log_abs_det_jacobian = fn
+        return fn
+
+    @staticmethod
+    def inv(x):
+        raise NotImplementedError
+
+    @staticmethod
+    def log_abs_det_jacobian(x, y):
+        raise NotImplementedError
```
eb8680: Do `inv` and `log_abs_det_jacobian` need to be methods, as opposed to living in tables like other facts about `Op`s?  Is this so we can support more advanced monolithic transforms later on?

fritzo: I guess the could live in tables. The method interface is closer to `torch.distributions.Transform`, which makes it a little more familiar.

fritzo: I imagine more complex transform later using `@solve.register` rather than this `TransformOp` interface. I'm not set on any of the interfaces in this PR, and I assume we'll readjust once `Joint` is more flexible and we can write more examples and tests.

eb8680: Ok, I'll go ahead and merge and we can revisit questions about interfaces when `Joint` is ready for it.

PR #105, opened 2019-03-28
Add Contract to optimizer

eb8680:
This PR adds `Contract` to the optimizer and parametrizes it with a sum and product op.  As far as I can tell there are no performance regressions, and we should now be able to optimize `Integrate` expressions.


funsor/optimizer.py
```diff
@@ -197,17 +205,14 @@ def optimize_reduction(op, arg, reduced_vars):
         # count new appearance of variables that aren't reduced
         reduce_dim_counter.update({d: 1 for d in reduced_vars & (both_vars - path_end_reduced_vars)})
 
-        path_end = Binary(finitary_op, ta, tb)
-        if path_end_reduced_vars:
-            path_end = Reduce(reduce_op, path_end, path_end_reduced_vars)
-
+        path_end = Contract(sum_op, prod_op, ta, tb, path_end_reduced_vars)
```
eb8680: Note that the path optimizer can be called recursively here if `ta` or `tb` are `Finitary`

PR #106, opened 2019-03-28
Sketch examples of partially analytic KL computation of ELBOs

eb8680:
Pair coded with @fritzo 

Blocked by #156


eb8680:
Will revisit at some point...

Issue #57, opened 2019-03-08
Fix funsor.minipyro and add tests


fritzo:
[funsor/minipyro.py](https://github.com/pyro-ppl/funsor/blob/master/funsor/minipyro.py) is currently only a storyboard: it does not function but does serve to demonstrate syntax and computations we'd like to support in funsor.

## Tasks
- [ ] add tests for eager sampling
- [ ] add tests for delayed sampling
- [ ] add tests for `trace`
- [ ] add tests for `replay`
- [ ] add tests for `block`
- [ ] add tests for `pyro.plate`
- [ ] add tests for `log_joint`
- [ ] add tests for `barrier`
- [ ] add tests for `ground`

PR #107, opened 2019-03-28
Make distributions lazy when used Pyro-style

fritzo:
Unblocks #57 

This makes distributions lazy when used Pyro-style, but keeps them eager when used Funsor-style.

In Pyro we construct distributions without specifying the variable name:
```py
fn = dist.Normal(0, 1)    # from a pyro.sample statement
log_prob += fn(value)     # from an internal elbo computation
```
Whereas in Funsor we idiomatically name the variate
```py
factor = dist.Normal(0, 1, Variable('x', reals()))
log_prob += factor
```
Prior to this PR, `dist.Normal(0,1)` would eagerly compile to a `Gaussian`. After this PR, `dist.Normal(0,1)` remains an inspectable `Normal` distribution until its value is substituted (possibly for another variable), e.g. in the second line above.

Advantage of this PR include:
1. We can avoid some computation (currently not even implemented, and inevitable numerically unstable) in pyro observe statements whose parameters are delayed variables, since a large (rank deficient) Gaussian will never be constructed. Instead we will wait until an observation is available as in the above example elbo computation.
2. Pyro statements will be a little more debuggable since the original distribution will be available.

This may seem a little gross, but I see it as a way for funsor.distributions to be used in both older pyro-style and newer funsor-style with no conflict (e.g. no funsor.compat.distributions shim).

## Questions

- [x] Should we instead simply disallow specifying the variate in the `__init__` method?
    This way all distributions would be lazily constructed, and could be converted to e.g. `Gaussian` or `Joint` by substitution of `variable`.

## Tested
- updated existing tests
- added a new delta check
- ran in `pyro/examples/minipyro.py --backend=funsor`

### DIFF
--- a/funsor/distributions.py
+++ b/funsor/distributions.py
@@ -11,7 +11,8 @@ import funsor.delta
 import funsor.ops as ops
 from funsor.domains import bint, reals
 from funsor.gaussian import Gaussian
-from funsor.terms import Funsor, FunsorMeta, Number, Subs, Variable, eager, to_funsor
+from funsor.interpreter import interpretation
+from funsor.terms import Funsor, FunsorMeta, Number, Subs, Variable, eager, lazy, to_funsor
 from funsor.torch import Tensor, align_tensors, materialize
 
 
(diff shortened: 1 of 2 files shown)


funsor/distributions.py
```diff
@@ -220,7 +231,7 @@ def eager_normal(loc, scale, value):
 def eager_normal(loc, scale, value):
     if not isinstance(loc, Tensor):
         loc, value = value, loc
-    return Normal(loc, scale)(value=value)
+    return Normal(loc, scale, None)(value=value)
```
fritzo: Note that `None` triggers eager construction of a `Gaussian` or `Joint`.

PR #108, opened 2019-03-28
Implement plate reductions for Gaussian, Joint

fritzo:
Addresses #57 

This also adds a more stable scalar case for `_sym_solve_mv()`.

## Tested
- [x] ran under `pyro/examples/minipyro.py --backend=funsor`
- [x] unit test for `Gaussian` plate reduction
- [x] unit test for `Joint` plate reduction

### DIFF
--- a/funsor/gaussian.py
+++ b/funsor/gaussian.py
@@ -335,10 +335,13 @@ class Gaussian(Funsor):
             precision = Tensor(self.precision, old_ints).reduce(ops.add, reduced_vars)
             precision_loc = Tensor(_mv(self.precision, self.loc),
                                    old_ints).reduce(ops.add, reduced_vars)
+            assert precision.inputs == new_ints
+            assert precision_loc.inputs == new_ints
             loc = Tensor(_sym_solve_mv(precision.data, precision_loc.data), new_ints)
             expanded_loc = align_tensor(old_ints, loc)
             quadratic_term = Tensor(_vmv(self.precision, expanded_loc - self.loc),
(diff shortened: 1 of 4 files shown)


funsor/gaussian.py
```diff
@@ -317,7 +323,24 @@ def eager_reduce(self, op, reduced_vars):
...
+            old_ints = OrderedDict((k, v) for k, v in self.inputs.items() if v.dtype != 'real')
+            new_ints = OrderedDict((k, v) for k, v in old_ints.items() if k not in reduced_vars)
+            inputs = OrderedDict((k, v) for k, v in self.inputs.items() if k not in reduced_vars)
+
+            precision = Tensor(self.precision, old_ints).reduce(ops.add, reduced_vars)
+            precision_loc = Tensor(_mv(self.precision, self.loc),
+                                   old_ints).reduce(ops.add, reduced_vars)
+            loc = Tensor(_sym_solve_mv(precision.data, precision_loc.data), new_ints)
+            expanded_loc = align_tensor(old_ints, loc)
+            quadratic_term = Tensor(_vmv(self.precision, expanded_loc - self.loc),
```
eb8680: The math looks right.  Are `self.precision` and `self.loc` guaranteed to be aligned correctly so that these linear algebra helpers do the right thing?

fritzo: Yes, `align_tensor` should take care of that, but I'll throw in an assertion to make it clear:
- `self.precision`, `self.loc`, and `expanded_loc` are aligned to `old_ints`
- `loc` is aligned to `new_loc`

PR #114, opened 2019-04-02
Add initial sphinx docs

fritzo:
This adds initial sphinx docs.
- To view docs, use Python 3 and locally `make docs && open docs/build/html/index.html`
- I have not yet added docs to CI testing, so as not to impede early development

## Tested
- generated docs locally and inspected them


fritzo:
hmm still need to fix some generation bugs...

PR #116, opened 2019-04-05
Add an Affine term to represent multilinear functions of real Variables

eb8680:
Pair coded with @fritzo.  Addresses #72.

This PR adds an `Affine` term for representing multilinear functions of `Variable`s and includes it in the `eager` interpretation by default.

Tasks:
- [x] Implement `eager_subs`
- [x] Fix tests that were broken by changes to `eager`

### DIFF
--- a/funsor/affine.py
+++ b/funsor/affine.py
@@ -1,8 +1,10 @@
 from __future__ import absolute_import, division, print_function
 
 from collections import OrderedDict
+from six.moves import reduce
 
 import funsor.ops as ops
+from funsor.domains import find_domain
 from funsor.ops import NegOp, Op
 from funsor.terms import (
     Binary,
@@ -17,7 +19,9 @@ from funsor.torch import Tensor
 
 
 class Affine(Funsor):
-
+    """
+    Pattern representing multilinear function of input variables
+    """
     def __init__(self, const, coeffs):
         assert isinstance(const, (Number, Tensor))
         assert isinstance(coeffs, tuple)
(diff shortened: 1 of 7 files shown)


funsor/affine.py
```diff
@@ -0,0 +1,172 @@
...
+    const = -affine.const
+    coeffs = affine.coeffs.copy()
+    for name, coeff in coeffs.items():
+        coeffs[name] = -coeff
+    return Affine(const, tuple(coeffs.items()))
+
+
+#########################################
+# patterns for creating new Affine terms
+#########################################
```
eb8680: Looks like a bunch of other tests got broken because these change the output type of formerly lazy operations

fritzo: I wonder if we should restrict `Affine` to the `moment_matching` interpretation? Either way we should fix those tests.

funsor/affine.py
```diff
@@ -0,0 +1,209 @@
+from __future__ import absolute_import, division, print_function
+
+from collections import OrderedDict
+
+import funsor.ops as ops
+from funsor.domains import find_domain
+from funsor.ops import NegOp, Op
+from funsor.terms import (
```
fritzo: nit: could you run `make format` or literally `isort -y`

funsor/affine.py
```diff
@@ -0,0 +1,209 @@
...
+    return None
+
+
+@eager.register(Binary, Op, Variable, Affine)
+def eager_binary_variable_affine(op, other, affine):
+    if op is ops.add:
+        return affine + other
+
+    if op is ops.sub:
+        return other + -affine
```
fritzo: nit: might as well save one step and
```py
return -affine + other
```

funsor/affine.py
```diff
@@ -0,0 +1,209 @@
...
+
+@eager.register(Binary, Op, (Number, Tensor), Variable)
+def eager_binary(op, other, var):
+    if other.dtype != "real" or var.dtype != "real":
+        return None
+
+    if op is ops.add or op is ops.mul:
+        return op(var, other)
+    elif op is ops.sub:
+        return other + -var
```
fritzo: nit: ditto, might as well save a step and
```py
return -var + other
```

test/test_affine.py
```diff
@@ -0,0 +1,76 @@
...
+    ("(t * x)(i=1, x=y)", Affine, {"j": bint(3), "y": reals()}, reals()),
+    ("(t * x + n)(x=y)", Affine, {"y": reals(), "i": bint(2), "j": bint(3)}, reals()),
+    ("(x + y)(y=z)", Affine, {"x": reals(), "z": reals()}, reals()),
+    ("(-x)(x=y+z)", Affine, {"y": reals(), "z": reals()}, reals()),
+    ("(t * x + t * y)(x=z)", Affine, {"y": reals(), "z": reals(), "i": bint(2), "j": bint(3)}, reals()),
+]
+
+
+@pytest.mark.parametrize("expr,expected_type,expected_inputs,expected_output", SUBS_TESTS)
+def test_affine_subs(expr, expected_type, expected_inputs, expected_output):
```
fritzo: I guess we can now hard-code `expected_output`

PR #125, opened 2019-04-09
Reinstate examples/minipyro.py, forked from pyro/examples

fritzo:
Joint work with @eb8680 

This replaces the outdated minipyro example with a fork of Pyro's minipyro example and adds it to our test suite.

## Tasks
- [x] update model
- [x] add unit test
- [x] fix failing tests

### DIFF
--- a/funsor/minipyro.py
+++ b/funsor/minipyro.py
@@ -345,15 +345,16 @@ class SVI(object):
 
 # TODO(eb8680) Replace this with funsor.Expectation.
 def Expectation(log_probs, costs, sum_vars, prod_vars):
-    probs = [p.exp() for p in log_probs]
     result = 0
     for cost in costs:
-        result += funsor.sum_product.sum_product(
-                sum_op=funsor.ops.add,
-                prod_op=funsor.ops.mul,
-                factors=probs + [cost],
-                plates=prod_vars,
-                eliminate=prod_vars | sum_vars)
+        log_prob = funsor.sum_product.sum_product(
+            sum_op=funsor.ops.logaddexp,
+            prod_op=funsor.ops.add,
+            factors=log_probs,
+            plates=prod_vars,
+            eliminate=(prod_vars | sum_vars) - frozenset(cost.inputs)
+        )
+        result += funsor.Integrate(log_prob, cost, frozenset(cost.inputs))
     return result
 
 
--- a/funsor/minipyro.py
+++ b/funsor/minipyro.py
@@ -355,7 +355,7 @@ def Expectation(log_probs, costs, sum_vars, prod_vars):
             eliminate=(prod_vars | sum_vars) - frozenset(cost.inputs)
         )
         term = funsor.Integrate(log_prob, cost, sum_vars & frozenset(cost.inputs))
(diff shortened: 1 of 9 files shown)


test/test_gaussian.py
```diff
@@ -342,3 +343,15 @@ def test_integrate_gaussian(int_inputs, real_inputs):
...
+@pytest.mark.xfail(reason="numerically unstable")
+def test_mc_plate_gaussian():
+    log_measure = Gaussian(torch.tensor([0.]), torch.tensor([[1.]]),
+                           (('loc', reals()),)) + torch.tensor(-0.9189)
+    integrand = Gaussian(torch.randn((100, 1)) + 3., torch.ones((100, 1, 1)),
+                         (('data', bint(100)), ('loc', reals())))
+    with interpretation(monte_carlo):
+        res = Integrate(log_measure, integrand, frozenset({'loc'}))
+        res = res.reduce(ops.mul, frozenset({'data'}))
+        assert not torch.isinf(res).any()
```
eb8680: @fritzo the minipyro example in this PR now fails with an `inf` loss.  This test demonstrates the fragment of computation within the ELBO that is the source of the numerical instability.

funsor/minipyro.py
```diff
@@ -345,15 +345,16 @@ def step(self, *args, **kwargs):
...
     for cost in costs:
-        result += funsor.sum_product.sum_product(
-                sum_op=funsor.ops.add,
-                prod_op=funsor.ops.mul,
-                factors=probs + [cost],
-                plates=prod_vars,
-                eliminate=prod_vars | sum_vars)
+        log_prob = funsor.sum_product.sum_product(
+            sum_op=funsor.ops.logaddexp,
+            prod_op=funsor.ops.add,
```
eb8680: @fritzo this fixes the numerical errors in the example, but needs more testing

funsor/minipyro.py
```diff
@@ -345,15 +345,16 @@ def step(self, *args, **kwargs):
...
+            factors=log_probs,
+            plates=prod_vars,
+            eliminate=(prod_vars | sum_vars) - frozenset(cost.inputs)
+        )
+        result += funsor.Integrate(log_prob, cost, frozenset(cost.inputs))
```
fritzo: But what if cost is plated? Or batched?

fritzo: Maybe fix the `reduced_vars` arg here and add a final optional `.reduce(ops.add, ..., remaining_plates)`?

funsor/minipyro.py
```diff
@@ -345,15 +345,18 @@ def step(self, *args, **kwargs):
...
-    probs = [p.exp() for p in log_probs]
     result = 0
     for cost in costs:
-        result += funsor.sum_product.sum_product(
-                sum_op=funsor.ops.add,
-                prod_op=funsor.ops.mul,
-                factors=probs + [cost],
-                plates=prod_vars,
-                eliminate=prod_vars | sum_vars)
+        log_prob = funsor.sum_product.sum_product(
```
eb8680: In `pyro.infer.util.Dice.compute_expectation`, we use `adjoint` to share intermediate computations for costs in the same ordinal.  We could do the same thing here.

PR #128, opened 2019-04-12
Support constraints in funsor.minipyro

fritzo:
This enables constraints in `pyro.param` statements backed by funsor.minipyro.

I've avoided recreating the tricky constraint logic and simply imported Pyro's `ParamStoreDict`. This seems not too bad since the `ParamStoreDict` is independent of the rest of Pyro.

The motivation is to enable learning of constrained parameters e.g. probability distributions of discrete distributions. I had tried to port the Kalman Filter example to funsor.minipyro, but it looked gross to store unconstrained parameters in the param store. After this PR it should be easier to create a KF example to use minipyro.

## Tested
- [x] added a smoke test
- [x] updated minipyro tests to run for 2 iterations so as to test both cases:
    param not found (on step 1); param found (on step 2)

### DIFF
--- a/funsor/minipyro.py
+++ b/funsor/minipyro.py
@@ -256,9 +256,8 @@ def param(name, init_value=None, constraint=None, event_dim=None):
             value = PARAM_STORE[name]
         else:
             value = PARAM_STORE.setdefault(name, init_value, constraint)
-            value._funsor_cond_indep_stack = cond_indep_stack
-            value._funsor_output = output
-        return tensor_to_funsor(value, value._funsor_cond_indep_stack, value._funsor_output)
+            value.unconstrained()._funsor_metadata = (cond_indep_stack, output)
+        return tensor_to_funsor(value, *value.unconstrained()._funsor_metadata)
 
     # if there are no active Messengers, we just draw a sample and return it as expected:
     if not PYRO_STACK:
--- a/funsor/minipyro.py
+++ b/funsor/minipyro.py
@@ -16,6 +16,7 @@ from __future__ import absolute_import, division, print_function
 from collections import OrderedDict, namedtuple
 
 import torch
+from pyro.params.param_store import ParamStoreDict
 
 import funsor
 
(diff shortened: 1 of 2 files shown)


test/test_minipyro.py
```diff
@@ -119,11 +120,7 @@ def guide():
         assert_ok(model, guide, elbo)
 
 
-@pytest.mark.parametrize("backend", [
-    "pyro",
-    "minipyro",
-    xfail_param("funsor", reason="missing patterns"),
```
fritzo: this was passing so I re-enabled it

funsor/minipyro.py
```diff
@@ -240,35 +241,35 @@ def sample(name, fn, obs=None):
...
     if init_value is not None:
         if event_dim is None:
             event_dim = init_value.dim()
         output = funsor.reals(*init_value.shape[init_value.dim() - event_dim:])
+    if constraint is None:
+        constraint = torch.distributions.constraints.real
 
-    def fn(init_value):
-        if name in PARAM_STORE:
+    def fn(init_value, constraint):
```
fritzo: The following logic was adapted from [ParamStoreDict.get_param()](https://github.com/pyro-ppl/pyro/blob/dbe0415/pyro/params/param_store.py#L186).

funsor/minipyro.py
```diff
@@ -60,7 +61,7 @@ def expand_inputs(self, name, size):
 #     See http://docs.pyro.ai/en/0.3.1/parameters.html
 
 PYRO_STACK = []
-PARAM_STORE = {}
+PARAM_STORE = ParamStoreDict()  # A dict-like object that also supports constraints.
```
eb8680: I guess this is OK for now, but the `ParamStoreDict` code is pretty gross.  It really only needs to be a dictionary whose values are tuples `(unconstrained_value, constraint)`, with `pyro.param` responsible for applying the constraint.  Maybe we should fix that in Pyro's minipyro and then port the fix to funsor.

PR #129, opened 2019-04-12
Remove dependency on pyro ParamStoreDict

fritzo:
Replaces #128 by making funsor.minipyro closer to pyro.contrib.minipyro https://github.com/pyro-ppl/pyro/pull/1820 , as suggested by @eb8680 .

## Tested

- [x] updated examples/minipyro.py
- [x] added a test case to test/test_minipyro.py

### DIFF
--- a/funsor/minipyro.py
+++ b/funsor/minipyro.py
@@ -13,10 +13,10 @@ found at examples/minipyro.py.
 """
 from __future__ import absolute_import, division, print_function
 
+import weakref
 from collections import OrderedDict, namedtuple
 
 import torch
-from pyro.params.param_store import ParamStoreDict
 
 import funsor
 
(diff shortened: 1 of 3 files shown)


funsor/minipyro.py
```diff
@@ -239,25 +239,33 @@ def sample(name, fn, obs=None):
...
-            value = PARAM_STORE.setdefault(name, init_value, constraint)
-            value.unconstrained()._funsor_metadata = (cond_indep_stack, output)
-        return tensor_to_funsor(value, *value.unconstrained()._funsor_metadata)
+            # Initialize with a constrained value.
+            assert init_value is not None
+            with torch.no_grad():
+                constrained_value = init_value.detach()
+                unconstrained_value = torch.distributions.transform_to(constraint).inv(constrained_value)
+            unconstrained_value.requires_grad_()
+            unconstrained_value._funsor_metadata = (cond_indep_stack, output)
```
fritzo: Only this line and the return statement below differ from pyro.contrib.minipyro

PR #130, opened 2019-04-22
Add RandomVariable classes that carry deferred sample calls

eb8680:
Blocked by #157


eb8680:
Will revisit at some point...

PR #135, opened 2019-05-01
Add BernoulliLogits distribution

fritzo:
This adds a `BernoulliLogits(logits)` distribution. In the future we may merge this with the Bernoulli distribution, but currently we don't have enough kwargs machinery to handle optional params in distributions.

The motivating use case is a linear-gaussian HMM with a probit likelihood model.

## Tested
- added a unit test for likelihood
- added a unit test for `Bernoulli` as well

### DIFF
--- a/funsor/distributions.py
+++ b/funsor/distributions.py
@@ -119,8 +119,27 @@ class Bernoulli(Distribution):
 
 
 @eager.register(Bernoulli, Tensor, Tensor)
-def eager_categorical(probs, value):
+def eager_bernoulli(probs, value):
     return Bernoulli.eager_log_prob(probs=probs, value=value)
+
+
+class BernoulliLogits(Distribution):
+    dist_class = dist.Bernoulli
+
+    @staticmethod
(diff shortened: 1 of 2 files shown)


fritzo:
Thanks for reviewing!

> in numpyro ... separate classes for different parametrizations ... is working out well

Interesting. I guess a compromise would be to keep separate classes under the hood (e.g. `BernoulliProbs`, `BernoulliLogits`) and let a top level function dispatch based on available kwargs, e.g.
```py
def Bernoulli(**kwargs):
    """Dispatch to either BernoulliProbs or BernoulliLogits"""
    if 'probs' in kwargs:
        probs = kwargs.pop('probs')
        return BernoulliProbs(probs, **kwargs)
    if 'logits' in kwargs:
        logits = kwargs.pop('logits')
        return BernoulliLogits(logits, **kwargs)
    raise ValueError
```

neerajprad:
I think that's a good suggestion to decouple the interface with the implementation, will make this change.

PR #139, opened 2019-05-01
rtd fix to not install torch 1.0

jpchen:


jpchen:
hm there seems to be a rtd python 3.7 issue, even though the build succeeds locally and on the `docs` branch. need to investigate further
```
Exception occurred:
  File "/home/docs/checkouts/readthedocs.org/user_builds/funsor/envs/latest/lib/python3.7/site-packages/sphinx/ext/autodoc/__init__.py", line 82, in members_option
    return [x.strip() for x in arg.split(',')]
AttributeError: 'bool' object has no attribute 'split'
The full traceback has been saved in /tmp/sphinx-err-bevkvl
```

fritzo:
Oh yeah, that's due to an outdated version of sphinx-autodoc. The offending line is in our sphinx config
https://github.com/pyro-ppl/funsor/blob/c77dc99/docs/source/conf.py#L64
We can just comment-out that line until rtd updates their sphinx install.

PR #147, opened 2019-05-24
Add an explicit stack-based interpreter

### LINKED ISSUE #145: Add tail call optimization to interpreter
Necessary for evaluating deeply nested expressions without hitting the Python interpreter stack limit, like marginal likelihoods for time series models

eb8680:
Resolves #145 

This PR adds a simple stack-based interpreter as a drop-in replacement for `funsor.interpreter.reinterpret`, so that it's possible to evaluate arbitrarily large deferred expressions that would otherwise exceed the Python interpreter's maximum recursion depth.  

The new interpreter is 2-3x slower than the current recursion-based one since it's using Python data structures, so I've also left the current interpreter in as the default.

The new interpreter is accessible via an environment variable `FUNSOR_USE_TCO=1`:
```
# fails with a RecursionError
python examples/kalman_filter.py -n 2 -t 500 --lazy

# this works
FUNSOR_USE_TCO=1 python examples/kalman_filter.py -n 2 -t 500 --lazy
```

Tested: exercised by running existing tests with `FUNSOR_USE_TCO=1`, and by running existing examples with longer time series.

### DIFF
--- a/funsor/interpreter.py
+++ b/funsor/interpreter.py
@@ -93,9 +93,9 @@ def _children_ground(x):
     return ()
 
 
-def is_ground(x):
+def is_atom(x):
     if isinstance(x, (tuple, frozenset)) and not isinstance(x, Domain):
-        return len(x) == 0 or all(is_ground(c) for c in x)
+        return len(x) == 0 or all(is_atom(c) for c in x)
     return isinstance(x, (
         int,
         str,
@@ -153,7 +153,7 @@ def reinterpret(x):
                 leaves.append(parent)
 
         h = node_vars[h_name]
-        if is_ground(h):
+        if is_atom(h):
             env[h_name] = h
         elif isinstance(h, (tuple, frozenset)):
             env[h_name] = type(h)(
--- a/funsor/interpreter.py
+++ b/funsor/interpreter.py
@@ -5,7 +5,7 @@ import inspect
 import os
 import re
 import types
-from collections import OrderedDict
+import uuid
 
 import torch
 from contextlib2 import contextmanager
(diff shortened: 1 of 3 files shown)


funsor/interpreter.py
```diff
@@ -87,39 +90,166 @@ def reinterpret_funsor(x):
...
+    :type x: A funsor or data structure holding funsors.
+    :return: A reinterpreted version of the input.
+    :raises: ValueError
+    """
+    node_vars = {}
+    env = {}
+    x_name = gensym()
+    stack = [(x_name, x)]
+    parent_to_children = OrderedDict()
+    child_to_parent = OrderedDict()
```
fritzo: Is there a reason you're treating funsors as trees rather than dags? I would have expected dag processing along the lines of 
```py
names = {x_name: x}
parents_to_children = OrderedDict()
children_to_parents = OrderedDict()
while stack:
    ...
    for c in children(h):
        if c in names:
            c_name = names[c]
        else:
            c_name = gensym()
            names[c] = c_name
            stack.append((c_name, c))
        children_to_parents.setdefault(c_name, []).append(h_name)
        parents_to_children[h_name].append(c_name)
```
IIUC these approaches can differ exponentially, e.g.
```py
def example(depth=100):
    x = Variable('x')
    for _ in range(depth):
        x = x * x
    return x 
```

funsor/interpreter.py
```diff
@@ -87,39 +90,166 @@ def reinterpret_funsor(x):
...
+
+    :param x: An input, typically involving deferred
+        :class:`~funsor.terms.Funsor` s.
+    :type x: A funsor or data structure holding funsors.
+    :return: A reinterpreted version of the input.
+    :raises: ValueError
+    """
+    node_vars = {}
+    env = {}
+    x_name = gensym()
```
fritzo: Is there a reason you're using `uuid` rather `Funsor.__hash__` (equivalent to `id(x)`)?

fritzo: Ah, I see this also needs to work for non-Funsors. I think it should still work to use `id(x)` though.

funsor/interpreter.py
```diff
@@ -87,39 +90,166 @@ def reinterpret_funsor(x):
...
+    return "V" + str(uuid.uuid4().hex)
+
+
+def stack_reinterpret(x):
+    r"""
+    Overloaded reinterpretation of a deferred expression.
+    This interpreter uses an explicit stack and no recursion but is much slower.
+
+    This handles a limited class of expressions, raising
+    ``ValueError`` in unhandled cases.
```
fritzo: Is this comment still true? When is `ValueError` raised?

eb8680: In the default behavior for `children` above

funsor/interpreter.py
eb8680: My first attempt at this wasn't hashing or refcounting correctly so I started with trees but I've fixed it now

PR #148, opened 2019-05-30
Alpha renaming of bound variables

### LINKED ISSUE #96: Implement alpha conversion logic
Many places in funsor raise `NotImplementedError` when a bound variable appears free in an expression being substituted. This should be resolved by alpha-converting the binder's expression before substitution (i.e. renaming the bound variable).

@eb8680 suggested handling this uniformly by refactoring e.g. the `Funsor` base class to know about subterms and bound variables. Probably the easiest way to resolve this issue is to refactor first.

## Tasks
- [x] Refactor to let `Funsor` base class know about variable binding and subterms
- [x] Implement alpha conversion once in the `Funsor` base class
- [ ] Make compatible with `funsor.adjoint`

eb8680:
Resolves #96.  Blocking #130.

This PR makes the following changes to substitution and interpretation:
1. Adds eager alpha-renaming whenever a variable is newly bound
2. Adds `.fresh` and `.bound` attributes to each term (empty by default) which are `frozenset`s of names of fresh variables introduced by a term and variables newly bound by a term respectively.  These must be manually populated in the `__init__` method of each new term if they are not empty for that term.
3. Changes the semantics of the `Subs` term slightly: there is now a `substitute` function that calls itself recursively, as opposed to `Subs` calling `eager_subs` methods which in turn recursively call `Subs`.  A `Subs` term now represents a lazy call to `substitute`.  I did this instead of manually changing every call to `Subs` inside an `eager_subs` method into recursive `eager_subs` calls.
4. Makes substitution more generic, removing the need for `.eager_subs` implementations in terms that don't introduce fresh variables and simplifying the implementation of new terms.  Hopefully most `eager_subs` methods of terms that don't introduce fresh variables can be removed in a followup PR.

Tasks:
- [x] Add `.fresh` and `.bound` to any remaining terms missing them and add some tests checking
- [x] Fix various minor regressions (`Affine`, `Joint`, `Stack`, `Lambda`, `Independent`, etc)
- [x] Add new tests exercising alpha-renaming (which would currently trigger `NotImplementedError`s)

Triaged:
- Make `funsor.adjoint` compatible with alpha-renaming

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,7 @@ format: FORCE
 
 test: lint FORCE
 	pytest -v test
+	FUNSOR_GENERIC_SUBS=1 pytest -v test  # TODO remove when removing eager_subs
 	FUNSOR_DEBUG=1 pytest -v test/test_gaussian.py
 	FUNSOR_USE_TCO=1 pytest -v test/test_terms.py
 	FUNSOR_USE_TCO=1 pytest -v test/test_einsum.py
--- a/funsor/affine.py
+++ b/funsor/affine.py
@@ -58,8 +58,10 @@ def eager_affine(const, coeffs):
     if not coeffs:
         return const
     if any(not isinstance(v, Variable) for v, c in coeffs):
-        return Affine(const + sum(v * c for v, c in coeffs if not isinstance(v, Variable)),
-                      tuple((v, c) for v, c in coeffs if isinstance(v, Variable)))
+        new_const = const + sum(
+            v * c for v, c in coeffs if not isinstance(v, Variable))
+        new_coeffs = tuple((v, c) for v, c in coeffs if isinstance(v, Variable))
+        return Affine(new_const, new_coeffs)
     return None
 
 
--- a/funsor/affine.py
+++ b/funsor/affine.py
@@ -57,6 +57,9 @@ class Affine(Funsor):
 def eager_affine(const, coeffs):
     if not coeffs:
         return const
+    if any(not isinstance(v, Variable) for v, c in coeffs):
+        return Affine(const + sum(v * c for v, c in coeffs if not isinstance(v, Variable)),
+                      tuple((v, c) for v, c in coeffs if isinstance(v, Variable)))
     return None
 
 
--- a/funsor/affine.py
+++ b/funsor/affine.py
@@ -57,6 +57,12 @@ class Affine(Funsor):
 def eager_affine(const, coeffs):
     if not coeffs:
         return const
+    if not all(isinstance(var, Variable) for var, coeff in coeffs):
+        result = Affine(const, tuple((var, coeff) for var, coeff in coeffs if isinstance(var, Variable)))
+        for var, coeff in coeffs:
+            if not isinstance(var, Variable):
+                result += var * coeff
+        return result
     return None
 
 
--- a/funsor/affine.py
+++ b/funsor/affine.py
@@ -57,11 +57,6 @@ class Affine(Funsor):
 def eager_affine(const, coeffs):
     if not coeffs:
         return const
-    if any(not isinstance(v, Variable) for v, c in coeffs):
-        new_const = const + sum(
-            v * c for v, c in coeffs if not isinstance(v, Variable))
-        new_coeffs = tuple((v, c) for v, c in coeffs if isinstance(v, Variable))
-        return Affine(new_const, new_coeffs)
     return None
 
 
--- a/funsor/gaussian.py
+++ b/funsor/gaussian.py
@@ -530,9 +530,9 @@ class Gaussian(Funsor):
         raise NotImplementedError('TODO implement partial sampling of real variables')
 
 
-@substitute.register(Gaussian, dict)
+@substitute.register(Gaussian, tuple)
 def subs_gaussian(expr, subs):
-    return expr.eager_subs(tuple((k, to_funsor(v, expr.inputs[k]) if k in expr.inputs else v) for k, v in subs.items()))
+    return expr.eager_subs(tuple((k, to_funsor(v, expr.inputs[k]) if k in expr.inputs else v) for k, v in subs))
 
 
 @eager.register(Binary, AddOp, Gaussian, Gaussian)
--- a/funsor/gaussian.py
+++ b/funsor/gaussian.py
@@ -14,7 +14,7 @@ from funsor.delta import Delta
 from funsor.domains import reals
 from funsor.integrate import Integrate, integrator
 from funsor.ops import AddOp, NegOp, SubOp
-from funsor.terms import Align, Binary, Funsor, FunsorMeta, Number, Subs, Unary, Variable, eager
+from funsor.terms import Align, Binary, Funsor, FunsorMeta, Number, Subs, Unary, Variable, eager, substitute, to_funsor
 from funsor.torch import Tensor, align_tensor, align_tensors, materialize
 from funsor.util import lazy_property
 
@@ -526,6 +526,11 @@ class Gaussian(Funsor):
             return reduce(ops.add, results)
 
         raise NotImplementedError('TODO implement partial sampling of real variables')
+
+
+@substitute.register(Gaussian, dict)
+def subs_gaussian(expr, subs):
+    return expr.eager_subs(tuple((k, to_funsor(v, expr.inputs[k]) if k in expr.inputs else v) for k, v in subs.items()))
 
 
 @eager.register(Binary, AddOp, Gaussian, Gaussian)
--- a/funsor/gaussian.py
+++ b/funsor/gaussian.py
@@ -322,7 +322,9 @@ class Gaussian(Funsor):
             assert _issubshape(precision.shape, batch_shape + (dim, dim))
 
         output = reals()
-        super(Gaussian, self).__init__(inputs, output)
+        fresh = frozenset(inputs.keys())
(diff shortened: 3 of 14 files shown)


funsor/joint.py
```diff
@@ -244,6 +245,11 @@ def eager_independent(joint, reals_var, bint_var):
     return None  # defer to default implementation
 
 
+@substitute.register(Joint, dict)
+def substitute_joint(expr, subs):
+    return expr.eager_subs(tuple(subs.items()))
```
eb8680: This is a bit of a hack, but we should be able to get rid of it since `Joint` does not introduce fresh variables.

funsor/gaussian.py
```diff
@@ -528,6 +530,11 @@ def unscaled_sample(self, sampled_vars, sample_inputs):
         raise NotImplementedError('TODO implement partial sampling of real variables')
 
 
+@substitute.register(Gaussian, dict)
+def subs_gaussian(expr, subs):
+    return expr.eager_subs(tuple((k, to_funsor(v, expr.inputs[k]) if k in expr.inputs else v) for k, v in subs.items()))
```
eb8680: Unlike `Joint`, `Gaussian` does introduce fresh variables, so calling `eager_subs` is fine here.

funsor/interpreter.py
```diff
@@ -23,6 +23,9 @@
 _INTERPRETATION = None  # To be set later in funsor.terms
 _USE_TCO = int(os.environ.get("FUNSOR_USE_TCO", 0))
 
+# TODO remove this, used temporarily for testing
+_GENERIC_SUBS = int(os.environ.get("FUNSOR_GENERIC_SUBS", 0))
```
eb8680: I added this temporary environment variable for testing, but it should be removed either before merging this PR or in a followup PR that removes non-essential `eager_subs` methods.

funsor/terms.py
```diff
@@ -950,6 +1052,14 @@ def eager_reduce(self, op, reduced_vars):
         return self.arg.reduce(op, reduced_vars)
 
 
+@eager.register(Align, Funsor, tuple)
+def eager_align(arg, names):
+    if not frozenset(names) == frozenset(arg.inputs.keys()):
+        # assume there's been a substitution and this align is no longer valid
+        return arg
```
eb8680: Compare to `Align.eager_subs`

test/test_alpha_conversion.py
```diff
@@ -0,0 +1,109 @@
...
+
+import funsor.ops as ops
+from funsor.domains import bint, reals
+from funsor.interpreter import gensym, interpretation, _GENERIC_SUBS
+from funsor.terms import Independent, Lambda, Variable, reflect
+from funsor.testing import assert_close, check_funsor, random_tensor
+
+
+# TODO remove this when removing eager_subs methods
+xfail_if_old_subs = pytest.mark.xfail(not _GENERIC_SUBS, reason="fails w/ old subs")
```
eb8680: I've marked these new tests exercising alpha-renaming with a conditional `xfail`.  I will remove this as part of addressing #149

test/test_adjoint.py
```diff
@@ -10,10 +10,15 @@
...
 from funsor.domains import bint
 from funsor.einsum import einsum, naive_einsum, naive_plated_einsum
-from funsor.interpreter import interpretation
+from funsor.interpreter import interpretation, _GENERIC_SUBS
 from funsor.terms import Variable, reflect
 from funsor.testing import make_einsum_example, make_plated_hmm_einsum
 
+
+# TODO remove this when removing eager_subs methods
+xfail_if_new_subs = pytest.mark.xfail(_GENERIC_SUBS, reason="fails w/ new subs")
```
eb8680: The changes to `funsor.adjoint` necessary to support alpha-renaming are beyond the scope of this PR, so I've marked these tests as xfailing and will refactor in a followup PR.

Makefile
```diff
@@ -16,6 +16,7 @@ format: FORCE
 
 test: lint FORCE
 	pytest -v test
+	FUNSOR_GENERIC_SUBS=1 pytest -v test  # TODO remove when removing eager_subs
```
eb8680: I've temporarily added a second test suite execution under the new subsitution changes in this PR for CI.  I will remove this as part of addressing #149.

funsor/affine.py
```diff
@@ -57,6 +57,12 @@ def eager_subs(self, subs):
 def eager_affine(const, coeffs):
     if not coeffs:
         return const
+    if not all(isinstance(var, Variable) for var, coeff in coeffs):
```
fritzo: When would a `Affine(...)` be called with non-variables?

funsor/joint.py
fritzo: Can you explain why this is a hack?

funsor/terms.py
```diff
@@ -20,6 +20,63 @@
...
 
 
+@dispatch(object, object)
+def substitute(expr, subs):
+    return expr  # default: do nothing
+
+
+@substitute.register(object, tuple)
+def substitute_with_tuple(expr, subs):
+    return substitute(expr, dict(subs))
```
fritzo: This introduces nondeterminism. I'd prefer to use tuples everywhere. What is your reason for using a dict?

funsor/terms.py
```diff
@@ -20,6 +20,63 @@
...
+    return frozenset(substitute(v, subs) for v in expr)
+
+
+@substitute.register(OrderedDict, dict)
+def substitute_ordereddict(expr, subs):
+    return OrderedDict([(k, substitute(v, subs)) for k, v in expr.items()])
+
+
+def alpha_convert(expr):
+    if not expr.bound or all("__BOUND" in name for name in expr.bound):
```
fritzo: nit: simplify
```py
if all("__BOUND" in name for name in expr.bound):
    return expr
```

funsor/joint.py
eb8680: Semantically, `substitute_joint` should not need to do anything `Joint`-specific and should be able to reuse the generic `substitute_funsor` logic from `terms.py`, but that doesn't quite work yet so I'm just calling `Joint.eager_subs`.

funsor/affine.py
eb8680: This happens as a result of the changes to substitution, which reuses the interpreter and patterns here in this file to normalize the result of substitution.  Compare to the logic in `Affine.eager_subs`, where that re-normalization was hard-coded.

funsor/terms.py
```diff
@@ -20,6 +20,63 @@
...
+        return expr
+
+    alpha_subs = {name: interpreter.gensym(name + "__BOUND")
+                  for name in expr.bound if "__BOUND" not in name}
+
+    new_values = []
+    for v in expr._ast_values:
+        v = substitute(v, alpha_subs)
+        if isinstance(v, str):
+            v = alpha_subs[v] if v in alpha_subs else v
```
fritzo: nit:
```py
v = alpha_subs.get(v, v)
```

funsor/terms.py
eb8680: Nothing important, I'll change it to use tuples.

funsor/joint.py
fritzo: I still don't understand why "that doesn't quite work yet", so it's probably worth adding a comment like `#TODO remove this unnecessary method after ___`

eb8680: > I still don't understand why "that doesn't quite work yet"

If I comment those lines out then a bunch of tests in `test_joint.py` fail due to missing patterns for `Delta` and `Joint`.  It's a more complicated version of the case of `Affine` where I had to add a pattern so that the result of a generic substitution could be normalized by the interpreter.

funsor/terms.py
```diff
@@ -20,6 +20,63 @@
...
+            v |= frozenset(alpha_subs[k] for k in swapped)
+            v -= swapped
+        elif isinstance(v, tuple) and isinstance(v[0], tuple) and len(v[0]) == 2 and \
+                isinstance(v[0][0], str) and isinstance(v[0][1], Funsor):
+            v = tuple((alpha_subs[k] if k in alpha_subs else k, vv) for k, vv in v)
+        elif isinstance(v, OrderedDict):  # XXX is this case ever actually triggered?
+            v = OrderedDict([(alpha_subs[k] if k in alpha_subs else k, vv) for k, vv in v.items()])
+        new_values.append(v)
+
+    # TODO should this call reflect explicitly?
```
fritzo: Yes, I think this should call `reflect()` explicitly.

It already feels dangerous to make `reflect()` reentrant, and making it subject to interpretation seems even more dangerous. If we assume interpretations are alpha-invariant, then it should be fine to call `reflect()` here. The only issue I can possibly think of is when running under some sort of tracing interpretation.

funsor/terms.py
```diff
@@ -521,6 +591,16 @@ def __getitem__(self, other):
 interpreter.children.register(Funsor)(interpreter.children_funsor)
 
 
+@substitute.register(Funsor, dict)
+def substitute_funsor(expr, subs):
+    subs = {name: sub for name, sub in subs.items() if name in expr.inputs}
+    if not subs:
+        return expr
+    assert not expr.fresh & frozenset(subs), "cannot generically substitute into a fresh variable"
```
fritzo: nit:
```py
assert expr.fresh.isdisjoint(subs), "..."
```

PR #155, opened 2019-06-26
Make substitution an interpretation

### LINKED ISSUE #149: Clean up substitution
#148 introduces several major changes to substitution as part of the implementation of alpha-conversion.  These changes should allow us to further simplify a lot of the old substitution logic:
- [x] #152 Remove `eager_subs` methods from all terms (e.g. `Binary`, `Joint`) that do not introduce fresh variables
- [x] #152 Remove conditional xfails and extra test runs added temporarily as part of #148 
- [x] #153 Factor out non-fresh substitution logic from remaining `eager_subs` methods (e.g in `Delta`) so that terms only have to be responsible for substituting into their fresh variables
- [x] #153 Make `Subs` compatible with `FUNSOR_DEBUG=1` again
- [x] #154 Update `Joint` and associated patterns so that `Joint.eager_subs` can be removed
- [ ] #155 Make the new `substitute` function into an interpreter, allowing tail-call optimized substitution via #145

eb8680:
Resolves #149 

This refactoring PR makes `substitute` into an interpretation, so that it can reuse the tail-call optimization added in #145 and duplicate code can be removed.  The computations actually done by `substitute` should remain unchanged and the cost should be identical.

Tested: exercised by existing tests, and also longer example runs that started failing after #148.

Tasks:
- [x] Make `substitute` into an interpretation
- [x] Update semantics and implementation of `Independent` to be compatible with this version of `substitute`

Triaged:
- Make tail-call optimization compatible with `funsor.numpy` #1 (added conditional xfail for now)

### DIFF
--- a/funsor/delta.py
+++ b/funsor/delta.py
@@ -129,14 +129,14 @@ eager.register(Binary, AddOp, Reduce, Delta)(
 
 @eager.register(Independent, Delta, str, str)
 def eager_independent(delta, reals_var, bint_var):
-    if delta.name == reals_var:
+    if delta.name == reals_var or delta.name.startswith(reals_var + "__BOUND"):
         i = Variable(bint_var, delta.inputs[bint_var])
         point = Lambda(i, delta.point)
         if bint_var in delta.log_density.inputs:
             log_density = delta.log_density.reduce(ops.add, bint_var)
         else:
             log_density = delta.log_density * delta.inputs[bint_var].dtype
-        return Delta(delta.name, point, log_density)
+        return Delta(reals_var, point, log_density)
 
     return None  # defer to default implementation
 
--- a/funsor/joint.py
+++ b/funsor/joint.py
@@ -234,7 +234,7 @@ def eager_joint(deltas, discrete, gaussian):
 @eager.register(Independent, Joint, str, str)
 def eager_independent(joint, reals_var, bint_var):
     for i, delta in enumerate(joint.deltas):
-        if delta.name == reals_var:
+        if delta.name == reals_var or delta.name.startswith(reals_var + "__BOUND"):
             delta = Independent(delta, reals_var, bint_var)
             deltas = joint.deltas[:i] + (delta,) + joint.deltas[1+i:]
             discrete = joint.discrete
--- a/funsor/terms.py
+++ b/funsor/terms.py
@@ -21,18 +21,16 @@ from funsor.six import getargspec, singledispatch
 
 
 def substitute(expr, subs):
-    if isinstance(subs, dict):
+    if isinstance(subs, (dict, OrderedDict)):
         subs = tuple(subs.items())
     assert isinstance(subs, tuple)
 
-    base_interpretation = interpreter._INTERPRETATION
-
+    @interpreter.interpretation(interpreter._INTERPRETATION)  # use base
     def subs_interpreter(cls, *args):
-        with interpreter.interpretation(base_interpretation):
-            expr = cls(*args)
-            fresh_subs = tuple((k, v) for k, v in subs if k in expr.fresh)
-            if fresh_subs:
-                expr = interpreter.debug_logged(expr.eager_subs)(fresh_subs)
+        expr = cls(*args)
+        fresh_subs = tuple((k, v) for k, v in subs if k in expr.fresh)
+        if fresh_subs:
+            expr = interpreter.debug_logged(expr.eager_subs)(fresh_subs)
         return expr
 
     with interpreter.interpretation(subs_interpreter):
--- a/funsor/terms.py
+++ b/funsor/terms.py
@@ -46,7 +46,7 @@ def alpha_convert(expr):
     new_values = []
     for v in expr._ast_values:
         v = substitute(v, alpha_subs)
-        if isinstance(v, str):
+        if isinstance(v, str) and v not in expr.fresh:
             v = alpha_subs.get(v, v)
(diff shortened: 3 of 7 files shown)


fritzo:
Note the numpy code is very out-of-date. It might be easier to either update that first (cc @neerajprad) or delete it and fork `funsor.torch` again.

neerajprad:
> It might be easier to either update that first (cc @neerajprad) or delete it and fork funsor.torch again.

@eb8680, @fritzo - I have not been able to keep up with funsor development, and I think maintaining the numpy backend at this stage is probably only adding more friction and slowing down things. My suggestion would be to remove the module (it is pretty self-contained) for now, and I'm happy to do so myself. I can reuse this work and pick it up later with a more fleshed-out plan if and when we would like to explore JAX integration. What do you think?

eb8680:
I think it's fine to leave the numpy module as is - the problem is with the stack-based interpreter, which is trying to hash numpy arrays.  I can fix that.  We should certainly revisit Numpyro/JAX integration at some point, though.

funsor/delta.py
```diff
@@ -129,14 +129,14 @@ def eager_add(op, lhs, rhs):
 
 @eager.register(Independent, Delta, str, str)
 def eager_independent(delta, reals_var, bint_var):
-    if delta.name == reals_var:
+    if delta.name == reals_var or delta.name.startswith(reals_var + "__BOUND"):
```
fritzo: Can you explain under what circumstances `delta` should substitute into a different variable? This seems fishy on first look.

funsor/joint.py
```diff
@@ -234,7 +234,7 @@ def eager_joint(deltas, discrete, gaussian):
 @eager.register(Independent, Joint, str, str)
 def eager_independent(joint, reals_var, bint_var):
     for i, delta in enumerate(joint.deltas):
-        if delta.name == reals_var:
+        if delta.name == reals_var or delta.name.startswith(reals_var + "__BOUND"):
```
fritzo: ditto: can you explain this?

funsor/terms.py
```diff
@@ -20,34 +20,21 @@
...
+        subs = tuple(subs.items())
+    assert isinstance(subs, tuple)
 
+    @interpreter.interpretation(interpreter._INTERPRETATION)  # use base
+    def subs_interpreter(cls, *args):
+        expr = cls(*args)
+        fresh_subs = tuple((k, v) for k, v in subs if k in expr.fresh)
+        if fresh_subs:
+            expr = interpreter.debug_logged(expr.eager_subs)(fresh_subs)
+        return expr
```
fritzo: I forget the `@interpreter.interpretation(-)` semantics. Does this first call `expr.eager_subs` and then pass the result for further reinterpretation by the base `interpreter._INTERPRETATION` captured above? If so then :+1:

funsor/terms.py
```diff
@@ -59,7 +46,7 @@ def alpha_convert(expr):
     new_values = []
     for v in expr._ast_values:
         v = substitute(v, alpha_subs)
-        if isinstance(v, str):
+        if isinstance(v, str) and v not in expr.fresh:
```
fritzo: Can you explain when the new condition is triggered? Thanks for the help.

funsor/delta.py
eb8680: Yes, there are some odd-looking changes here because the semantics of `Independent` are complicated (although I would argue that the changes in this PR clarify the type of `Independent`).  `Independent` binds `reals_var` in its input expression, leading to alpha-renaming, and introduces a fresh variable with the same name but a different type.  This condition is checking if `delta` contains the *bound* variable originally named `reals_var`, which may or may not have undergone alpha-conversion.

funsor/terms.py
eb8680: Since `Independent` has a fresh variable with the same name as a bound one, this check prevents `alpha_convert` from renaming `Independent`'s `reals_var` attribute.

funsor/terms.py
eb8680: The base interpretation is invoked in the line above (`expr = cls(*args)`).

test/test_numpy.py
```diff
@@ -5,10 +5,18 @@
 
 import funsor
 from funsor import Number, Variable, bint, reals
+from funsor.interpreter import _USE_TCO
 from funsor.numpy import Array
 from funsor.testing import assert_equiv, check_funsor, random_array
 
 
+# FIXME rewrite stack-based interpreter to be compatible with unhashable data
```
fritzo: Do you plan to use the same pattern we used when cons-hashing?
https://github.com/pyro-ppl/funsor/blob/ba0dfcb4ff0fbe7a8a14866bf178a2ef2b290feb/funsor/terms.py#L83

eb8680: Yes, something like that.  Note that this only matters for numpy arrays; every other data term works fine.

funsor/delta.py
fritzo: Hmm do you think it would be less error-prone to change `Independent` to require two user-provided names, one for the bound and one for the fresh? If you think it's fine as-is could you just add a little comment noting what you explained to me?

funsor/terms.py
fritzo: OK, let's just keep this edge case pattern in mind.

fritzo:
OK feel free to merge, or let me know if you make other changes and need another review.

PR #157, opened 2019-07-13
Add a single normal form for sum-product contractions

### LINKED ISSUE #156: Unify multilinear operations
Funsor currently contains many different representations of multilinear operations (e.g. `Joint`, `Affine`, `Finitary`, `Contract`, `Integrate`, `Binary`, `Reduce`, `einsum`, `sum_product`, ...) that approximately duplicate lots of logic and don't always play well together.

In a sequence of refactoring PRs, I would like to unify most of the duplicated linear algebra or normalization logic in the codebase while preserving all existing test cases and computational properties.  Specifically, I would like to:
1. #157 Add a `Contraction` normal form representing finitary sum-product operations with eager definitions implemented entirely in terms of unary and binary sum-product operations, and a `normalize` interpretation putting multilinear expressions into a DNF-like normal form
2. #157 Insert `normalize` as the underlying interpretation to which `eager` defaults, displacing `reflect` (optionally at first, using an environment variable as with the changes in #149).
3. #173 Remove `Affine` and replace with `Contraction` while preserving all existing test cases and functionality
4. #165 Simplify the optimizer by removing `Finitary`, `Contract`, `funsor.contract._simplify_contract`, and the `associate`, `distribute` and `desugar` interpretations used internally, and making it work with `normalize` and `Contraction`s directly
5. #169 Simplify `Integrate` by removing the duplicate functionality of `funsor.integrate._simplify_integrate`
6. #169 Remove `Joint` and replace with `Contraction` and `MultiDelta` while preserving all `Joint` test cases and computational properties.  This will also further simplify `eager`, since most of `Joint`'s functionality is already contained in the first PR.
7. #169 Remove `Delta` entirely in favor of `MultiDelta` so that deltas are closed under addition/logaddexp like `Gaussian` and `Tensor`
10. Troubleshoot performance issues related to interleaved interpreters (see `test_gaussian.py::test_eager_subs`, `test_samplers.py::test_gaussian_distribution`, `test_terms.py::test_reduce_subset` for cases of severe slowdowns)
8. Simplify `eager` by removing all purely algebraic computations from `eager` that are now handled by `normalize`, such as `funsor.terms.eager.distribute_binary_reduce_other`, and narrowing the focus of `eager` to only performing eager computations that actually invoke backend ops.
9. Rename `Contraction` -> `Contract`, `MultiDelta` -> `Empirical`, remove `Delta` entirely, move patterns in `joint.py` to `integrate.py`
11. Make optimization the default (merge `eager_contraction_generic_tuple` and the optimizer generic `Contraction` pattern from #165 )

More speculative: can we naturally and generically handle plate and scan dimensions #70 in `normalize`?

Plan for code review/merging:
1. Finish PRs for steps 1-6 above
2. Review and merge those PRs into #157, in descending order (i.e. 3, 4, 5, 6, 7)
3. Merge the final version of #157 into `master`
4. Make normalization the default and address steps 8-10 in cleanup PRs

Non-goals:
- I do not plan to replace or remove `Unary`, `Binary`, `Integrate`, or `Reduce`
- I do not plan to change any `eager` definitions or require them to be written in terms of `Contraction`.  `normalize` should not be proscriptive about the specification of eager operations.

eb8680:
Resolves #156 

As outlined in #156 I'm using this as a staging branch for the refactoring PRs associated with the proposed changes there:
- [x] #157 Original PR ([diff](https://github.com/pyro-ppl/funsor/pull/157/files/6fa04c681c39630fb2edaf8099ce3e23f3c71856))
- [x] #173 Refactor Affine into Contraction (https://github.com/pyro-ppl/funsor/pull/157/commits/e20978978283998341f5ed75a755cebc76fa2bab)
- [x] #165 Refactor optimizer to use Contraction and normalize (https://github.com/pyro-ppl/funsor/pull/157/commits/a4b2ec29b10f6649d86cf9fee24d56e574b279c8)
- [x] #169 Refactor Joint into multivariate Delta and Contraction (https://github.com/pyro-ppl/funsor/pull/157/commits/d2c79b0c8cdab99b370ca160de53dc8d62469dd9)
- [x] Separate unfolding from normalization (https://github.com/pyro-ppl/funsor/pull/157/commits/ffdd38176567242fde59ab1f070488ab98cce72e)
- [x] Troubleshoot performance issues (mostly caused by repeated unfolding)
- [x] Use #212 for pattern matching
- [x] Remove obsolete patterns ([diff](https://github.com/pyro-ppl/funsor/pull/157/files/8714349a6fb18cde2b8c04e96b0c289c8187015f..e59d85fa5a0db52e478827141eec95ada4a52645))
- [x] Update `MultiDelta` to keep separate `log_density`s for each `point` (https://github.com/pyro-ppl/funsor/pull/157/commits/30cbcb1c13806991cdeb70e5618799b60cf844b9)
- [x] Fix regressions in `adjoint` and `einsum`
- [x] Add missing patterns for `minipyro`
- [x] Update bound inference tests for `Number`

Original description:
This PR takes a first step towards the goals in #156 by adding two things:
1. A `Contraction` term representing finitary sum-product operations that takes advantage of `multipledispatch.Variadic` for pattern matching.  This is also necessary for #72.
2. A `normalize` interpretation that uses associative and distributive properties to put linear expressions into something like DNF, grouping operations into single `Contraction` terms where possible.

As a demonstration, I have included patterns and tests illustrating how `Joint` is equivalent to interleaving `eager` and `normalize`.

Tested:
- I copied smoke tests and test cases verbatim from `test_joint.py`, `test_gaussian.py`, `test_affine.py` and `test_einsum.py`


fritzo:
Looks like the [test failure](https://travis-ci.com/pyro-ppl/funsor/jobs/215836646#L3109) is an unreduced `0. + Gaussian(...)`, the same error we noticed last night in #161. This may be due to some unrelated missing pattern that only triggered after benign refactoring of patterns in this PR.

eb8680:
I'll start removing things in followup PRs

funsor/terms.py
```diff
@@ -64,6 +64,12 @@ def reflect(cls, *args):
     Construct a funsor, populate ``._ast_values``, and cons hash.
     This is the only interpretation allowed to construct funsors.
     """
+    if len(args) > len(cls._ast_fields):
```
fritzo: Does this logic need to be dulicated in `memoize`?

eb8680: Yes, I think so, good catch

funsor/cnf.py
```diff
@@ -0,0 +1,340 @@
...
+from funsor.gaussian import Gaussian
+from funsor.interpreter import recursion_reinterpret
+from funsor.ops import AssociativeOp, DISTRIBUTIVE_OPS
+from funsor.terms import Align, Binary, Funsor, Number, Reduce, Subs, Unary, Variable, \
+    eager, normalize, to_funsor
+from funsor.torch import Tensor
+
+
+class AnyOp(AssociativeOp):
+    """Placeholder associative op that unifies with any other op"""
```
eb8680: Rename to `NoOp` or something else

funsor/delta.py
```diff
@@ -24,126 +19,10 @@
...
-
-@eager.register(Integrate, Delta, Funsor, frozenset)
-@integrator
-def eager_integrate(delta, integrand, reduced_vars):
-    assert delta.name in reduced_vars
-    integrand = Subs(integrand, ((delta.name, delta.point),))
-    log_measure = delta.log_density
-    reduced_vars -= frozenset([delta.name])
-    return Integrate(log_measure, integrand, reduced_vars)
+def Delta(name, point, log_density=0):
```
eb8680: Get rid of this, rename `MultiDelta` to `Delta`

fritzo:
I'm seeing some breakage in my BART experiment. To reproduce
```sh
git clone git@github.com:pyro-ppl/sandbox
cd sandbox/2019-08-time-series/bart
git checkout funsor-bart
make test2
```
Inference dies with error message
```
TODO implement Cat.eager_subs for <class 'funsor.terms.Slice'>
```
but I believe that is due to either a missing pattern or overly-eager evaluation.
Currently investigating...

eb8680:
> I believe that is due to either a missing pattern or overly-eager evaluation

It's probably a missing pattern for a `Cat` of `Joint`-like `Contraction`s

funsor/joint.py
```diff
@@ -1,268 +1,24 @@
...
-            gaussian = joint.gaussian
-            if bint_var in gaussian.inputs:
-                gaussian = gaussian.reduce(ops.add, bint_var)
-            return Joint(deltas, discrete, gaussian)
-
-    return None  # defer to default implementation
-
-
-@dispatch(str, str, Variadic[Gaussian, Joint])
+@dispatch(str, str, Variadic[(Gaussian, GaussianMixture)])
```
fritzo: Is this pattern now too narrow?

eb8680: It shouldn't be, according to [this check in `master`](https://github.com/pyro-ppl/funsor/pull/157/files/8eb0930dcd1435f869cf51a66be50326974f0589#diff-48e50907b44e8bb24d6ee7485225e0ebL291)

eb8680: I think this might be caused by `normalize` being a bit too aggressive with pushing down `exp`, so that we get terms that look like `Contraction[..., Tuple[Unary[Exp, Gaussian], Tensor]]`

funsor/gaussian.py
```diff
@@ -579,11 +579,10 @@ def eager_neg(op, arg):
 
 
 @eager.register(Integrate, Gaussian, Variable, frozenset)
-@integrator
 def eager_integrate(log_measure, integrand, reduced_vars):
     real_vars = frozenset(k for k in reduced_vars if log_measure.inputs[k].dtype == 'real')
-    if real_vars:
-        assert real_vars == frozenset([integrand.name])
+    if real_vars == frozenset([integrand.name]):
```
fritzo: I'd rather error than be lazy here:
```py
if real_vars:
    if real_vars != frozenset([integrand.name]):
        raise NotImplementedError('TODO')
    ...
```
Also it appears there is an error in that int vars is ignored. I believe we should
```py
result = Tensor(data, inputs)
return result.reduce(ops.add, reduced_vars - real_vars)
```

funsor/joint.py
fritzo: Can we simply remove `Unary[Contraction]` simplification from `normalize`?

funsor/joint.py
```diff
@@ -286,11 +42,11 @@ def eager_cat_homogeneous(name, part_name, *parts):
...
-        elif isinstance(part, Joint):
-            if part.deltas:
-                raise NotImplementedError("TODO")
-            discrete = align_tensor(int_inputs, part.discrete).expand(shape)
-            gaussian = part.gaussian
+        elif issubclass(type(part), GaussianMixture):  # TODO figure out why isinstance isn't working
+            discrete, gaussian = part.terms[0], part.terms[1]
+            discrete = align_tensor(int_inputs, discrete).expand(shape)
+        else:
+            raise ValueError
```
fritzo: Should be `NotImplementedError('TODO')`

funsor/joint.py
eb8680: Hmm, it might be easier to add `normalize` patterns that prevent oversimplification e.g.
```py
@normalize.register(Unary, ExpOp, GaussianMixture)
def normalize_unary_joint(op, arg):
    return None
```

funsor/gaussian.py
eb8680: > I'd rather error than be lazy here:

The laziness here is necessary to allow `eager` to generically eliminate variables that only appear on one side (i.e. the job currently done by `@integrator`)

fritzo: Hmm, can you explain this in person? I didn't realize we could stage multiple `eager` rules in sequence, which we did with `@integrator`

eb8680: This comes from interleaving `eager` and `normalize` - if an `eager` pattern returns `None` and the resulting term isn't normalized, `normalize` will rewrite it and the expression inside the `normalize` pattern will again be evaluated under `eager`

Makefile
```diff
@@ -17,6 +17,7 @@ format: FORCE
 
 test: lint FORCE
 	pytest -v -n auto test/
+	FUNSOR_NORMALIZE=1 pytest -v test/test_cnf.py
```
fritzo: let's remove the `NORMALIZE` option to reduce our feature flag set. i.e. normalize by default.

funsor/cnf.py
```diff
@@ -0,0 +1,340 @@
...
+
+@eager.register(Contraction, AssociativeOp, AssociativeOp, frozenset, Variadic[Funsor])
+def eager_contraction_generic_to_tuple(red_op, bin_op, reduced_vars, *terms):
+    return eager(Contraction, red_op, bin_op, reduced_vars, tuple(terms))
+
+
+@eager.register(Contraction, AssociativeOp, AssociativeOp, frozenset, tuple)
+def eager_contraction_generic_recursive(red_op, bin_op, reduced_vars, terms):
+
+    # push down leaf reductions
```
fritzo: Do I understand correctly that this will be unable to reduce variables that occur in multiple factors of an integrand, which is strictly weaker than `@integrator`? Or are they made equivalent via a distributivity rule?

eb8680: They're made equivalent by the second part of this pattern

funsor/delta.py
```diff
@@ -188,7 +67,142 @@ def solve_unary(op, arg, y):
     return name, point, log_density
 
 
+class MultiDelta(Funsor):
```
fritzo: This interface seems verbose, but let's fix later later.

funsor/delta.py
```diff
@@ -188,7 +67,142 @@ def solve_unary(op, arg, y):
...
+        if not names or names == tuple(n for n, p in self.terms):
+            return self
+
+        new_terms = sorted(self.terms, key=lambda t: names.index(t[0]))
+        return MultiDelta(new_terms)
+
+    def eager_subs(self, subs):
+        terms = OrderedDict(self.terms)
+        new_terms = terms.copy()
+        log_density = to_funsor(0., self.output)
```
fritzo: simplify to `Number(0)`. We always know `self.output == reals()`

funsor/delta.py
```diff
@@ -188,7 +67,142 @@ def solve_unary(op, arg, y):
...
+            result = MultiDelta(tuple(result_terms)) + scale if result_terms else scale
+            return result.reduce(op, reduced_vars - self.fresh)
+
+        if op is ops.add:
+            raise NotImplementedError("TODO Implement ops.add to simulate .to_event().")
+
+        return None  # defer to default implementation
+
+    def unscaled_sample(self, sampled_vars, sample_inputs):
+        # raise NotImplementedError("TODO implement sample for particle indices")
```
fritzo: Maybe we should err if the wrong variables are being sampled?
```py
if not sampled_vars.issubset(self.fresh):
    raise ValueError
```

funsor/gaussian.py
```diff
@@ -579,11 +579,10 @@ def eager_neg(op, arg):
...
+        # assert real_vars == frozenset([integrand.name])
```
fritzo: nit: remove commented code

funsor/interpreter.py
```diff
@@ -20,6 +20,7 @@
 
 _INTERPRETATION = None  # To be set later in funsor.terms
 _USE_TCO = int(os.environ.get("FUNSOR_USE_TCO", 0))
+_NORMALIZE = int(os.environ.get("FUNSOR_NORMALIZE", 1))
```
fritzo: remove

funsor/minipyro.py
```diff
@@ -28,31 +28,43 @@
 # torch Distributions are samplers. This class is a compatibility wrapper
 # between the two. It is used only internally in the sample() function.
 class Distribution(object):
-    def __init__(self, funsor_dist):
+    def __init__(self, funsor_dist, sample_inputs=None):
```
fritzo: Why are these changes needed? Can we move them to a separate PR?

funsor/optimizer.py
```diff
@@ -1,197 +1,100 @@
...
+
+
+@interpreter.dispatched_interpretation
+def unfold(cls, *args):
+    result = unfold.dispatch(cls, *args)
+    if result is None:
+        result = normalize.dispatch(cls, *args)
     if result is None:
-        result = lazy(cls, *args)
+        result = lazy(cls, *args)  # lazy or reflect?
```
fritzo: either add FIXME or XXX or remove the comment

funsor/optimizer.py
```diff
@@ -1,197 +1,100 @@
...
+unfold.register(Contraction, AssociativeOp, AssociativeOp, frozenset, Variadic[Funsor])(
+    lambda r, b, v, *ts: unfold(Contraction, r, b, v, tuple(ts)))
+
+
+@interpreter.dispatched_interpretation
 def optimize(cls, *args):
     result = optimize.dispatch(cls, *args)
     if result is None:
-        result = lazy(cls, *args)
+        result = eager(cls, *args)
```
fritzo: Does this mean we can eliminate lots of `reinterpret` calls in examples?

funsor/optimizer.py
```diff
@@ -1,197 +1,100 @@
...
-@eager.register(Reduce, AssociativeOp, Binary, frozenset)
-def optimize_reduce_binary_exp(op, arg, reduced_vars):
-    if op is not ops.add or arg.op is not ops.mul or \
-            not isinstance(arg.lhs, Unary) or arg.lhs.op is not ops.exp:
-        return None
-    return Integrate(arg.lhs.arg, arg.rhs, reduced_vars)
+@optimize.register(Contraction, AssociativeOp, AssociativeOp, frozenset, Funsor, Funsor)
+@optimize.register(Contraction, AssociativeOp, AssociativeOp, frozenset, Funsor)
+def eager_contract_base(red_op, bin_op, reduced_vars, *terms):
+    return None  # eager.dispatch(Contraction, red_op, bin_op, reduced_vars, *terms)
```
fritzo: what does this comment mean?

funsor/pyro/convert.py
```diff
@@ -195,12 +195,12 @@ def funsor_to_cat_and_mvn(funsor_, ndims, event_inputs):
...
         dimension ranging over mixture components.
     """
-    assert isinstance(funsor_, Joint), funsor_
+    assert isinstance(funsor_, Contraction), funsor_
     assert sum(1 for d in funsor_.inputs.values() if d.dtype == "real") == 1
     assert event_inputs, "no components name found"
-    assert not funsor_.deltas
-    discrete = funsor_.discrete
-    gaussian = funsor_.gaussian
+    # assert not funsor_.deltas
```
fritzo: Fix this assertion.

funsor/pyro/distribution.py
```diff
@@ -67,23 +67,23 @@ def _sample_delta(self, sample_shape):
                 if shape[dim] > 1:
                     sample_inputs[DIM_TO_NAME[dim]] = bint(shape[dim])
         delta = self.funsor_dist.sample(frozenset({"value"}), sample_inputs)
-        if isinstance(delta, Joint):
-            delta, = delta.deltas
-        assert isinstance(delta, Delta)
+        if isinstance(delta, Contraction):
```
fritzo: Add an assertion that `len(delta.terms) == `. This was previously checked by the destructured assignment.

test/pyro/test_convert.py
```diff
@@ -105,7 +105,7 @@ def _rand(batch_shape, *event_shape):
     # Evaluate via two different patterns.
     expected = f(x=value_x)(y=value_y)
     actual = f(y=value_y)(x=value_x)
-    assert_close(actual, expected, atol=1e-5, rtol=1e-5)
+    assert_close(actual, expected, atol=1e-5, rtol=2e-4)
```
fritzo: a 20x error increase seems like a lot, any idea why?

eb8680: I also found this puzzling since it wasn't clear how the changes in this PR would affect `AffineNormal`, maybe it's due to nondeterminism somewhere in the test setup?

funsor/joint.py
fritzo: I guess I don't see why we'd want to move unary inside. Unary ops are pointwise and are generally cheaper to apply outside of reductions.

funsor/delta.py
eb8680: This is identical to the current behavior of `Joint.unscaled_sample`, and I believe it's necessary for `monte_carlo` in its current incarnation to work since it repeatedly `.sample`s until it stops making progress.  We should definitely revisit in Monte Carlo cleanup/refactoring PRs.

fritzo: Well either way, please remove commented code.

funsor/optimizer.py
eb8680: I think so, unless you want the more aggressive rewrite performed by `apply_optimizer`

funsor/optimizer.py
eb8680: It's just dead code from the switch from `lazy` to `eager` in `optimize`

eb8680: I'll remove

fritzo:
It seems like you really want `Reduce(Finitary)` rather than this monolithic `Contraction` and NullOp. I think that would be much clearer to library users and contributors.

Could we replace `Binary` with `Finitary`?

fritzo:
It is also tempting to replace `MultiDelta` with a `Finitary[Delta]` and require a normal form like
```
Reduce[SumOp, Finitary[ProdOp, Finitary[ProdOp, Delta], Tensor, Gaussian]]
```

eb8680:
I don't understand this notation, is Finitary variadic by default?

eb8680:
Hmm, I disagree. Having a single representation is a design choice that
prevents a combinatorial blowup of patterns and of interpreter stack size,
and representing sum product contractions in a single op is important for
performance. I also don't expect users to interact with Contraction - they
only have to write patterns for Binary and Reduce, or for pairwise
contractions.

funsor/joint.py
```diff
@@ -314,224 +70,118 @@ def eager_cat_homogeneous(name, part_name, *parts):
...
-    sampled_log_measure = log_measure.sample(reduced_vars, monte_carlo.sample_inputs)
-    if sampled_log_measure is not log_measure:
-        reduced_vars = reduced_vars | frozenset(monte_carlo.sample_inputs)
-        return Integrate(sampled_log_measure, integrand, reduced_vars)
+@eager.register(Independent,
+                (Contraction[NullOp, ops.AddOp, frozenset, Tuple[Delta, Union[Number, Tensor], Gaussian]],
+                 Contraction[NullOp, ops.AddOp, frozenset, Tuple[Delta, Union[Number, Tensor, Gaussian]]]),
+                str, str, str)
+def eager_independent_joint(joint, reals_var, bint_var, diag_var):
+    if diag_var not in joint.inputs:
```
fritzo: I believe this should be
```py
if diag_var not in joint.terms[0].fresh:
    ...
```

funsor/joint.py
eb8680: > I guess I don't see why we'd want to move unary inside

The main use case here (as in `optimize` and `Contract`/`Integrate` already in `master`) is in rewriting integrals of large joint distributions.  Pushing the unary ops down to the leaves makes pattern matching easier, simplifies the normal form and lets us reuse the optimizer for integrals.  In these situations it's unlikely that unary ops would actually be evaluated, since they would eventually be rewritten back into `Integrate`s, KL divergences, etc.

I don't see a reason to change this unless we also want to completely decouple `Integrate` from `Contraction` (i.e. never rewrite `Integrate` to `Contraction` or vice versa), which I suppose we could discuss separately.

funsor/pyro/convert.py
```diff
@@ -195,12 +196,12 @@ def funsor_to_cat_and_mvn(funsor_, ndims, event_inputs):
...
     """
-    assert isinstance(funsor_, Joint), funsor_
+    assert isinstance(funsor_, Contraction), funsor_
     assert sum(1 for d in funsor_.inputs.values() if d.dtype == "real") == 1
     assert event_inputs, "no components name found"
-    assert not funsor_.deltas
-    discrete = funsor_.discrete
-    gaussian = funsor_.gaussian
+    assert not any(isinstance(v, Delta) for v in funsor_.terms)
+    discrete = [v for v in funsor_.terms if isinstance(v, Tensor)][0]  # funsor_.discrete
```
fritzo: remove obsolete comment

eb8680:
> Let's try to keep future PRs small and targeted.

Agreed, but I'm not sure replacing `Joint` could have been done in a much more small and targeted way.  If this version is too invasive or is continuing to break experiments, let's just think of it as a protoype, close it, and revisit the ideas in the form of a design doc at some point after the first release.

fritzo:
Let's just merge this and forge ahead.

Issue #109, opened 2019-03-28
Implement an Expectation funsor


fritzo:
Expectation is combines ideas of TraceEnum_ELBO's `Dice` and Funsor's `Integrate` and `plate`:
```py
Expectation((p1, p2, ..., pn),  # log measures
            (c1, c2, ..., cn),  # costs
            sum_vars=frozenset(...),
            prod_vars=frozenset(...))
``` 
The optimizer should perform one instance of Tensor Variable Elimination for each cost term (against all p terms).

If `prod_vars` is empty (i.e. no plates), then
```py
Expectation((p1, p2, ..., pn),
            (c1, c2, ..., cn),
            sum_vars=sum_vars,
            prod_vars=frozenset())  # empty
  = Integrate(p1 + p2 + ... + pn,
              c1 + c2 + ... + cn,
              reduced_vars=sum_vars)
  = Contract(ops.exp(p1 + p2 + ... + pn),
             c1 + c2 + ... + cn,
             reduced_vars=sum_vars)
```

Issue #12, opened 2019-02-16
Add SMC example


fritzo:
We can use `pyro.markov` to trigger an optional resampling step based on perplexity. This behavior would be handled by an inference handler.

## Outline
- [ ] Add storyboard example using `pyro.markov` to perform SMC
- [ ] Implement SMC handler to interpret `pyro.sample` statements as multi-particle samples
- [ ] Implement resampling by interpreting `pyro.markov` statements

fritzo:
@eb8680 Do I understand correctly that this is blocked by #52 and #50 ?

eb8680:
Yes, I think so.  We might also need to update minipyro.

PR #160, opened 2019-07-14
Add a memoize interpreter/context manager

### LINKED ISSUE #80: Implement a memoize backend that takes another backend
This is the first handler that is parametrized by another handler.

~~This is required for efficient use of nn.Modules with multiple outputs #79 .~~

ref: http://okmij.org/ftp/tagless-final/sharing/index.html

eb8680:
Resolves #80.  Also part of fixing `monte_carlo` and enabling use cases like #12 and #109. 

This PR adds a context manager `funsor.memoize` that exploits cons-hashing to do common subexpression elimination, similar to `opt_einsum`'s `shared_intermediates`:

```py
with memoize(cache={}) as cache:
   x = some_funsor_expression()
   y = some_funsor_expression()

assert x is y
```
Remaining tasks:
- [x] More test cases dealing with alpha conversion

Tested:
- A couple of complete sharing test for various `einsum` expressions, of the above form
- Test demonstating failure of `sample` memoization


test/test_memoize.py
```diff
@@ -0,0 +1,90 @@
...
+def test_memoize_sample():
+
+    with memoize():
+        j1 = Normal(0, 1, 'x')
+        j2 = Normal(0, 1, 'x')
+        x1 = j1.sample(frozenset({'x'}))
+        x12 = j1.sample(frozenset({'x'}))
+        x2 = j2.sample(frozenset({'x'}))
+
+    assert j1 is j2
```
eb8680: This assertion fails, which doesn't make sense...

eb8680: Ah, looks like this was caused by conversion of numbers to tensors.

test/test_memoize.py
```diff
@@ -0,0 +1,97 @@
...
+@pytest.mark.parametrize('backend', ['torch', 'pyro.ops.einsum.torch_log'])
+@pytest.mark.parametrize('einsum_impl', [einsum, naive_plated_einsum])
+@pytest.mark.parametrize('same_lazy', [True, xfail_param(False, reason="issue w/ alpha conversion?")])
+def test_einsum_complete_sharing(equation, plates, backend, einsum_impl, same_lazy):
+    inputs, outputs, sizes, operands, funsor_operands = make_einsum_example(equation)
+
+    with interpretation(reflect):
+        lazy_expr1 = einsum_impl(equation, *funsor_operands, backend=backend, plates=plates)
+        lazy_expr2 = lazy_expr1 if same_lazy else \
+            einsum_impl(equation, *funsor_operands, backend=backend, plates=plates)
```
eb8680: Creating two different optimized lazy expressions here leads to different alpha-converted names for bound variables, preventing sharing.  This may be caused by the other interpretations invoked by `einsum` rather than by alpha conversion , however - more targeted tests needed.

eb8680:
This is OK to review/merge, I'll deal with other problems as they come up

funsor/memoize.py
```diff
@@ -0,0 +1,26 @@
...
+        key = (cls,) + tuple(id(arg) if not isinstance(arg, Hashable) else arg for arg in args)
+        if key not in cache:
+            cache[key] = cls(*args)
+        return cache[key]
+
+    try:
+        with interpreter.interpretation(memoize_interpretation):
+            yield cache
+    finally:
+        assert cache is not None  # XXX do anything more here?
```
fritzo: Why is the try-finally needed? The cache can never be `None` since it hasn't been modified from line 13.

test/test_memoize.py
fritzo: Hmm, I wonder if we should reduce the amount of renaming done by alpha binding...

eb8680: I'm not quite sure what's going on here, I think it's more likely due to the use of nested interpreters in the optimized `einsum`

PR #166, opened 2019-07-23
Add some interface documentation

fritzo:
I've updated existing docs in preparation for adding another interface for using Funsor in Pyro.

## Tested
- ran `make docs` locally and inspected results


eb8680:
Looks like test failure is due to the removal of Python 3 support from Pyro.

Issue #70, opened 2019-03-10
Support plates and convolution dims in Funsors?


fritzo:
@eb8680 suggested adding metadata to `Domain` objects or `.input` dicts to declare which inputs are plate indices. This is similar to @jpchen's suggestion of a `Cat` funsor, but would also support cat-broadcasting. From a mathematical perspective, adding funsors with different plate indices would be a direct sum.

I believe the best way to implement this is to first implement correct math in `funsor.minipyro.log_joint` and then to refactor it into `Domain` or `Funsor`.

## Tasks
- [x] #56 sketch plate math in `funsor.minipyro.log_joint`
- [x] #57 fix minipyro and add tests
- [ ] #109 add plate-aware contract/expectation ops
- [ ] Add a funsor `Conv(x, time_dim, curr_dim, next_dim, new_dim)` satisfying
     ```py
     Conv(f, "time", "curr", "next", "curr")(curr=x)
         = f(curr=x(time=arange("time", f.inputs["time"].dtype - 1)),
             next=x(time=arange("time", f.inputs["time"].dtype - 1) + 1))
     ```
- [x] #161 Implement log(time) convolution op following [this paper](https://arxiv.org/abs/1905.13002)
- [x] #292 Generalize `sequential_sum_product` to handle longer time lags
- [ ] Support time dims in `partial_sum_product` (see [convolve-method branch](https://github.com/pyro-ppl/funsor/compare/convolve-method?expand=1))
- [ ] Add a state space example using `Conv` applied to a contiguous stacked nn output

eb8680:
Looking at the Toeplitz and Kronecker `LazyTensor`s in GPyTorch might be helpful too.

PR #167, opened 2019-07-24
Implement log-time Markov sum product algorithm

fritzo:
Addresses #70 #177
Pair coded with @eb8680 

This aims to generalize `sequential_sum_product()` from simple contractions to arbitrary contractions with different window size. This aims to be maximally general and should be able to be integrated into the TVE algorithm `partial_sum_product()` (as in the [convolve-method branch](https://github.com/pyro-ppl/funsor/compare/convolve-method?expand=1)).

## Design Questions
- [ ] What problem class should we address? E.g. should this assume that the components have already been decomposed by a connected components algorithm? I've started to enumerate test cases, and I think we should spend a many hours whiteboarding examples to understand the problem class (as we did with TVE).
- [ ] How can we pass around the metadata of time lags? This PR attempts to do so via name mangling of `.inputs.keys()`, but that approach won't currently work because alpha binding mangles names in a different way. Indeed the alpha mangling appears to thwart other nearby designs such as passing in a side-information dict indicating which names correspond to which lags. One possible solution is to make alpha renaming less intrusive. @eb8680 can you think about possible solutions?

### DIFF
--- a/funsor/sum_product.py
+++ b/funsor/sum_product.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import, division, print_function
 
+import re
 from collections import OrderedDict, defaultdict
 
 import torch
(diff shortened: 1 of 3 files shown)


eb8680:
Abandoning this design (time lags > 1)

PR #169, opened 2019-07-24
Refactor Joint into multivariate Delta and Contraction

eb8680:
This PR removes `Joint` in favor of a normalized multivariate `Delta` which is closed under addition and `Contraction` from #156 and internally represents all log-densities with a single tensor.  It also deletes lots of duplicate test code made redundant by the removal of `Joint`, while preserving behavior in all existing tests that exercised `Joint`.

This is quite a large PR with lots of changes since `Joint` was used throughout the codebase.  I'll add a review/merge plan for this series of PRs to the master issue #156.

It's almost complete, except for these remaining todos:
- [x] Add default pattern to `moment_matching` so that tests pass
- [x] Add a couple of remaining missing patterns for reduction and integration with `MultiDelta`
- [x] Fix minor type/shape bug in lazy advanced indexing
- [x] Add pattern for `Independent` for `MultiDelta`, `Contraction`


fritzo:
Can you confirm that `.unscaled_sample()` will still be defined on compound terms that result from distributions (and from marginalizing etc.)?

I forget the original set of terms for which `.sample()` is defined, and there have since been large changes. I'd like to ensure that `.sample()` is supported by our `FunsorDistribution`s as in #171 .

eb8680:
> Can you confirm that .unscaled_sample() will still be defined on compound terms that result from distributions (and from marginalizing etc.)?

Yes, via the definition of `unscaled_sample` for `Contraction`

eb8680:
All tests pass, so I'm going to merge this into #157 to keep the git history from getting any uglier.

Issue #159, opened 2019-07-14
Generate Distribution and RandomVariable wrapper classes programmatically


eb8680:
The distribution wrappers in `funsor.distributions` and the `RandomVariable` classes in #130 are currently specified manually with a bit of boilerplate on top of the base `Distribution` class for each distribution of interest.

We should switch to generating them entirely programmatically [as in Pyro](https://github.com/pyro-ppl/pyro/blob/dev/pyro/distributions/torch.py#L69), which would make it easier to support new distributions and new backends and allow us to reuse more existing infrastructure in `eager`, like samplers, `torch.distributions.constraint`s, and `TransformedDistribution`s.

PR #170, opened 2019-07-24
Add FunsorDistribution to wrap funsors for use in Pyro

fritzo:
This adds a basic `FunsorDistribution` class inheriting from `TorchDistribution`. The class is intended to wrap a `Funsor` as a `Distribution` object for use in Pyro.

I've created a new module `funsor.pyro` and I plan to create new submodules like `funsor.pyro.hmm` #171 containing various hidden-markov models (discrete, gaussian, slds, ...). Each of these distributions will use fancy funsor syntax internally, but will expose their inference algorithms as black-box `Distributions` with `.sample()` and `.log_prob()` methods. These distributions should be fully compatible with broadcasting, plates, and enumeration dims (which should pass right through these algorithms).

## Tested

- added unit tests for conversion helpers
- added a basic test creating a `Categorical` distribution from a funsor


funsor/pyro/distribution.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def rsample(self, sample_shape=torch.Size()):
+        delta = self._sample_delta(sample_shape)
+        assert not delta.log_prob.requires_grad, "distribution is not fully reparametrized"
+        ndims = len(sample_shape) + len(self.batch_shape) + len(self.event_shape)
+        value = funsor_to_tensor(delta.point, ndims=ndims)
+        return value
+
+    def expand(self, batch_shape, _instance=None):
+        raise NotImplementedError("TODO")
```
eb8680: `.expand` will be necessary for most Pyro use cases, right?  Are you planning to implement this generically in a followup PR, or case-by-case for specific `FunsorDistribution` subclasses?  Also, can we use `funsor.Independent` to implement `FunsorDistribution.to_event`?

funsor/pyro/convert.py
```diff
@@ -0,0 +1,81 @@
...
+            batch_shape[dim] = size
+        batch_shape = torch.Size(batch_shape)
+        tensor = tensor.reshape(batch_shape + funsor_.output.shape)
+    if ndims != tensor.dim():
+        tensor = tensor.reshape((1,) * (ndims - tensor.dim()) + tensor.shape)
+    assert tensor.dim() == ndims
+    return tensor
+
+
+def dist_to_funsor(pyro_dist, reinterpreted_batch_ndims=0):
```
eb8680: We should be able to implement `dist_to_funsor` generically and use it to replace the current `funsor.distributions` as suggested in #159, with a few extra `eager` patterns for `Categorical`, `Normal`/`MultivariateNormal`, and `Delta`

.travis.yml
```diff
@@ -14,8 +14,10 @@ install:
     - pip install -U pip
     - pip install https://download.pytorch.org/whl/cpu/torch-1.1.0-cp36-cp36m-linux_x86_64.whl
 
-    # Keep track of Pyro dev branch
-    - pip install https://github.com/pyro-ppl/pyro/archive/dev.zip
+    # This can be replaced with dev once https://github.com/pyro-ppl/pyro/pull/1978 merges.
+    - pip install https://github.com/pyro-ppl/pyro/archive/fix-categorical-sample-view.zip
```
eb8680: https://github.com/pyro-ppl/pyro/pull/1978 was just merged, you can revert this

funsor/pyro/distribution.py
fritzo: I'm planning to implement `.expand()` generically when I eventually need it.

Hmm yes, I think we might be able to use `funsor.Independent` for `.to_event()`.

funsor/pyro/convert.py
fritzo: In theory, yes. In practice, there is a mismatch between the dtype and shapes in funsor versus distributions. I welcome any effort to implement this generically, but I suspect every distribution will be a new edge case.

fritzo:
Thanks for reviewing!

PR #175, opened 2019-07-25
Use @singledispatch in dist_to_funsor; add more distributions

fritzo:
This improves `funsor.pyro.dist_to_funsor()` in a few ways:
- Supports `dist.Independent()` using `funsor.Independent` as suggested by @eb8680 
- Supports `dist.Bernoulli`
- Fixes two funsor bugs blocking `dist.Independent` (one in `Tensor`, one in `Indepenedent`)

This should be sufficient support for our first release.

## Tested
- added unit tests in test/pyro/test_convert.py

### DIFF
--- a/funsor/pyro/convert.py
+++ b/funsor/pyro/convert.py
@@ -1,10 +1,12 @@
 from collections import OrderedDict
+from functools import singledispatch
 
 import pyro.distributions as dist
 import torch
 
-from funsor.distributions import MultivariateNormal, Normal
+from funsor.distributions import BernoulliLogits, MultivariateNormal, Normal
 from funsor.domains import bint
+from funsor.terms import Independent
 from funsor.torch import Tensor
 
 # Conversion functions use fixed names for Pyro batch dims, but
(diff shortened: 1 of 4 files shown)


funsor/torch.py
```diff
@@ -418,7 +418,7 @@ def eager_getitem_tensor_tensor(op, lhs, rhs):
         rhs_data = rhs_data.reshape(rhs_data.shape + (1,) * (len(lhs.output.shape) - 1))
 
     # Perform advanced indexing.
-    target_dim = len(lhs.inputs) + op.offset
+    target_dim = lhs_data.dim() - len(lhs.output.shape) + op.offset
```
fritzo: This fixes a bug whereby `lhs.inputs` is no longer valid after `align_tensors()` above.

funsor/terms.py
```diff
@@ -1130,7 +1130,8 @@ def unscaled_sample(self, sampled_vars, sample_inputs):
         return Independent(fn, self.reals_var, self.bint_var)
 
     def eager_subs(self, subs):
-        subs = tuple((self.reals_var_bound, v[self.bint_var])
+        subs = tuple((self.reals_var_bound,
+                      to_funsor(v, self.inputs[k])[self.bint_var])
```
fritzo: This fixes a bug where `isinstance(v, str)`, causing `v[self.bint_var]` to error. The solution is to promote `v` to a `Variable`.

funsor/pyro/convert.py
```diff
@@ -65,28 +67,47 @@ def funsor_to_tensor(funsor_, ndims, event_inputs=()):
...
-        scale = tensor_to_funsor(pyro_dist.scale, event_inputs)
-        return Normal(loc, scale)
 
-    if isinstance(pyro_dist, dist.MultivariateNormal):
-        loc = tensor_to_funsor(pyro_dist.loc, event_inputs, 1)
-        scale_tril = tensor_to_funsor(pyro_dist.scale_tril, event_inputs, 2)
-        return MultivariateNormal(loc, scale_tril)
+@dist_to_funsor.register(dist.Categorical)
+def _categorical_to_funsor(pyro_dist, event_inputs=()):
+    return tensor_to_funsor(pyro_dist.logits, event_inputs + ("value",))
```
eb8680: Why not return a `funsor.distributions.Categorical` here, as you did with `dist.Normal` below?  If possible I would like to confine the distribution special-casing to a single point in the conversion pipeline, ideally within `funsor.distributions` (which already handles conversion of `Categorical`s to `Tensor`s and `Normal`s to `Gaussian`s)

funsor/pyro/convert.py
```diff
@@ -65,28 +67,47 @@ def funsor_to_tensor(funsor_, ndims, event_inputs=()):
...
 
+@singledispatch
 def dist_to_funsor(pyro_dist, event_inputs=()):
     """
     Convert a :class:`torch.distributions.Distribution` to a
     :class:`~funsor.terms.Funsor` .
     """
     assert isinstance(pyro_dist, torch.distributions.Distribution)
+    raise ValueError("Cannot convert {} distribution to a Funsor"
+                     .format(type(pyro_dist).__name__))
```
eb8680: Do you think it would be possible to use the `torch.distributions.Distribution.param_shape` attribute to handle `dist_to_funsor` conversion generically for the subset of distributions that implement it?

funsor/pyro/convert.py
fritzo: The reason is that `Categorical` currently requires `probs` rather than `logits`, whereas the rest of our HMM-like code works with logits. Eventually I'd like to fork `Categorical` into `CategoricalProbs`+`CategoricalLogits` as we've done with `Bernoulli`, but that seemed out of scope for this PR.

funsor/pyro/convert.py
fritzo: Hmm, I'd have to look at more special cases to generalize.

eb8680: Ok, no need to address it in this PR, just something to keep in mind

fritzo:
Thanks for reviewing!

Issue #162, opened 2019-07-17
Implement more advanced indexing operations


eb8680:
#161 and related efforts will require additional advanced indexing functionality.
- [ ] Implement `Cat` term backed by `torch.cat` for `Gaussian`, `Tensor`, `Delta`/`Joint`
- [ ] Implement `Slice` term that can be used instead of `arange` for affine indexing operations that do not trigger memory copies
- [ ] Implement interpretation of `Stack` backed by `torch.stack`

PR #176, opened 2019-07-25
Support Gaussian funsors in sequential_sum_product

fritzo:
Addresses #177 

This adds `Gaussian` support to `sequential_sum_product()`. This required:
- supporting `Gaussian` and `Joint` in our ad-hoc `Cat` funsor
- adding a `Joint.align()` method
- fixing typos in `sequential_sum_product()` (as @eb8680 and I noticed offline)

Note this means we can now implement asymptotically fast Kalman filter learning with funsor. (Provisos include: (1) only for full-rank observation distributions, (2) fast asymptotically, but with a big constant factor due to our broadcast-multiply-reduce implementation of gaussian_gemm).

## Tested
- added a grid of normal and multivariate normal tests to test_sum_product.py

### DIFF
--- a/funsor/sum_product.py
+++ b/funsor/sum_product.py
@@ -136,8 +136,6 @@ def Cat(parts, name):
             inputs[name] = part.inputs[name]  # typically a smaller bint
             int_inputs[name] = inputs[name]
             shape = tuple(d.size for d in int_inputs.values())
-            loc_shape = shape + (-1,)
-            precision_shape = shape + (-1, -1)
             if isinstance(part, Gaussian):
                 discrete = None
                 gaussian = part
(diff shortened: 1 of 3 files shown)


fritzo:
cc @fehiepsi

funsor/sum_product.py
```diff
@@ -165,23 +209,23 @@ def sequential_sum_product(sum_op, prod_op, trans, time, prev, curr):
 
     but does so efficiently in parallel in O(log(time)).
     """
-    assert isinstance(sum_op, Op)
```
fritzo: Changes in this file were pair coded with @eb8680 .

fritzo:
> can track promoting Cat to a first-class funsor in #162

Sure. Note I'm tracking filtering-related work in #177, which has some overlap with #162.

Issue #177, opened 2019-07-26
Project: Implement time series models usable in Pyro


fritzo:
This issue plans PRs required to implement Bayesian filtering models in funsor that can then be used within Pyro models. We'd like to include a number of these in the initial release.

## Low level algorithms and math
- [x] #161  Implement `sequential_sum_product()` for discrete HMMs
- [x] #176 Implement `sequential_sum_product()` for Gaussian HMMs
- [x] #183 Support multiple variables in `sequential_sum_product()`
- [x] #173 Support multivariate linear pattern matching for linear-Gaussian filtering
- [x] #178 Implement partial substitution of Gaussian funsors 
- [x] #186 #184 Support rank-deficient operands in Gaussians ops
- ~~Implement a "Gaussian GEMM" that avoids the broadcast-multiply-reduce pattern (see `pyro.ops.gaussian.gaussian_tensordot()` https://github.com/pyro-ppl/pyro/pull/1980)~~ moved to https://github.com/pyro-ppl/funsor/issues/238
- ~~#167 Implement a `markov_sum_product()` for structured HMMs with arbitrary time window~~

## High level components
- [x] #170 Implement a `FunsorDistribution` wrapper to embed funsor distributions in Pyro
- [x] #174 #175 Implement funsor-pyro conversion functions
- [x] #205 Promote `funsor.sum_product.Slice` from an ad-hoc function to a `Funsor`
- [x] #206 Implement `eager_cat` for `Tensor`, `Gaussian`, `Joint` (cross listed in #162)
- ~~#70 Add a `Markov` or `Convolve` funsor~~

## Models as distributions
- [x] #171 Implement `DiscreteHMM` as a `FunsorDistribution`
- [x] #179 Implement a `GaussianMRF`
- [x] #185 Implement a `GaussianHMM`
- ~~#182 Implement a `GaussianDiscreteMRF`~~
- ~~#180 Implement a switching linear dynamical system MRF~~
- [x]  #187 Implement a switching linear dynamical system HMM
- ~~Implement a Tensor Monte Carlo smoothing `FunsorDistribution`~~

PR #178, opened 2019-07-30
Implement partial substitution of Gaussian funsors

fritzo:
Addresses #177 
Blocking #179

This implements `Gaussian.eager_subs()` for strict subsets of real inputs, analogous to [pyro.ops.gaussian.Gaussian.condition()](https://github.com/pyro-ppl/pyro/blob/3619772/pyro/ops/gaussian.py#L125). Differences include: (1) funsor is more complex due to arbitrary substitution shape combinations, and (2) funsor uses loc=mean rather than the info_vec parametrization.

This op is required for e.g. conditioning a joint (hidden,observed) Gaussian factor on observed data, as will be used in the `GaussianMRF` and possibly `GaussianHMM`.

## Tested
- exercised by existing tests (removed an `xfail_if_not_implemented`)

### DIFF
--- a/funsor/gaussian.py
+++ b/funsor/gaussian.py
@@ -445,7 +445,7 @@ class Gaussian(Funsor, metaclass=GaussianMeta):
                                 .squeeze(-1))
         loc = loc_a - update
         log_scale = 0.5 * (_vv(update, prec_ab_diff_b) - _vmv(prec_bb, diff_b))
-        precision = prec_aa
+        precision = prec_aa.expand(loc.shape + (-1,))
         inputs = int_inputs.copy()
         for k, d in self.inputs.items():
             if k not in real_subs:
--- a/funsor/gaussian.py
+++ b/funsor/gaussian.py
@@ -422,7 +422,7 @@ class Gaussian(Funsor, metaclass=GaussianMeta):
 
         # Perform a partial substution of a subset of real variables, resulting in a Joint.
         # We split real inputs into two sets: a for the preserved and b for the substituted.
-        raise NotImplementedError('TODO')
+        # raise NotImplementedError('TODO')
         b = frozenset(k for k, v in real_subs.items())
         a = frozenset(k for k, d in self.inputs.items() if d.dtype == 'real' and k not in b)
         loc_a = torch.cat([loc[..., i] for k, i in slices if k in a], dim=-1)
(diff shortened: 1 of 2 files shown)


funsor/gaussian.py
```diff
@@ -380,42 +382,74 @@ def eager_subs(self, subs):
...
-            result = Tensor(result, inputs)
+            result = Tensor(result, int_inputs)
             assert result.output == reals()
             return Subs(result, lazy_subs)
 
         # Perform a partial substution of a subset of real variables, resulting in a Joint.
-        # See "The Matrix Cookbook" (November 15, 2012) ss. 8.1.3 eq. 353.
-        # http://www.math.uwaterloo.ca/~hwolkowi/matrixcookbook.pdf
-        raise NotImplementedError('TODO implement partial substitution of real variables')
+        # We split real inputs into two sets: a for the preserved and b for the substituted.
```
fritzo: @fehiepsi if get a chance, could you check my math here? Tests are failing, but the failure could be in either linear algebra or dimension bookkeeping.

funsor/gaussian.py
```diff
@@ -380,42 +382,74 @@ def eager_subs(self, subs):
...
+        prec_ab = torch.cat([torch.cat([
+            precision[..., i1, i2]
+            for k2, i2 in slices if k2 in b], dim=-1)
+            for k1, i1 in slices if k1 in a], dim=-2)
+        prec_bb = torch.cat([torch.cat([
+            precision[..., i1, i2]
+            for k2, i2 in slices if k2 in b], dim=-1)
+            for k1, i1 in slices if k1 in b], dim=-2)
+        prec_ab_diff_b = _mv(prec_ab, diff_b)
+        loc = loc_a + (prec_ab_diff_b.unsqueeze(-1)
```
fehiepsi: I think here the sign should be `loc_a - ...`

funsor/gaussian.py
```diff
@@ -380,42 +382,74 @@ def eager_subs(self, subs):
...
+                                     .cholesky_solve(prec_aa.cholesky())
+                                     .squeeze(-1))
+        log_scale = _vv(loc_a, prec_ab_diff_b) - 0.5 * _vmv(prec_bb, diff_b)
```
fehiepsi: I think the first term `_vv(loc_a, prec_ab_diff_b)` -> `0.5 * _vv(loc - loc_a, prec_ab_diff_b)`, where `loc - loc_a` is obtained from the above computation.

fritzo:
Thanks for the math help @fehiepsi!

PR #180, opened 2019-07-30
Implement a Switching Linear Dynamical System as an MRF

fritzo:
Addresses #177 
pair coded with @martinjankowiak 

## Tested
- [ ] unit tests


fritzo:
Closing in favor of #187

Issue #89, opened 2019-03-21
Implement a linearize interpretation


fritzo:
Blocked by #72 

One approach to implementing Extended Kalman Filter-like updates is to linearize transformations. A motivating example is sensor fusion #85 .

PR #182, opened 2019-08-01
Implement funsor.pyro.GaussianDiscreteMRF

fritzo:
Based on #181 

This adds a FunsorDistribution with latent Gaussian state and discrete observations. It uses the moment matching approximation to compute an approximate log normalizer. This also adds a `matrix_and_mvn_to_funsor()` helper similar to `pyro.ops.gaussian.matrix_and_mvn_to_gaussian()` in https://github.com/pyro-ppl/pyro/pull/1984.

## Questions for reviewers

1. Any ideas why the log normalizer is so far off? I would have expected moment matching to be more accurate, but test threshold is 4.0 (in log space)
2. Any ideas for how to test `.log_prob()` for this model? I guess I could use SVI...

## Tested
- [x] unit tests for `matrix_and_mvn_to_funsor()`
- [x] shape tests for `GaussianDiscreteMRF`
- [x] normalization tests for `GaussianDiscreteMRF` (these are very inaccurate)
- [ ] test for `GaussianDiscreteMRF.log_prob()`

### DIFF
--- a/funsor/pyro/hmm.py
+++ b/funsor/pyro/hmm.py
@@ -2,7 +2,6 @@ from collections import OrderedDict
 
 import torch
 from pyro.distributions.util import broadcast_shape
-from torch.distributions import constraints
 
 import funsor.ops as ops
 from funsor.domains import reals
@@ -62,6 +61,8 @@ class DiscreteHMM(FunsorDistribution):
 
     # TODO remove this once self.funsor_dist is defined.
     def log_prob(self, value):
+        if self._validate_args:
+            self._validate_sample(value)
         ndims = max(len(self.batch_shape), value.dim() - self.event_dim)
         value = tensor_to_funsor(value, ("time",), event_output=self.event_dim - 1,
                                  dtype=self.dtype)
@@ -131,6 +132,8 @@ class GaussianMRF(FunsorDistribution):
 
     # TODO remove this once self.funsor_dist is defined.
     def log_prob(self, value):
+        if self._validate_args:
+            self._validate_sample(value)
         ndims = max(len(self.batch_shape), value.dim() - 2)
         value = tensor_to_funsor(value, ("time",), 1)
 
@@ -236,10 +239,6 @@ class GaussianDiscreteMRF(FunsorDistribution):
             self._obs = obs
 
         super(GaussianDiscreteMRF, self).__init__(funsor_dist, batch_shape, event_shape, dtype)
-
-    @constraints.dependent_property
-    def support(self):
-        return constraints.integer_interval(0, self.dtype - 1)
 
     # TODO remove this once self.funsor_dist is defined.
     def log_prob(self, value):
--- a/funsor/pyro/hmm.py
+++ b/funsor/pyro/hmm.py
@@ -2,6 +2,7 @@ from collections import OrderedDict
 
 import torch
 from pyro.distributions.util import broadcast_shape
+from torch.distributions import constraints
 
 import funsor.ops as ops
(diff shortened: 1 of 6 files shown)


test/pyro/test_hmm.py
```diff
@@ -144,3 +146,58 @@ def test_gaussian_mrf_log_prob(init_shape, trans_shape, obs_shape, hidden_dim, o
...
+    obs_dist = random_mvn(batch_shape + (num_steps, obs_dim), hidden_dim)
+    actual_dist = GaussianDiscreteMRF(init_dist, trans_matrix, trans_dist, obs_logits, obs_dist)
+
+    data = torch.tensor(list(itertools.product(*[list(range(obs_dim))] * num_steps)))
+    assert data.shape == (obs_dim ** num_steps, num_steps)
+    data = data.reshape(data.shape[:1] + (1,) * len(batch_shape) + data.shape[1:])
+    log_prob = actual_dist.log_prob(data)
+    assert log_prob.shape[1:] == batch_shape
+    log_total = log_prob.logsumexp(0)
+    assert log_total.abs().max() < 4, log_total
```
fritzo: `exp(4.0)` seems like a big error factor 🤔

martinjankowiak: what happens if you make the moment matching more expensive, i.e. keep more components?

funsor/pyro/hmm.py
```diff
@@ -156,3 +167,114 @@ def expand(self, batch_shape, _instance=None):
...
+
+    :param ~torch.distributions.MultivariateNormal initial_dist: Represents
+        ``p(state[0])``.
+    :param ~torch.Tensor transition_matrix: Transforms ``state[t]`` to
+        ``prediction[t+1]``.
+    :param ~torch.distributions.MultivariateNormal transition_dist: Represents
+        ``p(state[t+1] | prediction[t+1])``.
+    :param ~torch.Tensor observation_logits: Represents ``p(obs[t+1])``.
+    :param ~torch.distributions.MultivariateNormal observation_dist: Represents
+        ``p(state[t+1] | obs[t+1])``.
```
fehiepsi: @fritzo I am trying to understand this model. Could you confirm that the docs above is correct?

fritzo: The docstrings are correct (except for minor off-by-1 time shifts; I'll fix those shortly).

I agree this is kind of a weird model: it is not directed and not normalized. However as a CRF/MRF it is a well-posed non-normalized density and admits tractable approximate inference via moment matching. To normalize we use the same trick as with the `GaussianMRF` in Pyro.

fritzo: I'm also open to other models that admit approximate inference, in case you can think of any. Eventually we'd like to implement a `linearize` interpretation #89 which would enable Extended Kalman Filter -like inference, e.g. linear-gaussian hidden dynamics and a probit or logistic observation model. I've started with the CRF/MRF in this PR because we already have already implemented the `moment_matching` interpretation #115 .

test/pyro/test_hmm.py
fritzo: Hmm, I'm not sure how to code that up. Another possible explanation is that the moment-matching approximation might be bad for random models (as in the test), but might be good for more plausible identifiable models. To test this I guess I could use plausible parameters above rather than random.

funsor/pyro/hmm.py
fehiepsi: Thanks! I think I start understanding now. :D

PR #187, opened 2019-08-06
Implement funsor.pyro.SwitchingLinearHMM

fritzo:
Addresses #177 

This implements a Switching Linear Dynamical System `FunsorDistribution` using the `moment_matching` interpretation and parallel scan algorithm.

This PR also:
- fixes `.expand()` method of other `FunsorDistribution`s
- adds a check to `GaussianHMM` that `obs_dim >= hidden_dim // 2`
- reorders reduction order in `GaussianHMM` to be more numerically stable

## Tested
- [x] test `SwitchingLinearHMM` shapes
- [x] test `SwitchingLinearHMM.log_prob()` against `GaussianHMM`
- [x] unit test for `.expand()` method

### DIFF
--- a/funsor/pyro/hmm.py
+++ b/funsor/pyro/hmm.py
@@ -6,7 +6,13 @@ from pyro.distributions.util import broadcast_shape
 import funsor.ops as ops
 from funsor.domains import reals
 from funsor.interpreter import interpretation
-from funsor.pyro.convert import dist_to_funsor, funsor_to_tensor, mvn_to_funsor, tensor_to_funsor
+from funsor.pyro.convert import (
+    dist_to_funsor,
+    funsor_to_tensor,
+    matrix_and_mvn_to_funsor,
+    mvn_to_funsor,
+    tensor_to_funsor
+)
 from funsor.pyro.distribution import FunsorDistribution
 from funsor.sum_product import sequential_sum_product
 from funsor.terms import Variable, lazy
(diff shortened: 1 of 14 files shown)


funsor/sum_product.py
```diff
@@ -226,6 +226,8 @@ def sequential_sum_product(sum_op, prod_op, trans, time, prev, curr):
     assert isinstance(prev, tuple) and all(isinstance(n, str) for n in prev)
     assert isinstance(curr, tuple) and all(isinstance(n, str) for n in curr)
     assert len(prev) == len(curr)
+    if time not in trans.inputs:
```
fritzo: This fixes an edge case bug in `sequential_sum_product`

fritzo: Moved to #189

funsor/pyro/hmm.py
```diff
@@ -300,3 +300,135 @@ def expand(self, batch_shape, _instance=None):
...
+
+        result = funsor_to_tensor(result, ndims=ndims)
+        return result
+
+    # TODO remove this once self.funsor_dist is defined.
+    def _sample_delta(self, sample_shape):
+        raise NotImplementedError("TODO")
+
+    def expand(self, batch_shape, _instance=None):
+        new = self._get_checked_instance(DiscreteHMM, _instance)
```
martinjankowiak: `DiscreteHMM`?

test/pyro/test_hmm.py
```diff
@@ -185,3 +185,89 @@ def test_gaussian_mrf_log_prob(init_shape, trans_shape, obs_shape, hidden_dim, o
...
+
+    data = obs_mvn.expand(shape).sample()[..., 0, :]
+    actual_log_prob = actual_dist.log_prob(data)
+    assert actual_log_prob.shape == expected_batch_shape
+
+
+@pytest.mark.parametrize("num_components", [2, 3])
+@pytest.mark.parametrize("obs_dim,hidden_dim", [(1, 1), (2, 1), (2, 2)])
+@pytest.mark.parametrize("num_steps", [1, 2, 3, 4, 5, 6])
+def test_switching_linear_hmm_log_prob(num_steps, hidden_dim, obs_dim, num_components):
```
martinjankowiak: 💯

funsor/pyro/hmm.py
fritzo: Good catch. I've added a test for this, then fixed a bunch of similar bugs.

fritzo:
Thanks for your first review, @martinjankowiak. Since then I've added a bunch of tests and fixes for `.expand()` method.

funsor/pyro/hmm.py
```diff
@@ -207,8 +197,8 @@ def log_prob(self, value):
         result = self._trans + obs
         result = sequential_sum_product(ops.logaddexp, ops.add,
                                         result, "time", {"state": "state(time=1)"})
-        result = self._init + result.reduce(ops.logaddexp, "state(time=1)")
-        result = result.reduce(ops.logaddexp, "state")
+        result += self._init
```
fritzo: @martinjankowiak @fehiepsi This reorders so that `.reduce()` is called only after adding in the initial state prior, which improves conditioning of Cholesky decompositions inside `.reduce()`. This reordering fixed lots of actual errors in the `SwitchingLinearHMM`, and in the `GaussianHMM` allowed me to reduce test tolerance from 1e-4 to 1e-5.

PR #185, opened 2019-08-04
Implement funsor.pyro.GaussianHMM

fritzo:
Addresses #177 
~~Blocked by #186~~

This implements a `GaussianHMM` distribution using a parallel scan algorithm. This is nearly identical to https://github.com/pyro-ppl/pyro/pull/1984, including the helper `matrix_and_mvn_to_funsor()`.

## Tested
- tested against `pyro.distributions.GaussianHMM`
- unit test for `matrix_and_mvn_to_funsor()`

PR #188, opened 2019-08-07
Fix normalization bug in moment_matching interpretation

fritzo:
Blocking #187 

This fixes a normalization bug in our `moment_matching` interpretation.

Our original tests only checked mean and covariance, but since funsor distributions are non-normalized, we should also check total. This PR adds such tests and fixes the error.

I've also switched from `sym_inverse()` to Cholesky-based inverse in `Joint.moment_matching_reduce()` as suggested by @fehiepsi in #185 .

## Tested
- added total checks to existing tests
- added a monte carlo test for zeroth, first, and second moments

### DIFF
--- a/examples/slds.py
+++ b/examples/slds.py
@@ -72,7 +72,7 @@ if __name__ == '__main__':
     parser = argparse.ArgumentParser(description="Switching linear dynamical system")
     parser.add_argument("-t", "--time-steps", default=10, type=int)
     parser.add_argument("-n", "--train-steps", default=101, type=int)
-    parser.add_argument("-lr", "--learning-rate", default=0.05, type=float)
+    parser.add_argument("-lr", "--learning-rate", default=0.01, type=float)
     parser.add_argument("--filter", action='store_true')
     parser.add_argument("-v", "--verbose", action="store_true")
     args = parser.parse_args()
--- a/funsor/joint.py
+++ b/funsor/joint.py
@@ -1,5 +1,4 @@
 import functools
-import math
 from collections import OrderedDict
 from functools import reduce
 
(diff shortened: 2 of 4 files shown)


examples/slds.py
```diff
@@ -72,7 +72,7 @@ def model(data):
     parser = argparse.ArgumentParser(description="Switching linear dynamical system")
     parser.add_argument("-t", "--time-steps", default=10, type=int)
     parser.add_argument("-n", "--train-steps", default=101, type=int)
-    parser.add_argument("-lr", "--learning-rate", default=0.05, type=float)
+    parser.add_argument("-lr", "--learning-rate", default=0.01, type=float)
```
fritzo: I was seeing NaNs; decreasing learning rate fixed those NaNs.

fritzo:
@martinjankowiak thanks for reviewing!

funsor/joint.py
```diff
@@ -155,31 +154,28 @@ def moment_matching_reduce(self, op, reduced_vars):
...
 
-            gaussian = self.gaussian
             int_inputs = OrderedDict((k, d) for k, d in gaussian.inputs.items() if d.dtype != 'real')
             probs = (discrete - new_discrete).exp()
             old_loc = Tensor(cholesky_solve(gaussian.info_vec.unsqueeze(-1),
                                             gaussian._precision_chol).squeeze(-1),
                              int_inputs)
             new_loc = (probs * old_loc).reduce(ops.add, approx_vars)
-            old_cov = Tensor(sym_inverse(gaussian.precision), int_inputs)
+            old_cov = Tensor(cholesky_inverse(gaussian._precision_chol), int_inputs)
```
fehiepsi: @fritzo I think that you can compute `old_cov` first, then compute `old_loc` based on `old_cov`. For `new_precision`, I can't find a better identity to compute it. For `new_cov`, can we add `old_cov` and `outers` before applying `reduce(ops.add...)`? I don't know enough funsor to justify it.

fritzo: > you can compute old_cov first, then compute old_loc based on old_cov

`cholesky_solve` is more numerically stable than `cholesky_inverse(...).matmul()` and has same asymptotic complexity.

> can we add `old_cov` and outers before applying `reduce(ops.add...)`

It is cheaper to reduce first and add smaller things (even though it adds a tensor op).

PR #189, opened 2019-08-07
Fix sequential_sum_product, add naive_sequential_sum_product

fritzo:
Addresses #177 
Blocking #187 

This PR:
1. Fixes an edge case bug in `sequential_sum_product` that happens with length-1 time series.
2. Simplifies the interface of `sequential_sum_product` to input a dict rather than two tuples
3. Adds a reference implementation `naive_sequential_sum_product()` for use in testing. This version does not rely on `Cat` and `Stack`, and can be used to test e.g. an exact version of `SwitchingLinearHMM` #187 that avoids a moment matching approximation.

## Tested
- added new test cases for `num_steps=1`
- enabled previously xfailing hmm test cases with `num_steps=1`
- added new test cases for `naive_sequential_sum_product()`

### DIFF
--- a/funsor/sum_product.py
+++ b/funsor/sum_product.py
@@ -193,15 +193,41 @@ def Slice(name, *args):
     return Tensor(data, inputs, dtype=bound)
 
 
-def sequential_sum_product(sum_op, prod_op, trans, time, prev, curr):
+def naive_sequential_sum_product(sum_op, prod_op, trans, time, step):
+    assert isinstance(sum_op, AssociativeOp)
+    assert isinstance(prod_op, AssociativeOp)
+    assert isinstance(trans, Funsor)
+    assert isinstance(time, str)
+    assert isinstance(step, dict)
+    assert all(isinstance(k, str) for k in step.keys())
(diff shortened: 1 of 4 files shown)


funsor/sum_product.py
```diff
@@ -193,15 +193,41 @@ def Slice(name, *args):
...
+    For a funsor ``trans`` with dimensions ``time``, ``step.keys()``,
+    and ``step.values()``, computes a recursion equivalent to::
 
         tail_time = 1 + arange("time", trans.inputs["time"].size - 1)
         tail = sequential_sum_product(sum_op, prod_op,
                                       trans(time=tail_time),
-                                      "time", "prev", "curr")
+                                      "time", step)
         return prod_op(trans(time=0)(curr="drop"), tail(prev="drop")) \
            .reduce(sum_op, "drop")
```
eb8680: nit: update these docstrings as well

fritzo: I've updated while intentionally restricting to a single variable case, since the multivariable case seems too complex for a docstring.

PR #195, opened 2019-08-13
add failing gaussian hmm test

martinjankowiak:
is the test correct?


martinjankowiak:
ok this test is now correct and passes

Issue #124, opened 2019-04-09
Add a forward-filter-backward-sample example


fritzo:
This example should demonstrate how to use `funsor.adjoint` to perform exact inference on a foward pass, followed by monte carlo on the backward pass. A discrete HMM example would be nice.

fritzo:
@eb8680 is this already completed? IIRC you did this as part of our paper submission?

eb8680:
> is this already completed?

I don't have a backward-sampling example yet, I guess I could add that to the seal example.  Let's leave this open for now.

PR #197, opened 2019-08-16
Update adjoint to support different forward and backward interpretations

### LINKED ISSUE #164: Update adjoint to be compatible with alpha-conversion
`funsor.adjoint` was broken by the introduction of alpha-conversion in #148.  In particular, when the target terms passed to `adjoint(expr, targets)`, which are subexpressions of `expr`, contain variables that are bound in `expr` (e.g. by a `Reduce`), the variables are renamed and `adjoint` cannot find the target terms.  One simple fix would be to introduce an "unbind" operation that reverses alpha-renaming and apply it as part of `adjoint`.

eb8680:
Addresses #124.  Resolves #164.

This PR updates `funsor.adjoint` to be compatible with alpha-renaming and adds support for separate forward and backward interpretations, enabling use cases like Viterbi decoding and forward filtering/backward sampling algorithms.

The interface has changed somewhat - now the user explicitly creates a tape and records the forward computation, then separately calls `adjoint` to run the backward computation:
```py
with interpretation(some_interpretation), AdjointTape() as tape:
    y = do_some_stuff(targets)

with interpretation(another_interpretation):
    adjoints = tape.adjoint(ops.logaddexp, ops.add, y, targets)
```

### DIFF
--- a/funsor/adjoint.py
+++ b/funsor/adjoint.py
@@ -2,49 +2,83 @@ from collections import defaultdict
 
 import torch
 
+import funsor.interpreter as interpreter
 import funsor.ops as ops
 from funsor.contract import Contract
-from funsor.interpreter import interpretation, reinterpret
+from funsor.interpreter import interpretation
 from funsor.ops import AssociativeOp
 from funsor.registry import KeyedRegistry
-from funsor.terms import Binary, Funsor, Number, Reduce, Variable, eager
+from funsor.terms import Binary, Funsor, Reduce, Variable, alpha_substitute, lazy, to_funsor
 from funsor.torch import Tensor
+
+
+def _alpha_deconvert(expr):
+    alpha_subs = {name: name.split("__BOUND")[0]
+                  for name in expr.bound if "__BOUND" in name}
+    if not alpha_subs:
+        return tuple(expr._ast_values)
+
+    return alpha_substitute(expr, alpha_subs)
 
 
 class AdjointTape(object):
 
     def __init__(self):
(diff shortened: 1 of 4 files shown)


fritzo:
Feel free to add argmax test cases in a follow-up PR.

funsor/adjoint.py
```diff
@@ -2,49 +2,82 @@
...
 
-    while tape_recorder.tape:
-        output, fn, inputs = tape_recorder.tape.pop()
-        in_adjs = adjoint_ops(fn, adjoint_values[output], output, *inputs)
-        for v, adjv in in_adjs.items():
-            multiplicities[v] += 1
-            adjoint_values[v] = adjoint_values[v] + adjv  # product in logspace
+        target_adjs = {}
+        for v in targets:
+            target_adjs[v] = adjoint_values[v] / multiplicities[v]  # TODO use correct op here with bin_op
```
fritzo: This seems fishy. I'd not expect a `.mean()` but rather a `.sum()` here. Maybe this is caused by the `modulo_total=True` option in Pyro, in which case you could change the test to also compute and divide by the total `Z`.

fritzo:
(from in-person review:)
It would be good to add algebraic tests in a follow-up PR, roughly:
- adjoint(x + x) = adjoint(x) + adjoint(x)
- adjoint(x + y) = adjoint(x) + adjoint(y)
- adjoint(c * x) = c * adjoint(x)
- adjoint(adjoint(x)) == x

PR #198, opened 2019-08-19
Add additional einsum backend support

eb8680:
This PR adds some support for a couple other backends to `funsor.einsum`, includes `torch_map` in `test_einsum.py` and `test_optimizer.py`, and adds some xfailing adjoint tests for the `torch_map` backend.  To make these additional test cases less annoying, I also fixed a longstanding bug that made einsum test generation nondeterministic, which was preventing us from using `pytest-xdist` to parallelize test execution, and updated `make test` to use `pytest-xdist`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,12 +18,12 @@ setup(
         'unification',
     ],
     extras_require={
-        'test': ['flake8', 'pytest>=4.1', 'pytest-xdist>=1.27.0', 'torchvision==0.2.1'],
+        'test': ['flake8', 'pytest>=4.1', 'pytest-xdist==1.27.0', 'torchvision==0.2.1'],
         'dev': [
             'flake8',
             'isort',
             'pytest>=4.1',
-            'pytest-xdist>=1.27.0',
+            'pytest-xdist==1.27.0',
             'sphinx>=2.0',
             'sphinx_rtd_theme',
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -18,11 +18,12 @@
         'unification',
     ],
     extras_require={
-        'test': ['flake8', 'pytest>=4.1', 'torchvision==0.2.1'],
+        'test': ['flake8', 'pytest>=4.1', 'pytest-xdist==1.27.0', 'torchvision==0.2.1'],
         'dev': [
             'flake8',
             'isort',
             'pytest>=4.1',
+            'pytest-xdist==1.27.0',
```
fritzo: why not `>=`?

eb8680: Travis was complaining with `pytest>=4.1` and `pytest-xdist>=1.27.0`, so I had to either use `pytest-xdist==1.27.0` or `pytest>=4.4`.

eb8680:
> What do you think is causing the test failures?

Some of the algebra may be wrong, and also `torch_map` computes the argmax rather than max on the backward pass, so it doesn't do the same thing as `ops.max`/`ops.add` at the moment.

eb8680:
Going to merge this since it doesn't make anything worse and fixes the `pytest-xdist` bug.

PR #200, opened 2019-08-23
fix broken docs

jpchen:
Addresses #201 

[voila](https://funsor.readthedocs.io/en/docs/)


fritzo:
Thanks for fixing this @jpchen !

Issue #203, opened 2019-08-27
Approximate funsor variable elimination


fritzo:
This issue proposes a way to address intractability due to multiple plates or time axes by
1. creating an approximate interpretation that proposes multiple samples via Sarkka along one axis, then
2. applying tensor monte carlo on all remaining axes, again via Sarkka on all axes.
3. Further, a meta optimizer might decide which axis should come first, e.g. based on mutual information.

PR #204, opened 2019-08-28
Support Markov dimensions in sum_product()

fritzo:
This aims to combine Tensor Variable Elimination with parallel-scan Sarkka-style reduction over markov dimensions. (This does not address intractability due to multiple time dimensions; we may address that in a follow-up PR #203)

## Tasks

- [x] use `sequential_sum_product()` inside `partial_sum_product()`
- [x] refactor types of `sum_product()`
- [x] expose generalized interface in `einsum`

## Tested
- [x] refactoring is exercised by existing tests
- [ ] test edge case of `sequential_sum_product()`
- [ ] test with a single time dimension (currently fails)
- [ ] test with two time dimensions
- [ ] test generalized `einsum` interface


fritzo:
Closing as out of scope for now.

PR #205, opened 2019-08-29
Promote Slice to a funsor

fritzo:
Addresses #177 

This adds a `Slice` funsor for use in `sequential_sum_product()`. Before this PR, `Slice` was simulated as `arange` and thus triggered a tensor copy. After this PR slicing is zero-copy.

## Tested
- unit tests
- refactoring is covered by `sequential_sum_product()` tests


fritzo:
I've implemented `Slice` for `Stack`, but I don't want to slow us down maintaining `Array` support; the entire funsor.numpy needs to be rewritten from scratch and I'd rather delete it than maintain it.

PR #206, opened 2019-08-29
Implement Cat funsor

fritzo:
Addresses #162, #177 

This promotes `Cat` to a funsor, implemented using variadic pattern matching following #157 so that homogeneous Stack/Tensor/Gaussian/Joint funsors can be concatenated eagerly.

This also reorders `Stack(components,name)` -> `Stack(name,parts)` to agree with `Cat` and make it a little cleaner for both funsors to use variadic pattern matching.

This does not implement every possible `Cat` pattern; rather this PR aims to implement only what is needed for `sequential_sum_product()`.

## Tested
- unit tests for `Cat` and eager Tensor fusion
- refactoring is exercised by `sequential_sum_product()` tests

### DIFF
--- a/funsor/joint.py
+++ b/funsor/joint.py
@@ -3,13 +3,15 @@ from collections import OrderedDict
 from functools import reduce
 
 import torch
+from multipledispatch import dispatch
+from multipledispatch.variadic import Variadic
 
 import funsor.interpreter as interpreter
 import funsor.ops as ops
 import funsor.terms
 from funsor.delta import Delta
-from funsor.domains import reals
-from funsor.gaussian import Gaussian, cholesky_inverse, cholesky_solve
+from funsor.domains import bint, reals
+from funsor.gaussian import Gaussian, align_gaussian, cholesky_inverse, cholesky_solve
 from funsor.integrate import Integrate, integrator
 from funsor.montecarlo import monte_carlo
 from funsor.ops import AddOp, NegOp, SubOp
@@ -27,7 +29,7 @@ from funsor.terms import (
     eager,
     to_funsor
 )
-from funsor.torch import Tensor, arange
+from funsor.torch import Tensor, align_tensor, arange
 
 
 class JointMeta(FunsorMeta):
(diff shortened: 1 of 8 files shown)


funsor/joint.py
```diff
@@ -260,6 +262,55 @@ def eager_independent(joint, reals_var, bint_var):
     return None  # defer to default implementation
 
 
+@dispatch(str, Variadic[Gaussian, Joint])
```
fritzo: This code was moved from the previous `funsor.sum_product.Cat` function.

funsor/torch.py
```diff
@@ -509,6 +510,28 @@ def eager_contract(sum_op, prod_op, lhs, rhs, reduced_vars):
     return Tensor(data, inputs, dtype)
 
 
+@dispatch(str, Variadic[Tensor])
```
fritzo: This code was moved from the previous `funsor.sum_product.Cat` function.

fritzo:
bump

funsor/terms.py
```diff
@@ -1057,53 +1058,95 @@ class Stack(Funsor):
...
+
+@eager.register(Cat, str, tuple)
+def eager_cat(name, parts):
+    if len(parts) == 1:
+        return parts[0]
+    return eager_cat_homogeneous(name, *parts)
+
+
+@dispatch(str, Variadic[Funsor])
+def eager_cat_homogeneous(name, *parts):
```
eb8680: Why use this pattern instead of `eager.register(Cat, str, Variadic[Funsor])`?

funsor/terms.py
```diff
@@ -1057,53 +1058,95 @@ class Stack(Funsor):
...
+    return None  # defer to default implementation
+
+
+@dispatch(str, Variadic[Stack])
+def eager_cat_homogeneous(name, *parts):
+    parts = sum((p.parts for p in parts), ())
+    return Stack(name, parts)
```
eb8680: This behavior isn't consistent with `Tensor` if `any(part.name != name for part in parts)` - maybe raise an error in that case?

fritzo: Good catch! I'll fix shortly...

funsor/terms.py
fritzo: This is a weak aesthetic preference. It feels like we're compromising by using `multipledispatch` (thank you matt!) whereas funsor really deserves nested pattern matching. So I've made a nice tidy `Cat` interface with a nested datatype, and added this shim to deal with a library that can't handle nested patterns. Hopefully funsor will someday be reimplemented in a cleaner language than Python and this nested `Cat` datatype will survive.

But I'm happy to refactor to a flattened `Cat` signature if you think it's cleaner to hard code to current limitations of `multipledispatch`. If you are using the flattened pattern elsewhere in funsor, I guess we can be consistent and use that pattern here.

funsor/terms.py
fritzo: Fixed by deferring to default (lazy) interpretation.

fritzo:
Thanks @eb8680 for reviewing!

PR #208, opened 2019-09-02
Add ClippedAdam to funsor.minipyro

fritzo:
This adds Martin's `ClippedAdam` optimizer to funsor.minipyro. I've been using this for the [BART example](https://github.com/pyro-ppl/sandbox/blob/master/2019-08-time-series/bart/forecast.py) and I'd like to port this example to funsor.

## Tested
- added a unit test

### DIFF
--- a/funsor/minipyro.py
+++ b/funsor/minipyro.py
@@ -340,7 +340,7 @@ class PyroOptim(object):
             optim.step()
 
 
-# We wrap the two most commonly used PyTorch optimizers.
+# We wrap some commonly used PyTorch optimizers.
 class Adam(PyroOptim):
     TorchOptimizer = torch.optim.Adam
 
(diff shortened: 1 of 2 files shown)


funsor/minipyro.py
```diff
@@ -333,12 +334,21 @@ def __call__(self, params):
...
             # an Adam optimizer and keep track of it.
             else:
-                optim = torch.optim.Adam([param], **self.optim_args)
+                optim = self.TorchOptimizer([param], **self.optim_args)
                 self.optim_objs[param] = optim
             # Take a gradient step for the parameter param.
             optim.step()
 
 
+# We wrap the two most commonly used PyTorch optimizers.
```
martinjankowiak: minus "most"?

PR #209, opened 2019-09-02
Add a Poisson distribution

fritzo:
This adds a `Poisson` distribution following the pattern of the `BernoulliProbs` distribution. This is required for a funsor implementation of the BART forecasting example.

Note that in Funsor we encode Poisson samples as `reals()` since they don't fit into the `bint` type; this is a slight type mismatch, but it would be a huge effort to add a `nat` datatype.

## Tested
- added a unit test following other examples

### DIFF
--- a/funsor/distributions.py
+++ b/funsor/distributions.py
@@ -463,6 +463,25 @@ def eager_mvn(loc, scale_tril, value):
     return Tensor(log_prob, int_inputs) + Gaussian(info_vec, precision, inputs)
 
 
+class Poisson(Distribution):
+    dist_class = dist.Poisson
+
+    @staticmethod
+    def _fill_defaults(rate, value='value'):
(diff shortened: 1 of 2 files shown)


fritzo:
> @martinjankowiak what's the purpose of the `._fill_defaults()` methods?

The `._fill_defaults()` method makes default value creation compatible with cons hashing (i.e. memoizing funsor `.__init__()` methods). Usually that kind of logic would live directly in the `.__init__()` method of each class, but in funsor we need to fill in defaults before calling `cls.__call__()`, which is where the cons hashing comparisons are done. For example we want the following to hold:
```py
assert Normal(0, 1) is Normal(0, 1, value="value")
```
and `DistributionMeta.__call__()` can fill in `value="value"` on the first so that their `.__init__()` methods see the same inputs.

PR #212, opened 2019-09-03
Add parametric types for more convenient pattern matching

eb8680:
This PR adds parametric typing to Funsor terms with a syntax identical to that used in the `typing` standard library module: terms like `Unary` can be subtyped with brackets: `Unary[TransformOp, Tensor]` such that `issubclass(Unary[...], Unary) == True`.

The intended use case is in writing more specific patterns for use by `multipledispatch`, which works by using `issubclass`.  For example, we have several definitions of the form
```py
@eager.register(Binary, AssociativeOp, Unary, Unary)
def eager_binary_unary(op, lhs, rhs):
    if lhs.op is ops.exp and rhs.op is ops.exp:
        ...
    if lhs.op is ops.add and rhs.op is ops.add:
        ...
```
The new functionality in this PR allows such definitions to be split up:
```py
@eager.register(Binary, AssociativeOp, Unary[ExpOp, Funsor], Unary[ExpOp, Funsor])
def eager_binary_exps(op, lhs, rhs):
    ...

@eager.register(Binary, AssociativeOp, Unary[LogOp, Funsor], Unary[LogOp, Funsor])
def eager_binary_logs(op, lhs, rhs):
    ...
```

I also removed the `funsor.pattern` module, which we never actually used for anything and which is made obsolete by this PR.

Tested:
- Added tests that check some edge cases of subclassing
- Exercised by existing tests - does not break cons hashing or interfere with existing code


eb8680:
> Could you also add one or two example simplifications in this PR?

Sure.  In fact, assuming the performance hit isn't too bad, it would be nice to work towards removing all such cases, so that only runtime dispatching (e.g. dispatching on the values or number of reduced variables in a `frozenset`) happens inside pattern definitions.

eb8680:
> Does this support nesting?

Yes, that pattern should work fine:
```py
assert issubclass(Reduce[AddOp, Reduce[AddOp, Funsor, frozenset], frozenset], Funsor)
assert issubclass(Reduce[AddOp, Reduce[AddOp, Funsor, frozenset], frozenset], Reduce)
assert issubclass(Reduce[AddOp, Reduce[AddOp, Funsor, frozenset], frozenset], Reduce[AddOp, Funsor, frozenset])
```

> some logic will still be needed e.g. to unify variables as in the `Cat[str, tuple[Stack]]` pattern

Nesting of variadic types or types with variables is not supported here, so you could only write `Cat[str, tuple]`, not `Cat[str, VariadicTuple[Stack]]`.  Given the substantial extra complexity it would require, I'm not sure it's worth supporting right now.  I do intend to add support for tuples of fixed length though, since I'd like to use such patterns in #157 .

eb8680:
Ok, I added support for fixed-length tuples and added a couple of simple tests.

funsor/terms.py
```diff
@@ -171,6 +188,60 @@ def __call__(cls, *args, **kwargs):
...
+            assert len(arg_types) == len(cls._ast_fields), "must provide types for all params"
+            new_name = cls.__name__ + "[{}]".format(", ".join(
+                t.__name__ if hasattr(t, "__name__") else str(t) for t in arg_types))  # Tuple doesn't have __name__
+            new_dct = cls.__dict__.copy()
+            new_dct.update({"__args__": arg_types})
+            # type(cls) to handle FunsorMeta subclasses
+            cls._type_cache[arg_types] = type(cls)(new_name, (cls,), new_dct)
+        return cls._type_cache[arg_types]
+
+    def __subclasscheck__(cls, subcls):  # issubclass(subcls, cls)
```
fritzo: nit: could we simplify this, maybe something like:
```py
if cls is subcls:
    return True
if not isinstance(subcls, FunsorMeta):
    return super(FunsorMeta, getattr(cls, "__origin__", cls)).__subclasscheck__(subcls)

cls_origin = getattr(cls, "__origin__", cls)
subcls_origin = getattr(subcls, "__origin__", subcls)
if not super(FunsorMeta, cls_origin).__subclasscheck__(subcls_origin):
    return False

if cls.__args__ and subcls.__args__:
    if len(cls.__args__) != len(subcls.__args__):
        return False
    for subcls_param, param in zip(subcls.__args__, cls.__args__):
        if not _issubclass_tuple(subcls_param, param):
            return False
return True
```

funsor/terms.py
```diff
@@ -171,6 +188,60 @@ def __call__(cls, *args, **kwargs):
...
+        raise TypeError("could not unify: {}, {}".format(cls, subcls))
+
+
+def _issubclass_tuple(subcls, cls):
+    """
+    utility for pattern matching with tuple subexpressions
+    """
+    subcls_is_tuple = hasattr(cls, "__origin__") and subcls.__origin__ is tuple
+    cls_is_tuple = hasattr(cls, "__origin__") and cls.__origin__ is tuple
+    if not subcls_is_tuple and not cls_is_tuple:
```
fritzo: nit: again, I find it a bit easier to read the early-exit pattern. maybe something like
```py
if subclas_is_tuple != cls_is_tuple:
    return False
if not cls_is_tuple:
    return issubclass(subcls, cls)
if not cls.__args__:
    return True
if len(subcls.__args__) != len(cls.__args__)
    return False
for a, b in zip(subcls.__args__, cls.__args__):
    if not _issubclass_tuple(a, b):
        return False
return True
```

PR #215, opened 2019-09-03
Omit typing information from .__str__() and .pretty() methods

fritzo:
This removes extraneous information introduced in #212, which creates unreadably complex debug messages.

@eb8680 would it instead be possible to avoid munging the `.__name__` attribute?


fritzo:
Name munging causes so many issues 😢 I'm only making this PR to debug problems with name munging in alpha conversion.

fritzo:
Closed in favor of #216

PR #217, opened 2019-09-04
Support Independent(Normal) in matrix_and_mvn_to_funsor()

fritzo:
This ports Pyro's https://github.com/pyro-ppl/pyro/pull/2005 to funsor. The math and code structure are nearly identical. As in Pyro we create a temporary `AffineNormal` object to avoid expensive intermediate computations.

One difference in Funsor is that the `AffineNormal.condition()` method becomes a pattern. For testing purposes I've added another simple pattern, fixing `x` rather than `y`.

## Tested
- added unit test for the `AffineNormal` funsor
- added a unit test comparing `Independent(Normal)` vs `MultivariateNormal(..., scale.diag_embed())` (ported from Pyro)

### DIFF
--- a/docs/source/pyro.rst
+++ b/docs/source/pyro.rst
@@ -27,6 +27,5 @@ Conversion Utilities
 ====================
 .. automodule:: funsor.pyro.convert
     :members:
-    :undoc-members:
     :show-inheritance:
     :member-order: bysource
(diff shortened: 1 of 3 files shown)


fritzo:
Thanks for reviewing!

Issue #211, opened 2019-09-03
FR: syntactic sugar and API tweaks


martinjankowiak:
- [x] #218 wrap passed args in `frozenset` under the hood so i can write `log_prob.reduce(ops.logaddexp, {a, b})` instead of `log_prob.reduce(ops.logaddexp, frozenset([a, b]))` ?
- [x] #276 similarly support passing in `funsor.Variable`s into `reduce` so i can write `log_prob.reduce(ops.logaddexp, frozenset([var_a, var_b]))` instead of `log_prob.reduce(ops.logaddexp, frozenset([var_a.name, var_b.name]))` ?
- [x] #276 support passing str and set args to `Integrate`, just like `.reduce()`.

## Triaged
- some of the funsor conversion methods, e.g. `matrix_and_mvn_to_funsor`, should probably error if passed spurious `event_dims`; c.f. [slds example](https://github.com/pyro-ppl/funsor/blob/fcc9a0e90b65f3a17ea243a16acf29fddab24336/examples/eeg_slds.py#L97)
  - let's revisit this as we see more usage patterns. In the long term we may deprecate these methods in favor of Gaussian-affine pattern matching anyway. -@fritzo

PR #218, opened 2019-09-04
Support set-valued args to Funsor.reduce()

fritzo:
Addresses #211 

This extends `.reduce()` to accept `set`s of strings, since those are much easier to read and write than `frozenset([])`s. I've intentionally avoided generalizing this gratuitously, e.g. to lists, tuples, iterables etc.

## Tested
- added a unit test
- converted some examples to use this terser notation

### DIFF
--- a/examples/slds.py
+++ b/examples/slds.py
@@ -44,8 +44,7 @@ def main(args):
 
             # Marginalize out previous delayed sample statements.
             if t > 0:
-                log_prob = log_prob.reduce(
-                    ops.logaddexp, frozenset([s_prev.name, x_prev.name]))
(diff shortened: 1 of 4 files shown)


fritzo:
> are there other methods in funsor ...?

The `.reduce()` and `.__call__()` methods are special in that they are the preferred interface to `Reduce` and `Subs` funsors. I'm happy adding sugar to these wrappers but I'm more conservative in adding sugar to `Funsor` constructor args because funsor construction is already quite subtle.

One place we could possibly do this is in `funsor.Integrate()` which also takes a set of reduced variables. I think I'd prefer to wrap this in a `funsor.integrate()` function rather than make the `Integrate` funsor more magical.

PR #221, opened 2019-09-05
Add eager_or_die interpretation to diagnose missing patterns

fritzo:
This adds an interpretation like `eager` but that raises `NotImplemntedError` if it is unable to match a pattern for the most common funsor types: `Subs`, `Unary`, `Binary`, `Reduce`.

The motivation is that, in large computations missing patterns result in huge difficult-to-debug lazy expressions (e.g. @jpchen and I both hit missing-pattern bugs today and ended up with multi-page funsors). An extreme solution would be to error as soon as any pattern is missing.

## Questions

1. Is the set `{Subs, Unary, Binary, Reduce}` reasonable? In practice I find almost all missing patterns result in Binary, Reduce, or Subs.

## Tested
- unit tests
- ran on a failing test #220 and received a helpful error message 😄 
  ```
  NotImplementedError: Missing pattern for Binary(GetitemOp(0), bias, Tensor(tensor([0]), OrderedDict([('time', bint(1))]), 2))
  ```

### DIFF
--- a/funsor/terms.py
+++ b/funsor/terms.py
@@ -869,6 +869,13 @@ def eager_unary(op, arg):
     return interpreter.debug_logged(arg.eager_unary)(op)
 
 
+@eager.register(Unary, AssociativeOp, Funsor)
+def eager_unary(op, arg):
+    if not arg.output.shape:
+        return arg
+    return interpreter.debug_logged(arg.eager_unary)(op)
+
+
 _INFIX = {
     ops.add: '+',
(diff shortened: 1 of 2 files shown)


eb8680:
> Is the set {Subs, Unary, Binary, Reduce} reasonable?

Maybe add `Integrate` and `Contract`?

eb8680:
This is great, merging because I need it too!

PR #222, opened 2019-09-05
Add support for nested union types

eb8680:
This PR follows up on #212 by adding support for pattern matching with nested union types:
```py
@eager.register(Binary, AssociativeOp, (Tensor, Reduce[Op, (Reduce[Op, Funsor, frozenset], Gaussian), frozenset]), Funsor)
```
This should probably be used somewhat sparingly since it hasn't been optimized for performance.  Note that type variables and nested variadic types are still disallowed and I currently have no plans to support them.

Tested:
- Added extra test cases in `test_terms.py`

### DIFF
--- a/funsor/terms.py
+++ b/funsor/terms.py
@@ -192,8 +192,6 @@ class FunsorMeta(type):
         if not isinstance(arg_types, tuple):
             arg_types = (arg_types,)
         assert not any(isvariadic(arg_type) for arg_type in arg_types), "nested variadic types not supported"
-        # TODO add support for nested union types (i.e. remove the following assertion)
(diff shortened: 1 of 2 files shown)


eb8680:
> Can you explain the design choice of representing unions as tuples rather than Union[a,b] objects?

Either one should work with the implementation in this PR, and it seems stylistically preferable to use `Union`.  Note that as of Python 3.7, `typing` does not natively support subclass or instance checks for parametric types, so one isn't necessarily more natural than the other (I can't quite figure out why, but I guess it's something to do with `typing` being mostly intended for use by static type checkers like mypy).

PR #223, opened 2019-09-05
Ensure substitute() calls to_funsor()

fritzo:
This aims to ensure `.eager_subs()` is only called with `Funsor`s.

## Tested
- refactoring exercised by existing tests


fritzo:
This doesn't appear compatible with `Independent`.

PR #224, opened 2019-09-05
Independent rename

fritzo:
Based on #223 

This ensures `Independent` funsors are preserved under renaming, and thus can be sampled from. Prior to this PR `Independent(...)(value="foo")` would become a `Reduce(...)` and then not implement sampling.

## Tested
- added a unit test
- tested in a BART forecasting example


fritzo:
Closing due to alpha renaming issues.

PR #226, opened 2019-09-06
Make Independent into a three-parameter funsor

fritzo:
Addresses #228 
~~Blocked by #227~~

This refactors `Independent` to take three different names, as Noah suggested long ago. The intention is to fix a subtle bug involving `Independent` and `sequential_sum_product()` in the BART forecasting example.

## Tested
- ran BART example #228 to demonstrate this fixes a renaming error
  (the example not fails due to a missing pattern, which is better than an error :smile:)

### DIFF
--- a/funsor/terms.py
+++ b/funsor/terms.py
@@ -114,6 +114,24 @@ def eager(cls, *args):
 
 
 @dispatched_interpretation
+def eager_or_die(cls, *args):
+    """
+    Eagerly execute ops with known implementations.
+    Disallows lazy :class:`Subs` , :class:`Unary` , :class:`Binary` , and
+    :class:`Reduce` .
+
+    :raises: :py:class:`NotImplementedError` no pattern is found.
+    """
+    result = eager.dispatch(cls, *args)
+    if result is None:
+        if cls in (Subs, Unary, Binary, Reduce):
+            raise NotImplementedError("Missing pattern for {}({})".format(
+                cls.__name__, ", ".join(map(str, args))))
+        result = reflect(cls, *args)
+    return result
+
+
+@dispatched_interpretation
 def sequential(cls, *args):
     """
     Eagerly execute ops with known implementations; additonally execute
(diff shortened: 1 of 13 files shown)


funsor/terms.py
```diff
@@ -1316,63 +1319,71 @@ class Independent(Funsor):
...
         self.fn = fn
         self.reals_var = reals_var
         self.bint_var = bint_var
-        self.reals_var_bound = reals_var_bound
+        self.diag_var = diag_var
+
+    def _alpha_convert(self, alpha_subs):
+        fn, reals_var, bint_var, diag_var = super()._alpha_convert(alpha_subs)
+        bint_var = alpha_subs.get(bint_var, bint_var)
+        diag_var = alpha_subs.get(diag_var, diag_var)
```
fritzo: This one line is the entire purpose of #226 and #227 :smile:

funsor/terms.py
```diff
@@ -1316,63 +1327,73 @@ class Independent(Funsor):
...
         assert isinstance(bint_var, str)
         assert bint_var in fn.inputs
         assert isinstance(fn.inputs[bint_var].dtype, int)
+        assert isinstance(diag_var, str)
+        assert diag_var in fn.inputs
+        assert fn.inputs[diag_var].dtype == 'real'
         inputs = fn.inputs.copy()
-        shape = (inputs.pop(bint_var).dtype,) + inputs.pop(reals_var_bound).shape
+        shape = (inputs.pop(bint_var).dtype,) + inputs.pop(diag_var).shape
+        assert reals_var not in inputs
```
eb8680: This doesn't forbid `reals_var == diag_var`, is that still allowed?

fritzo: Yes, that is still allowed.

eb8680: Can you add some test cases exercising that case?

fritzo: Done. Note this is pattern is used by `dist_to_funsor()` and is also tested in test/pyro/test_convert.py.

PR #227, opened 2019-09-06
Refactor alpha conversion logic to allow overloads

fritzo:
Addresses #228 
Blocking #226

This PR has two aims.

1. Refactor alpha conversion logic to allow a fix to `Independent` in #226 .
Before this PR, `alpha_substitute()` was a single function with ad hoc logic to hopefully cover most cases. This PR refactors to a method `._alpha_subs()`, splitting the old if-else handling into method overloads, and adds new method overloads for a few Funsors.

2. Ensure `to_funsor()` is called outside of `substitute()`.
Before this PR, alpha_convert would sometimes call `substitute()` with non-funsor values, necessitating `to_funsor()` conversions in various `eager_subs()` methods. This PR removes those old `to_funsor()` calls and adds new `to_funsor()` calls in the `._alpha_subs()` methods of each `Funsor` subclass with bound variables. This appears to be the narrowest place where relevant type information is available, since types of bound variables are not available in the `.inputs` dict.

## Tested
- added an assertion to `substitute()` checking that all inputs are funsors.

### DIFF
--- a/funsor/terms.py
+++ b/funsor/terms.py
@@ -23,6 +23,7 @@ def substitute(expr, subs):
     if isinstance(subs, (dict, OrderedDict)):
         subs = tuple(subs.items())
     assert isinstance(subs, tuple)
+    assert all(isinstance(v, Funsor) for k, v in subs)
 
     @interpreter.interpretation(interpreter._INTERPRETATION)  # use base
     def subs_interpreter(cls, *args):
@@ -48,7 +49,7 @@ def _alpha_mangle(expr):
         return expr
 
     ast_values = expr._alpha_convert(alpha_subs)
-    return reflect(type(expr), *ast_values, alpha_mangle=False)
+    return reflect(type(expr), *ast_values)
 
 
 def reflect(cls, *args, **kwargs):
@@ -69,8 +70,7 @@ def reflect(cls, *args, **kwargs):
     result._ast_values = args
 
     # alpha-convert eagerly upon binding any variable
-    if kwargs.pop("alpha_mangle", True):
-        result = _alpha_mangle(result)
+    result = _alpha_mangle(result)
 
     cls._cons_cache[cache_key] = result
     return result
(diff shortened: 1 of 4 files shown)


funsor/terms.py
```diff
@@ -23,6 +23,7 @@ def substitute(expr, subs):
     if isinstance(subs, (dict, OrderedDict)):
         subs = tuple(subs.items())
     assert isinstance(subs, tuple)
+    assert all(isinstance(v, Funsor) for k, v in subs)
```
fritzo: This check is half of the purpose of this PR.

funsor/terms.py
```diff
@@ -36,38 +37,22 @@ def subs_interpreter(cls, *args):
         return interpreter.reinterpret(expr)
 
 
-def alpha_substitute(expr, alpha_subs):
-
-    new_values = []
-    for v in expr._ast_values:
-        v = substitute(v, alpha_subs)
-        if isinstance(v, str) and v not in expr.fresh:
-            v = alpha_subs.get(v, v)
```
eb8680: Is this the only piece of logic that actually changed here or in #226 ?

funsor/terms.py
```diff
@@ -328,6 +313,14 @@ def pretty(self, maxlen=40):
     def __contains__(self, item):
         raise TypeError
 
+    def _alpha_convert(self, alpha_subs):
+        """
+        Rename bound variables while preserving all free variables.
+        """
+        # Substitute all funsor values.
+        # Subclasses must handle string conversion.
+        return tuple(substitute(v, alpha_subs) for v in self._ast_values)
```
eb8680: Now that the logic has been split up across so many terms, can you add an assertion here checking that `alpha_subs` is only applied to bound variables?  Something like
```py
assert not frozenset(alpha_subs) - self.bound
```

funsor/terms.py
fritzo: Yes, you could see this PR as 1. changing the logic for alpha converting non-Funsors; and 2. (unrelated to this line) moving the `to_funsor()` logic around. #226 additionally changes the `Independent` signature.

eb8680: I mean lines 44 and 45 specifically - it looks like everything else is the same, just duplicated?

funsor/terms.py
fritzo: Done.

fritzo:
Thanks for reviewing!

Issue #228, opened 2019-09-06
Add BART forecasting example


fritzo:
This issues tracks changes needed to port the [BART forecasting example](https://github.com/pyro-ppl/sandbox/blob/funsor-bart/2019-08-time-series/bart/forecast.py) to funsor.

PR #230, opened 2019-09-06
Implement substitution into Slice funsors

fritzo:
Addresses #228 

This implements substitution into `Slice` funsors and fixes a bug in `.fresh`.

This also adds some assertions in funsor.pyro.convert that helped catch the missing patterns.

## Tested
- added unit tests
- refactoring is exercised by `test_sum_product.py` and `test_torch.py`
- tested in BART forecasting example #228


fritzo:
Thasks for reviewing!

PR #232, opened 2019-09-08
Fix alpha conversion bugs in Subs, Cat, Slice

fritzo:
Addresses #228 

This fixes bugs in alpha conversion logic of `Subs`, `Cat`, and `Slice`. Additionally this moves type checking logic from `substitute` down to `SubsMeta` so that type mismatch errors are caught earlier.

The `Subs` bugs involve an incorrect definition of `.bound` and incorrect handling of substitution in `._alpha_convert()`. For example this case previously failed:
```py
with interpretation(reflect):
    Slice('t', 3)(t=Slice('t', 2, dtype=3)).reduce(ops.add)
```

The main `Cat` bug is deeper and involves type mismatch between the axis variable of the component parts (this variable is bound) and the resulting axis variable of the concatenation. Before this PR `Cat` did not bind any variables. After this PR `Cat` has two real variables that can share the same name: a longer `fresh` concatenated variable and a shorter `bound` part variable; only the bound variable is renamed during `._alpha_convert()`; and only the fresh variable is substituted during `.eager_subs()`.

The `Slice.eager_subs` bug from #230 involved not preserving `.dtype` information.

## Tested

- [x] refactoring is exercised by existing tests
- [x] made substitution type checking stricter
- [x] added a regression test for `Subs` and `Slice`
- [x] added a regression test for `Cat`
- [x] ran BART example #228

### DIFF
--- a/funsor/terms.py
+++ b/funsor/terms.py
@@ -271,6 +271,10 @@ class Funsor(object, metaclass=FunsorMeta):
     hashable ``*args`` and no optional ``**kwargs`` so as to support cons
     hashing.
 
+    Derived classes with ``.fresh`` variables must implement an
+    :meth:`eager_subs` method. Derived classes with ``.bound`` variables must
+    implement an :meth:`_alpha_convert` method.
+
     :param OrderedDict inputs: A mapping from input name to domain.
         This can be viewed as a typed context or a mapping from
         free variables to domains.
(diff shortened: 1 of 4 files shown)


funsor/terms.py
```diff
@@ -271,6 +271,10 @@ class Funsor(object, metaclass=FunsorMeta):
     hashable ``*args`` and no optional ``**kwargs`` so as to support cons
     hashing.
 
+    Derived classes with ``.fresh`` variables must implement an
```
fritzo: This has pleasing symmetry 🙂

funsor/terms.py
```diff
@@ -326,17 +330,11 @@ def __call__(self, *args, **kwargs):
...
-        # Eagerly restrict to this funsor's inputs and convert to_funsor().
+        # Eagerly restrict to this funsor's inputs.
         subs = OrderedDict(zip(self.inputs, args))
         for k in self.inputs:
             if k in kwargs:
                 subs[k] = kwargs[k]
-        for k, v in subs.items():
-            v = to_funsor(v, self.inputs[k])
-            if v.output != self.inputs[k]:
-                raise ValueError("Expected substitution of {} to have type {}, but got {}"
```
fritzo: This check now lives in `to_funsor()` as called from `SubsMeta`.

funsor/terms.py
```diff
@@ -1275,47 +1285,82 @@ def eager_stack_homogeneous(name, *parts):
     return None  # defer to default implementation
 
 
-class Cat(Funsor):
+class CatMeta(FunsorMeta):
+    """
+    Wrapper to fill in default value for ``part_name``.
```
fritzo: The default value allows the old `Cat(name, parts)` syntax to still work.

PR #245, opened 2019-09-19
Add pattern for MultivariateNormal(affine)

fritzo:
Addresses #72 
pair coded with @eb8680 

## Tested
- [x] unit tests for `MultivariateNormal(x + y, ...)` etc.
- [x] sensor fusion example using `dist.MultivariateNormal` directly


funsor/distributions.py
```diff
@@ -451,22 +452,77 @@ def eager_mvn(loc, scale_tril, value):
...
+                             coeff, prec_sqrt)
+        tensors[i] = coeff
+
+    tensors = torch.broadcast_tensors(*tensors)
+    coeffs, const, scale_tril = tensors[:-2], tensors[-2], tensors[-1]
+    batch_shape = const.shape
+    dim = sum(d.num_elements for d in real_inputs.values())
+
+    loc = BlockVector(batch_shape + (dim,))
+    loc[..., 0] = -const / coeffs[0]  # FIXME consider directly constructing info_vec
```
fehiepsi: I think this should be fine. Constructing info_vec directly seems complicated to me.

funsor/distributions.py
```diff
@@ -451,22 +452,77 @@ def eager_mvn(loc, scale_tril, value):
...
-    int_inputs = OrderedDict((k, v) for k, v in inputs.items() if v.dtype != 'real')
+    real_inputs = OrderedDict((k, v) for k, v in affine.inputs.items() if v.dtype == 'real')
+    assert len(real_inputs) == len(coeffs)
+    tensors = [coeffs[k] for k in real_inputs] + [const, scale_tril]
+    inputs, tensors = align_tensors(*tensors)
 
-    precision = cholesky_inverse(scale_tril)
+    # Incorporate scale_tril before broadcasting.
+    scale_tril = tensors[-1]
+    prec_sqrt = cholesky_inverse(scale_tril)  # FIXME is transpose correct?
```
fehiepsi: I think that this is triangular inverse (inverse of triangular matrix), not cholesky_inverse.

funsor/distributions.py
```diff
@@ -451,22 +452,77 @@ def eager_mvn(loc, scale_tril, value):
...
+    for i in range(len(coeffs)):
+        coeff = tensors[i]
+        inputs1, output1 = equations1[i].split('->')
+        input11, _ = inputs1.split(',')
+        inputs2, output2 = equations2[i].split('->')
+        input21, _ = inputs2.split(',')
+        coeff = torch.einsum(f'...{input11},...{output1}{output2}->...{input21}',
```
fehiepsi: super interesting! I am not sure if this is {output1}{output2} or {output2}{output1} (probably depends on the definition of prec_sqrt - i.e. if transpose is involved). In addition, if we do not broadcast `scale_tril`, I guess `torch.einsum` will fail here because Ellipses of coeff and prec_sqrt are mismatched.

funsor/distributions.py
```diff
@@ -451,22 +452,77 @@ def eager_mvn(loc, scale_tril, value):
...
+    offset1 = 0
+    for i1, (v1, c1) in enumerate(zip(real_inputs, coeffs)):
+        slice1 = slice(offset1, offset1 + real_inputs[v1].num_elements)
+        input11, input12 = equations1[i1].split('->')[0].split(',')
+        offset2 = 0
+        for i2, (v2, c2) in enumerate(zip(real_inputs, coeffs)):
+            slice2 = slice(offset2, offset2 + real_inputs[v2].num_elements)
+            input21, input22 = equations2[i2].split('->')[0].split(',')
+            precision[..., slice1, slice2] = torch.einsum(
+                f'...{input11}{input12},...{input21}{input22}->...{input12}{input22}',
```
fehiepsi: I guess this is not `input11` and `input21`, but something like `input11[0]`.

funsor/distributions.py
fritzo: @fehiepsi can you help me out here, I copied this from the univariate case and could use help with the linear algebra to incorporate `const`.

funsor/distributions.py
fritzo: Good point, maybe I can leverage the new `Einsum` logic from #249 ...

funsor/distributions.py
fehiepsi: Rethinking about this, I do agree that it is better to compute info_vec directly. The way we do this in the univariate case might not be suitable for the case `coeffs[0] = 0`. I think we can process as
```
(a + xB + yC)Psqrt @ transpose_of_previous_term
= (a @ Psqrt + xBPsqrt + yCPsqrt) @ transpose...
= (a' + xB' + yC') @ transpose...
Hence, precision = (B' C') @ (B' C').T
and info_vec = (a' @ B'.T, a' @ C'.T)
```

fehiepsi: I think `loc @ info_vec` is simply `a' @ a'.T`...

fritzo: Great, I've pushed attempted fixes. Can you take another look?

fritzo:
OK, I've fixed most of the shape errors and we're down to numerical errors only :smile:
I'd appreciate any help from reviewers.

funsor/distributions.py
```diff
@@ -449,22 +450,81 @@ def eager_mvn(loc, scale_tril, value):
...
+            f'...{input11},...{output1}->...{input12}', c1, const) \
+            .reshape(batch_shape + (size1,))
+        offset2 = 0
+        for i2, (v2, c2) in enumerate(zip(real_inputs, coeffs)):
+            size2 = real_inputs[v2].num_elements
+            slice2 = slice(offset2, offset2 + size2)
+            inputs2, output2 = equations2[i2].split('->')
+            input21, input22 = inputs2.split(',')
+            precision[..., slice1, slice2] = torch.einsum(
+                f'...{input11},...{input21}->...{input12}{input22}', c1, c2) \
```
fehiepsi: hmm, I thought that `equations1` and `equations2` do not have common notations. So assume input11='ab', input21='AB', I am not sure how this `einsum` work... I guess we have to make `input21='aB'`?

funsor/distributions.py
```diff
@@ -449,22 +450,81 @@ def eager_mvn(loc, scale_tril, value):
...
+                .reshape(batch_shape + (size1, size2))
+            offset2 += size2
+        offset1 += size1
+
+    info_vec = info_vec.as_tensor()
+    precision = precision.as_tensor()
     log_prob = (-0.5 * dim * math.log(2 * math.pi)
```
fehiepsi: I think `dim` should be replaced by scale_tril.shape[-1].

funsor/distributions.py
```diff
@@ -449,22 +450,81 @@ def eager_mvn(loc, scale_tril, value):
...
+    assert tuple(real_inputs) == tuple(coeffs)
+
+    # Incorporate scale_tril before broadcasting.
+    eye = torch.eye(scale_tril.data.size(-1)).expand(scale_tril.data.shape)
+    prec_sqrt = Tensor(eye.triangular_solve(scale_tril.data, upper=False).solution,
+                       scale_tril.inputs)
+    tensors = []
+    for k, (coeff, eqn) in coeffs.items():
+        shape = (prec_sqrt.shape[-1], real_inputs[k].num_elements)
+        tensors.append((prec_sqrt @ coeff.reshape(shape)).reshape(coeff.shape))
```
fehiepsi: I thought that affine (const, coef A, ...) is interpreted as const + x.T @ A + ... (based on einsum formula `ab,a->b`). So when the affine is scaled by `scale_tril`, the new coeff will be `A @ prec_sqrt`, where `prec_sqrt = scale_tril^{-1}.T`. If you define `prec_sqrt = scale_tril^{-1}`, then the new coeff will be `A @ prec_sqrt.T`.

However, I think I miss something here because you have resolved shape errors.

funsor/distributions.py
```diff
@@ -449,22 +450,81 @@ def eager_mvn(loc, scale_tril, value):
...
+@eager.register(MultivariateNormal, (Variable, Contraction), Tensor, Tensor)
+@eager.register(MultivariateNormal, Tensor, Tensor, (Variable, Contraction))
 def eager_mvn(loc, scale_tril, value):
-    if isinstance(loc, Variable):
-        loc, value = value, loc
+    assert len(loc.shape) == 1
+    assert len(scale_tril.shape) == 2
+    assert value.output == loc.output
+
+    affine = loc - value
```
fehiepsi: It might be good to add a comment here that we can represent `(loc - scale) @ prec_sqrt` as an affine, so part of the logic in this implementation can be moved to `extract_affine` utility.

fritzo: Oh good point, that would save some code here!

funsor/distributions.py
fritzo: Thanks! done.

funsor/distributions.py
fritzo: Done.

funsor/distributions.py
fritzo: Good catch! done.

funsor/distributions.py
fritzo: OK, I've used your observation to simplify code.

funsor/distributions.py
```diff
@@ -449,23 +450,79 @@ def eager_mvn(loc, scale_tril, value):
...
+    offset1 = 0
+    for i1, (v1, c1) in enumerate(zip(real_inputs, coeffs)):
+        size1 = real_inputs[v1].num_elements
+        slice1 = slice(offset1, offset1 + size1)
+        inputs1, output1 = equations1[i1].split('->')
+        input11, input12 = inputs1.split(',')
+        assert input11 == input12 + output1
+        info_vec[..., slice1] = torch.einsum(
+            f'...{input11},...{output1}->...{input12}', c1, const) \
+            .reshape(batch_shape + (size1,))
```
fehiepsi: @fritzo I guess we need negative sign here. Everything else looks good to me.

fehiepsi: Nevermind, you catch it. :D

fritzo: done :smile:

fritzo:
Yay all tests pass :smile: Thanks for all the help @fehiepsi !

test/examples/test_sensor_fusion.py
```diff
@@ -13,8 +14,49 @@
 from funsor.torch import Tensor
 
 
+# This version constructs factors using funsor.distributions.
+@pytest.mark.parametrize('state_dim,obs_dim', [(3, 2), (2, 3)])
+def test_distributions(state_dim, obs_dim):
```
fritzo: @jpchen you can follow the idioms of this test in your experiment.

funsor/distributions.py
```diff
@@ -449,23 +450,80 @@ def eager_mvn(loc, scale_tril, value):
...
+    affine = prec_sqrt @ (loc - value)
+    const, coeffs = extract_affine(affine)
+    if not isinstance(const, Tensor):
+        return None  # lazy
+    if not all(isinstance(coeff, Tensor) for coeff, _ in coeffs.values()):
+        return None  # lazy
+
+    # Compute log_prob using funsors.
+    scale_diag = Tensor(scale_tril.data.diagonal(dim1=-1, dim2=-2), scale_tril.inputs)
+    log_prob = (-0.5 * scale_diag.shape[0] * math.log(2 * math.pi)
```
fehiepsi: Just curious: why we use `shape[0]` instead of `shape[-1]`, `scale_diag.log().sum()` instead of `scale_diag.log().sum(-1)` here?

fritzo: Here `scale_diag` is a funsor, and it separates "batch" `.inputs` from "event" `.shape`. In fact `scale_diag.shape == (dim,)` regardless of batching. This also allows us to call `.sum()` below rather than `.sum(-1)`, since there is only one tensor dimension.

fehiepsi: That's nice! Thanks for explaining.

Issue #235, opened 2019-09-12
Add lazy multidimensional Unit terms


eb8680:
It would be useful to have unit terms corresponding to deferred calls to `torch.zeros`/`torch.ones` that conform to constant propagation rules #163 .  This would improve cons hashing accuracy and allow us to avoid reifying large unit tensors as can sometimes happen in e.g. `funsor.adjoint`.

eb8680:
I'm no longer convinced this is necessary - we should never need to `expand` constants.

PR #249, opened 2019-09-20
Implement affine funsor approximation

fritzo:
Addresses #72 
Blocking #245 (from which this PR was extracted)

This implements a general `extract_affine(-)` function that produces a canonical representation of any affine function, including constants, sums, einsums, indexing, reshaping and any other linear combination. The only requirement is that the function can be eagerly evaluated on `Tensor`s. I was surprised how general this turned out to be.

This required some fixes to `Einsum` to handle differently-batched inputs.

## Tested
- [x] unit test for `extract_affine()`
- [x] extra test for batched `Einsum`

### DIFF
--- a/funsor/affine.py
+++ b/funsor/affine.py
@@ -19,7 +19,12 @@ def extract_affine(fn):
                 for var, (coeff, eqn) in coeffs.items())
         assert_close(y, x)
 
-    :param Funsor fn: A funsor assumed to be affine (this is not checked).
+    The affine approximation is computed by ev evaluating ``fn`` at
+    zero and each basis vector. To improve performance, users may want to run
+    under the :func:`~funsor.memoize.memoize` interpretation.
+
+    :param Funsor fn: A funsor assumed to be affine wrt the (add,mul) semiring.
+       The affine assumption is not checked.
     :return: A pair ``(const, coeffs)`` where const is a funsor with no real
         inputs and ``coeffs`` is an OrderedDict mapping input name to a
         ``(coefficient, eqn)`` pair in einsum form.
--- a/funsor/affine.py
+++ b/funsor/affine.py
@@ -1,0 +1,48 @@
+from collections import OrderedDict
+
+import opt_einsum
+import torch
+
+from funsor.interpreter import gensym
+from funsor.terms import Lambda, Variable, bint
+from funsor.torch import Tensor
+
+
+def extract_affine(fn):
+    """
+    Extracts an affine representation of a funsor, which is exact for affine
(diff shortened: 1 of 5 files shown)


funsor/affine.py
```diff
@@ -0,0 +1,47 @@
...
+    """
+    # Determine constant part by evaluating fn at zero.
+    real_inputs = OrderedDict((k, v) for k, v in fn.inputs.items() if v.dtype == 'real')
+    zeros = {k: Tensor(torch.zeros(v.shape)) for k, v in real_inputs.items()}
+    const = fn(**zeros)
+
+    # Determine linear coefficients by evaluating fn on basis vectors.
+    name = gensym('probe')
+    coeffs = OrderedDict()
+    for k, v in real_inputs.items():
```
eb8680: Wrap this loop in `with funsor.memoize()` to avoid repeated tensor computations?  We should also be able to use #235 for further savings once that's implemented

funsor/affine.py
```diff
@@ -0,0 +1,47 @@
...
+
+from funsor.interpreter import gensym
+from funsor.terms import Lambda, Variable, bint
+from funsor.torch import Tensor
+
+
+def extract_affine(fn):
+    """
+    Extracts an affine representation of a funsor, which is exact for affine
+    funsors and approximate otherwise. For affine funsors this satisfies::
```
eb8680: Add a brief explanation of the approximation here?

funsor/affine.py
```diff
@@ -0,0 +1,47 @@
...
+
+        x = ...
+        const, coeffs = extract_affine(x)
+        y = sum(Einsum(eqn, (coeff, Variable(var, coeff.output)))
+                for var, (coeff, eqn) in coeffs.items())
+        assert_close(y, x)
+
+    :param Funsor fn: A funsor assumed to be affine (this is not checked).
```
eb8680: nit: mention explicitly that this currently only applies to the `add`/`mul` semiring

fehiepsi:
@fritzo @eb8680 I have just walked through `Normal(Affine...)` implementation. Now, I am looking at `MVN(Affine...)` one. I guess I can mimic `MVN(Affine...)` for Gaussian affine without the need of understanding `extract_affine` functionality. :D So please merge this as your convenience.

funsor/affine.py
fritzo: I'd prefer to implement this simple version first. Most of our funsor expressions have 1 or 2 real free variables, and I'd rather get lots of tests in place before introducing extra complexity.

fritzo: I've added a note that "users may want to run this under the memoize interpretation".

Issue #207, opened 2019-08-31
Support multiple backends


fritzo:
This issue tracks changes needed to fully support multiple backends in an interchangeable/generic way. The initial release of Funsor implements most funsors using PyTorch (funsor.torch implements `Tensor`, but all of funsor.gaussian, .joint, .distributions use PyTorch internally.

## Approach 1
One possibility is to replicate backend-specific logic in a consistent way, e.g. with module structure
- funsor
  - torch
    - tensor
    - gaussian
    - cnf
    - distributions
  - numpy / jax
    - tensor
    - gaussian
    - cnf
    - distributions
  - tensorflow
    - ...
  - birch
    - ...

## Approach 2
A second possibility (suggested by @eb8680 below and in #159) is to
1. subclass `Tensor` into `TorchTensor` and `NumpyTensor`, then
2. refactor `Gaussian` and `Contraction` etc. to use `Tensor`s rather than torch.Tensors internally.

This approach would still require duplicating distributions.

## Potential backends
- [x] PyTorch / `pyro.distributions`
- [x] #58 NumPy
- [x] JAX / `numpyro.distributions`
- [ ] Tensorflow 2.0 / `tfp.distributions`
- [ ] Trax's TensorFlow-backed NumPy API [`tf_numpy`](https://github.com/google/trax/tree/master/trax/tf_numpy)
- [ ] Birch 

## Other tasks
- [ ] Remove stray dependencies on PyTorch or NumPy e.g. in `funsor.testing` and `funsor.domains`
- [x] Add backend-specific installation options `funsor[torch]`, `funsor[numpy]` etc.

eb8680:
Related suggestion in #159.  Agree we should be able to make `Number`, `Tensor`, `Gaussian`, `Delta`, `Distribution` interfaces backend-agnostic and have backend independence follow easily from there.

eb8680:
I believe the bulk of this work is finished for the PyTorch and JAX backends as of #327, but I'll leave this issue open until #332 is addressed.

eb8680:
I'm going to close this as resolved. If we want to add more backends in the future or make `Gaussian` use `funsor.Tensor` for its parameters, we can track progress in separate issues.

PR #250, opened 2019-09-20
Implement ops.matmul and align_tensors(..., expand=True)

fritzo:
Blocking #245 

This adds an op `x @ y = ops.matmul(x, y)` which is proving to simplify math in #245 . This is a step in the direction of performing more underlying math on Funsors and less on torch.Tensors, addressing #207 .

This also adds an `expand=True` kwarg to `align_tensor()` and `align_tensors()`, thereby avoiding an error-prone `torch.broadcast_tensors()` in subsequent code.

## Tested
- [x] unit tests for matmul
- [x] refactored to use `x @ y` in some code (covered by existing tests)
- [x] refactored to use `expand=True` in some code (covered by existing tests)

### DIFF
--- a/funsor/distributions.py
+++ b/funsor/distributions.py
@@ -368,8 +368,7 @@ def eager_normal(loc, scale, value):
     if isinstance(loc, Variable):
         loc, value = value, loc
 
-    inputs, (loc, scale) = align_tensors(loc, scale)
-    loc, scale = torch.broadcast_tensors(loc, scale)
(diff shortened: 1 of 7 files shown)


eb8680:
WDYT about desugaring `matmul` to `funsor.torch.Einsum` instead of implementing separate `Binary(matmul, ...` patterns?

fritzo:
I'd rather implement native `matmul` than desugar to `Einsum` for now.

PR #252, opened 2019-09-23
Fix strategy of gaussian mixture .unscaled_sample()

fritzo:
This fixes a subtle statistical inefficiency bug in `Contraction.unscaled_sample()`, especially important for scale mixtures.

Before this PR our strategy for sampling the discrete component of a `Tensor+Gaussian` mixture was to monte carlo sample from the `Tensor` and importance weight from the `Gaussian.log_normalizer`.

After this PR our strategy is to monte carlo sample from the true distribution.

## Tested
- added a mixture sampling test (passes before and after this PR though)

### DIFF
--- a/funsor/cnf.py
+++ b/funsor/cnf.py
@@ -17,7 +17,12 @@ from funsor.util import quote
 
 class Contraction(Funsor):
     """
-    Declarative representation of a finitary sum-product operation
+    Declarative representation of a finitary sum-product operation.
+
+    After normalization via the :func:`~funsor.terms.normalize` interpretation
+    contractions will canonically order their terms by type::
+
+        Delta, Number, Tensor, Gaussian
     """
     def __init__(self, red_op, bin_op, reduced_vars, terms):
         terms = (terms,) if isinstance(terms, Funsor) else terms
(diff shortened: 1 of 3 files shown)


funsor/cnf.py
```diff
@@ -70,9 +75,46 @@ def _is_affine(self):
         return True
 
     def unscaled_sample(self, sampled_vars, sample_inputs):
-        if self.red_op in (ops.logaddexp, nullop) and self.bin_op in (ops.logaddexp, ops.add, nullop):
```
fritzo: The previous handling of case `self.bin_op == ops.add` was just barely correct due to subtleties in `Gaussian`. I've made `Gaussian` a little more transparent, and written out more detailed logic here so we can see what is going on.

funsor/gaussian.py
```diff
@@ -498,11 +498,13 @@ def eager_reduce(self, op, reduced_vars):
         return None  # defer to default implementation
 
     def unscaled_sample(self, sampled_vars, sample_inputs):
-        # Sample only the real variables.
-        sampled_vars = frozenset(k for k, v in self.inputs.items()
-                                 if k in sampled_vars if v.dtype == 'real')
+        sampled_vars = sampled_vars.intersection(self.inputs)
         if not sampled_vars:
             return self
+        if any(self.inputs[k].dtype != 'real' for k in sampled_vars):
```
fritzo: After this PR we don't allow sampling discrete parts of a raw `Gaussian`; in practice this should never happen, any real distribution should pair a `Gaussian` with a normalizing `Tensor`.

PR #253, opened 2019-09-23
Refactor Integrate patterns

fritzo:
Addresses #228 

This is an experimental refactoring of `Integrate` patterns to make logic more straight-forward.

After this PR `Integrate()` restricts approximation to only `log_measure` and no subterms. To achieve this I have remove the `integrate <--> binary` patterns and exp/log logic introduced in #157.

## Tested
- [ ] fix `test_reduce_moment_matching_moments` which started xfailing in #157

### DIFF
--- a/funsor/integrate.py
+++ b/funsor/integrate.py
@@ -1,11 +1,13 @@
+import functools
 from collections import OrderedDict
+from functools import reduce
 from typing import Union
 
 import funsor.ops as ops
 from funsor.cnf import Contraction, GaussianMixture
 from funsor.delta import Delta
 from funsor.gaussian import Gaussian, _mv, _trace_mm, _vv, align_gaussian, cholesky_inverse, cholesky_solve
-from funsor.terms import Funsor, Number, Subs, Unary, Variable, eager, normalize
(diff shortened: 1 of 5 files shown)


eb8680:
+1 to the file reorganization in this PR, can we keep that without reintroducing the duplicate logic of `integrator`?  The changes in #254 should achieve the same goal.

fritzo:
> reintroducing the duplicate logic of `integrator`

Wait, last week we agreed we needed the following duplicate logic prefixing each eager integrate pattern:
```py
if not (reduced_vars.issubset(log_measure.inputs) and
        reduced_vars.issubset(integrand.inputs)):
    return None  # normalize
```
This PR only reintroduces the `@integrator` decorator to collect this new minimal logic. This PR does not reintroduce the duplicated normalization logic from before #157 .

Can you clarify your comment? Do you want me to inline those three lines in each pattern?

eb8680:
> Can you clarify your comment?

Sorry for the confusion, I meant the `normalize_integrate` pattern that reduces variables only appearing in one side of an `Integrate` (which was formerly in `integrator` pre-#157).  That should be taken care of by the changes in #254.  `integrator` in this PR is fine.

PR #259, opened 2019-09-26
Add adjoint methods to support backwards marginal through MarkovProduct

eb8680:
Addresses #228 

This PR adds adjoint machinery necessary for computing backward marginals through `sequential_sum_product`.  In particular, it adds general adjoint methods for `Subs` of `Tensor` and for `Cat`, completing support for the operations appearing in `sequential_sum_product` and unblocking the adjoint experiments.  The adjoint for `Subs` of `Tensor` is implemented in such a way that generalizing it to integer variable substitution in `Gaussian`s should be very straightforward, and I'll do that in a followup PR.

I've also simplified `adjoint` slightly by removing the multiplicities that were being tracked previously and removing the one suspect test that required their presence.

Remaining tasks:
- [x] In theory `adjoint` should work for eager `MarkovProduct` evaluation as well, but even though it simply calls `sequential_sum_product` internally, the test cases that call `MarkovProduct` are failing.  I'm not sure what's happening there, but would like to include a fix in this PR. (Update: fixed)

Tested:
- Added test comparing serial and parallel sequential sum product adjoints

### DIFF
--- a/funsor/adjoint.py
+++ b/funsor/adjoint.py
@@ -8,7 +8,7 @@ from funsor.cnf import Contraction, nullop
 from funsor.interpreter import interpretation
 from funsor.ops import AssociativeOp
 from funsor.registry import KeyedRegistry
-from funsor.terms import Binary, Funsor, Reduce, Variable, lazy, to_funsor
+from funsor.terms import Binary, Funsor, Reduce, Subs, Variable, lazy, to_funsor
 from funsor.torch import Tensor
 
 
@@ -30,7 +30,7 @@ class AdjointTape(object):
     def __call__(self, cls, *args):
         with interpretation(self._old_interpretation):
             result = cls(*args)
-        if issubclass(cls, (Reduce, Contraction, Binary, Tensor)):  # TODO make generic
+        if issubclass(cls, (Reduce, Contraction, Binary, Tensor, Subs)):  # TODO make generic
             self.tape.append((result, cls, args))
         return result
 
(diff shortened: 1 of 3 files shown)


funsor/adjoint.py
```diff
@@ -149,3 +154,112 @@ def adjoint_contract(adj_redop, adj_binop, out_adj, sum_op, prod_op, reduced_var
...
+    return {arg: arg_adj}
+
+
+def _scatter(src, res, subs):
+    # inverse of advanced indexing
+    assert isinstance(res, Tensor)
+    assert isinstance(src, Tensor)
+    # TODO check types of subs, in case some logic from eager_subs was accidentally left out?
+
+    # use advanced indexing logic copied from Tensor.eager_subs:
```
eb8680: The following code is copied verbatim from `Tensor.eager_subs`

funsor/adjoint.py
```diff
@@ -149,3 +154,112 @@ def adjoint_contract(adj_redop, adj_binop, out_adj, sum_op, prod_op, reduced_var
...
+                (-1,) + (1,) * offset_from_right))
+
+    # Construct a [:] slice for the output.
+    for i, size in enumerate(res.output.shape):
+        offset_from_right = len(res.output.shape) - i - 1
+        index.append(torch.arange(size).reshape(
+            (-1,) + (1,) * offset_from_right))
+
+    # the only difference from Tensor.eager_subs is here:
+    # instead of indexing the rhs (lhs = rhs[index]), we index the lhs (lhs[index] = rhs)
```
eb8680: The preceding code is copied verbatim from `Tensor.eager_subs`

funsor/adjoint.py
```diff
@@ -149,3 +154,112 @@ def adjoint_contract(adj_redop, adj_binop, out_adj, sum_op, prod_op, reduced_var
...
+    # unsqueeze to make broadcasting work
+    src_inputs, src_data = src.inputs.copy(), src.data
+    for k, v in res.inputs.items():
+        if k not in src.inputs and isinstance(subs[k], Number):
+            src_inputs[k] = bint(1)
+            src_data = src_data.unsqueeze(-1)
+    src = Tensor(src_data, src_inputs, src.output.dtype).align(tuple(res.inputs.keys()))
+
+    data = res.data
+    data[tuple(index)] = src.data
```
eb8680: Here we reuse the indexing logic from `Tensor.eager_subs` to do a fancy broadcasted assignment instead of a fancy broadcasted selection.

test/test_adjoint.py
```diff
@@ -167,3 +158,60 @@ def test_optimized_plated_einsum_adjoint(equation, plates, backend):
...
+], ids=str)
+@pytest.mark.parametrize('batch_inputs', [
+    {},
+    {"foo": bint(5)},
+    {"foo": bint(2), "bar": bint(4)},
+], ids=lambda d: ",".join(d.keys()))
+@pytest.mark.parametrize('impl', [
+    sequential_sum_product,
+    naive_sequential_sum_product,
+    xfail_param(MarkovProduct, reason="something missing?"),
```
eb8680: These test cases fail mysteriously at the final assertion with incorrect tensor values

funsor/adjoint.py
fritzo: Would it be easy to this out as a helper, say `funsor.torch.make_advanced_index(tensor, subs)`, that could be used in `funsor.torch.Tensor.eager_subs` and imported here?

test/test_adjoint.py
fritzo: My first guess is an alpha renaming issue, which is the main difference between `sequential_sum_product` and `MarkovProduct`.

funsor/adjoint.py
fritzo: In particular, the keys and values of the `MarkovProduct.step_names` dict will typically unmangle to the same name.

test/test_adjoint.py
fritzo: In particular, the keys and values of the `MarkovProduct.step_names` dict will typically unmangle to the same name.

funsor/adjoint.py
eb8680: Probably, I'll do a cleanup PR after the `Gaussian` version of this PR

funsor/adjoint.py
```diff
@@ -28,10 +41,12 @@ def __init__(self):
...
     def __call__(self, cls, *args):
-        with interpretation(self._old_interpretation):
-            result = cls(*args)
-        if issubclass(cls, (Reduce, Contraction, Binary, Tensor)):  # TODO make generic
+        if cls in adjoint_ops:  # atomic op, don't trace internals
+            with interpretation(self._old_interpretation):
+                result = cls(*args)
             self.tape.append((result, cls, args))
+        else:
+            result = self._old_interpretation(cls, *args)
```
eb8680: FYI this change fixed the issues with `MarkovProduct` - the internal operations weren't being recorded

funsor/adjoint.py
```diff
@@ -61,19 +74,19 @@ def adjoint(self, red_op, bin_op, root, targets):
...
 
             in_adjs = adjoint_ops(fn, red_op, bin_op, adjoint_values[output], *inputs)
             for v, adjv in in_adjs.items():
-                multiplicities[v] += 1
                 adjoint_values[v] = bin_op(adjoint_values[v], adjv)
 
         target_adjs = {}
         for v in targets:
-            target_adjs[v] = adjoint_values[v] / multiplicities[v]  # TODO use correct op here with bin_op
+            target_adjs[v] = adjoint_values[v]
```
fritzo: :+1:

PR #246, opened 2019-09-19
bias estimation experiment

jpchen:
paircoded with @fritzo 

~~Blocked by #72~~

This adds an example using exact Gaussian sensor fusion to estimate sensor bias in a multi-sensor single-object tracking problem. Right now the numbers in the paper is using random acceleration model.  It also is doing manual sequential sum product rather than GaussianHMM (missing pattern) or `sequential_sum_product()` (todo).  The probabilistic model is
```py
process_noise = pyro.param(...)
sensor_noise = pyro.param(...)
bias_scale = pyro.param(...)

with plate("sensors", num_sensors):
    bias ~ MVN(0, bias_scale)
states = {}
states[-1] ~ MVN(0, init_state_cov)
for t in range(num_time_steps):
    states[t] ~ MVN(states[t - 1] @ ncv_dynamics, process_noise)
    with plate("sensors", num_sensors) as sensor_id:
        pyro.sample(MVN(states[t] @ obs_matrix + bias[sensor_id], sensor_noise)
                    obs=data[t])
```
you can run the experiments via `make sensor`.

~~i didn't include the plotting code which is in a notebook since it clutters the code.~~  @fritzo feel free to push directly to this branch. when this is ready to merge, i'd like to force-push squash the commits before merging.

PR #263, opened 2019-10-03
Add MultivariateNormal pattern for Binary to cover matmul

fritzo:
Unblocks #246 

This adds a pattern for `Binary` to the `MultivariateNormal` pattern matching code. Note this pattern is more complete but still unsound; soundness will be addressed by `.is_affine` #72

## Tested
- added a unit test


jpchen:
thanks for your help in finding this!

PR #266, opened 2019-10-10
Add test of pyro-api

fritzo:
Addresses https://github.com/pyro-ppl/pyro-api/issues/2

This adds tests for the generic `pyro-api`. I think we can copy this pattern in numpyro and pyro nearly verbatim.


fritzo:
@eb8680 until we publish pyroapi on pypi (soon), you'll need to install pyro-api to run tests locally:
```sh
git clone git@github.com:pyro-ppl/pyro-api
cd pyro-api
pip install -e .
```

PR #269, opened 2019-10-11
Add seed handler and tests

fritzo:
Blocked by https://github.com/pyro-ppl/pyro/pull/2061

## Tested

- added unit tests

### DIFF
--- a/test/test_minipyro.py
+++ b/test/test_minipyro.py
@@ -78,10 +78,10 @@ def assert_warning(model, guide, elbo):
 @pytest.mark.parametrize("backend", ["pyro", "minipyro", "funsor"])
 def test_generate_data(backend):
 
-    def model(data=None):
+    def model():
         loc = pyro.param("loc", torch.tensor(2.0))
         scale = pyro.param("scale", torch.tensor(1.0))
-        x = pyro.sample("x", dist.Normal(loc, scale), obs=data)
+        x = pyro.sample("x", dist.Normal(loc, scale))
         return x
 
     with pyro_backend(backend):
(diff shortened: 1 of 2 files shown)


test/test_minipyro.py
```diff
@@ -90,6 +90,59 @@ def model(data=None):
...
+        with handlers.seed(rng_seed=0):
+            model()
+            with handlers.seed(rng_seed=0):
+                model()
+            actual = model()
+        assert_close(actual, expected)
+
+
+@pytest.mark.parametrize("backend", ["pyro", "minipyro", "funsor"])
+def test_rng_state(backend):
```
neerajprad: Is the rng_state being tested because of the default seed handler in `pyro_backend`?

fritzo: No, the rng_state is being tested irrelevant of the behavior of pyro_backend. I just wanted to add a test, and the natural place for that was in test_minipyro.py.

test/test_minipyro.py
```diff
@@ -3,7 +3,7 @@
 import pytest
 import torch
 from pyro.generic import distributions as dist
-from pyro.generic import infer, optim, pyro, pyro_backend
+from pyro.generic import handlers, infer, optim, pyro, pyro_backend
```
neerajprad: You can probably start importing this from pyroapi now (since it is a test dependency)?

test/test_minipyro.py
neerajprad: It seems like there's a name clash with the test above?

fritzo: fixed.

test/test_minipyro.py
fritzo: fixed.

PR #270, opened 2019-10-11
Add zeros, ones, randn to funsor.compat.ops

fritzo:
Addresses https://github.com/pyro-ppl/pyro-api/pull/5

This creates a new `funsor.compat.ops` with zeros, ones, and randn. I've also moved `tensor` here and out of the core `funsor.ops`. After this PR I will update pyroapi.dispatch.

### DIFF
--- a/funsor/compat/ops.py
+++ b/funsor/compat/ops.py
@@ -1,0 +1,3 @@
+from torch import ones, randn, tensor, zeros
+
+from .ops import *  # noqa F401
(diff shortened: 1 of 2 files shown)


funsor/compat/ops.py
```diff
@@ -0,0 +1,3 @@
+from torch import ones, randn, tensor, zeros
```
neerajprad: You'll need noqa to make the linter happy.

Issue #233, opened 2019-09-08
Change .reduce() interface to require Variables?


fritzo:
Currently the method `Funsor.reduce(op, vars)` requires a set of strings. This design choice forces `.reduce()` to require all vars to appear in the `.inputs` of the funsor being reduced. If instead we required `.reduce()` to input `Variables` (as requested in #211 ), then those variables would have sufficient type metadata to allow reduction over constant terms: for `bint()` reduction variables we would multiply by `var.output.size`, and for `real()` variables we could either do nothing (now) or later multiply by the unit infinite as suggested in #53 .

One advantage of this refactoring is that funsors could drop trivial inputs of type `bint(1)`.

Note following #211 we could still allow string inputs, but raise say a `TypeError` if there is insufficient type metadata to convert all strings to `Variables` (i.e. if any reduction variable does not appear free in the term being reduced). This would ensure backward compatibility.

fritzo:
Fixed by #276

PR #276, opened 2019-10-16
Accept more types to .reduce(), .sample(), Integrate()

fritzo:
Addresses #211 #233 

This adds a `_convert_reduced_vars()` helper used in `.reduce()`, `.sample()`, and `Integrate()` to convert strings, Variables, and sets and frozensets thereof to frozensets of strings. This refactoring will also be useful in #233 .

## Tested
- added unit tests for `.reduce()`, `Integrate()`
- rewrote many examples and tests to use the new sugared syntax

### DIFF
--- a/funsor/terms.py
+++ b/funsor/terms.py
@@ -288,6 +288,27 @@ def _issubclass_tuple(subcls, cls):
     return all(_issubclass_tuple(a, b) for a, b in zip(subcls.__args__, cls.__args__))
 
 
+def _convert_reduced_vars(reduced_vars):
+    """
+    Helper to convert the reduced_vars arg of ``.reduce()`` and friends.
+
+    :param reduced_vars:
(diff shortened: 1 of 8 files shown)


funsor/terms.py
```diff
@@ -440,12 +457,14 @@ def sample(self, sampled_vars, sample_inputs=None):
         If ``sample_inputs`` is provided, this creates a batch of samples
         scaled samples.
 
-        :param frozenset sampled_vars: A set of input variables to sample.
+        :param sampled_vars: A set of input variables to sample.
+        :type sampled_vars: str, Variable, or set or frozenset thereof.
```
eb8680: Does Sphinx render this `:type ...:` correctly?

fritzo: Yes, it sphinx does pretty well here:
image

Issue #201, opened 2019-08-23
Fix docs for initial release


fritzo:
## Tasks
- [x] #200 Fix readthedocs build
- [x] #202 ensure all modules appear in index.rst
- [x] #277 add docstrings to funsor.pyro.hmm
- [x] #280 ensure all user-facing Funsor constructors are documented (i.e. with `:param: ...`)
- [ ] #283 update README.md, especially the section on organization

## Triaged
- work around lack of support for default options (commented out in #200)

PR #280, opened 2019-10-17
Add Funsor docstrings

fritzo:
Addresses #201 

This adds docstrings to all remaining funsors. I've also added some magic to make ops show up in sphinx.

## Tested
- examined docs and verified some links locally


funsor/ops.py
```diff
@@ -347,3 +347,14 @@ def reciprocal(x):
...
+__doc__ = """
+Built-in operations
+-------------------
+
+{}
+
+Operation classes
+-----------------
+""".format("\n".join(f".. autodata:: {_name}\n"
+                     for _name in __all__ if isinstance(globals()[_name], Op)))
```
eb8680: Neat - should do something similar for patterns/interpretations?  We could add docstrings to (important) patterns and auto-generate separate doc pages for each interpretation.

fritzo: Hmm, interesting. I'll have to think about how to do that. @jpchen had suggested making automatically making latex tables as in our paper.

Issue #238, opened 2019-09-16
Implement optimized GEMM-like ops for Tensor and Gaussian


fritzo:
- [ ] for `Tensor` use a GEMM op, fixing the bugs in https://github.com/pyro-ppl/funsor/blob/d18c78e/funsor/torch.py#L495 (which only works for scalar event shape)
- [ ] for `Gaussian` see pyro.ops.gaussian.gaussian_tensordot() pyro-ppl/pyro#1980

fritzo:
@eb8680 can you confirm that I can implement these as something like
```py
@eager.register(Contraction, AddOp, MulOp, frozenset, Variadic[Tensor])
def eager_contract_tensor(red_op, bin_op, reduced_vars, *operands):
    equation = "TODO define an einsum string"
    data = pyro.ops.einsum.contract(equation, *(x.data for x in operands))
    inputs = "TODO"
    output = "TODO"
    return Tensor(data, inputs, output)

@eager.register(Contraction, LogaddexpOp, AddOp, frozenset, Variadic[Tensor])
def eager_contract_tensor(red_op, bin_op, reduced_vars, *operands):
    ...backend='pyro.ops.einsum.torch_log'...
```

eb8680:
> can you confirm

Yep, this should work as expected.

fritzo:
@fehiepsi did you already complete this?

PR #281, opened 2019-10-17
Implement eager contract (Tensor,Tensor) via einsum

fritzo:
Addresses #238
Blocked by https://github.com/pyro-ppl/pyro/pull/2065

This implements `Contraction(-,-,-, Tensor, Tensor)` via `einsum`, using either the `torch` or `pyro.ops.einsum.torch_log` backend.

This also starts to use `Contraction()` inside `sequential_sum_product()` so as to make this cheaper op available under the eager interpretation.

## Tested
- added a unit test

### DIFF
--- a/funsor/cnf.py
+++ b/funsor/cnf.py
@@ -43,7 +43,6 @@ class Contraction(Funsor):
         elif isinstance(bin_op, NullOp):
             assert len(terms) == 1
         else:
-            assert reduced_vars and len(terms) > 1
             assert (red_op, bin_op) in DISTRIBUTIVE_OPS
 
         inputs = OrderedDict()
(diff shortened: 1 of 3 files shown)


fritzo:
Tests should pass as soon as https://github.com/pyro-ppl/pyro/pull/2065 merges upstream.

test/test_cnf.py
```diff
@@ -59,3 +64,32 @@ def test_normalize_einsum(equation, plates, backend, einsum_impl):
...
+@pytest.mark.parametrize("x_inputs,y_inputs", [
+    ("", ""),
+    ("i", ""),
+    ("", "i"),
+    ("j", "i"),
+    ("ij", "i"),
+    ("i", "ik"),
+    ("ij", "ik"),
+])
+@pytest.mark.parametrize("red_op,bin_op", [(ops.add, ops.mul), (ops.logaddexp, ops.add)], ids=str)
```
eb8680: Add `(ops.max, ops.add)` for `torch_map` here?

funsor/cnf.py
```diff
@@ -220,14 +225,71 @@ def eager_contraction_to_binary(red_op, bin_op, reduced_vars, lhs, rhs):
     return result
 
 
+@eager.register(Contraction, ops.AddOp, ops.MulOp, frozenset, Tensor, Tensor)
```
eb8680: Add a pattern for `ops.MaxOp`/`ops.AddOp` that uses the `torch_map` backend?  I'm not sure that does anything smarter but it would keep execution behavior more consistent.

test/test_cnf.py
fritzo: Pyro doesn't actually have a (max,add) ring, rather it has an (argmax,add) ring that is tightly coupled to our pyro adjoint algorithms including requirements for `._pyro_dims` attribute. While we could work around this with some effort (effort I could spend elsewhere), the (argmax,add) ring simply adds-then-reduces under the hood, so there is no computational advantage.

eb8680: > Pyro doesn't actually have a (max,add) ring

The forward pass looks to me like [it computes `max`](https://github.com/pyro-ppl/pyro/blob/dev/pyro/ops/einsum/torch_map.py#L37), but as you say there's no computational advantage and the adjoint behavior is different.

PR #282, opened 2019-10-18
Implement sound is_affine() test

fritzo:
Addresses #72 

Recall that our affine pattern matching algorithm consists of three steps:
1. (#245) an optimistic (necessarily incomplete) unsound pattern matching step to match possibly affine funsors; 
2. (this PR) a sound but incomplete `is_affine()` test to eliminate non-affine funsors;
3. (#249) an algorithm to substitute in probe funsors and extract affine coefficients from a funsor known to be affine.

This also removes the obsolete `.is_affine` attribute of `Contraction`.

## Tested

- positive cases are exercised by existing tests
- added new positive check and cases to `test_extract_affine()`
- added a new unit test for negative cases

### DIFF
--- a/funsor/affine.py
+++ b/funsor/affine.py
@@ -1,16 +1,96 @@
 from collections import OrderedDict
+from functools import reduce
 
 import opt_einsum
 import torch
+from multipledispatch import dispatch
 
-from funsor.interpreter import gensym
-from funsor.terms import Lambda, Variable, bint
-from funsor.torch import Tensor
+from funsor.cnf import Contraction
+from funsor.interpreter import gensym, interpretation
+from funsor.terms import Binary, Funsor, Lambda, Reduce, Unary, Variable, bint, reflect
+from funsor.torch import Einsum, Tensor
+
+from . import ops
 
 
-# FIXME change this to a sound but incomplete test using pattern matching.
 def is_affine(fn):
-    return True
+    """
+    A sound but incomplete test to determine whether a funsor is affine with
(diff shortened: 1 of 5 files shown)


fritzo:
cc @fehiepsi in case you are interested.

funsor/affine.py
```diff
@@ -1,16 +1,103 @@
 from collections import OrderedDict
+from functools import reduce
 
 import opt_einsum
 import torch
+from multipledispatch import dispatch
```
eb8680: nit: use `singledispatch` here instead?  I think it's faster

test/test_affine.py
```diff
@@ -116,3 +119,21 @@ def test_extract_affine(expr):
                          for k, (coeff, eqn) in coeffs.items())
     assert isinstance(actual, Tensor)
     assert_close(actual, expected)
+
+
+@pytest.mark.parametrize("expr", [
```
eb8680: Add some test cases covering `Unary`?

fritzo:
> Can you add some positive test cases ... with random substitutions ...

Yes, I forgot to mention positive tests are tested by `test_extract_affine()` via random substitution.

test/test_affine.py
fritzo: done.

funsor/affine.py
fritzo: done.

fritzo:
Thanks for the careful review!

PR #284, opened 2019-10-21
Refactor towards Gaussian._eager_subs_affine()

fritzo:
Addresses #72 

This is step 1/3 in implementing affine recognition for Gaussian funsors, and is a pure refactoring PR. This does not actually implement `._eager_affine_subs()`, and I will implement that in step 2/3 and remove the old MultivariateNormal pattern in step 3/3. This PR:
- exposes the old `_affine_inputs()` function as a public `affine_inputs()` function
- adds momoziation to `affine_inputs()`
- moves the `affine_inputs` pattern for Contraction to cnf.py to allow a new dependency chain
  `cnf.py -> gaussian.py -> affine.py` where affine is used inside gaussian
- refactors `extract_affine()` so that it is always correct but produces an affine representation only wrt those inputs that are actually affine. This is both less dangerous and the behavior we will want in `Gaussian._eager_affine()`
- refactors `Gaussian.eager_subs()` into multiple functions.

## Tested
- pure refactoring is covered by existing tests

### DIFF
--- a/funsor/cnf.py
+++ b/funsor/cnf.py
@@ -9,12 +9,26 @@ from multipledispatch.variadic import Variadic
 from pyro.distributions.util import broadcast_shape
 
 import funsor.ops as ops
+from funsor.affine import affine_inputs
 from funsor.delta import Delta
 from funsor.domains import find_domain
 from funsor.gaussian import Gaussian
-from funsor.interpreter import recursion_reinterpret
+from funsor.interpreter import interpretation, recursion_reinterpret
 from funsor.ops import DISTRIBUTIVE_OPS, AssociativeOp, NullOp, nullop
-from funsor.terms import Align, Binary, Funsor, Number, Reduce, Subs, Unary, Variable, eager, normalize, to_funsor
+from funsor.terms import (
+    Align,
+    Binary,
+    Funsor,
+    Number,
+    Reduce,
+    Subs,
+    Unary,
+    Variable,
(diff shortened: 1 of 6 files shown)


funsor/gaussian.py
```diff
@@ -334,47 +335,56 @@ def eager_subs(self, subs):
...
-            inputs = funsors[0].inputs.copy()
-            inputs.update(real_inputs)
-            int_result = Gaussian(funsors[0].data, funsors[1].data, inputs)
-            new_subs = real_subs + lazy_subs
-            return Subs(int_result, new_subs) if new_subs else int_result
-
+            return self._eager_subs_int(int_subs, real_subs + affine_subs + lazy_subs)
+        if real_subs:
+            return self._eager_subs_real(real_subs, affine_subs + lazy_subs)
+        # TODO
```
fritzo: I will implement this in the next PR. I don't think Du Phan would be interested in this PR, so I'll add the interesting stuff in the follow-up PR.

funsor/cnf.py
```diff
@@ -269,6 +283,13 @@ def eager_contraction_gaussian(red_op, bin_op, reduced_vars, x, y):
     return (x + y).reduce(red_op, reduced_vars)
 
 
+@affine_inputs.register(Contraction)
```
fritzo: moved verbatim from affine.py

PR #297, opened 2020-01-06
Add some ops required for Gaussian

fehiepsi:
This PR adds ops required for #296. I also add corresponding NumPy ops so we can test if Gaussian is actually backend-agnostic there.

I am not sure how to properly test those ops. Probably we can add an additional `backend` arg for pytest.parameterize in `test_gaussian` or other test files?

Currently, tests for `test_numpy.py::test_unary` are failing... which might come from the fact that NumPy array does not have attributes such as `.sigmoid()` like PyTorch. But I have no clue to process further... I appreciate any help to resolve this issue!

### DIFF
--- a/funsor/gaussian.py
+++ b/funsor/gaussian.py
@@ -16,18 +16,18 @@ from funsor.util import lazy_property
 
 
 def _log_det_tri(x):
-    return x.diagonal(dim1=-1, dim2=-2).log().sum(-1)
+    return ops.log(ops.diagonal(x, -1, -2)).sum(-1)
 
 
 def _vv(vec1, vec2):
     """
     Computes the inner product ``< vec1 | vec 2 >``.
     """
-    return vec1.unsqueeze(-2).matmul(vec2.unsqueeze(-1)).squeeze(-1).squeeze(-1)
+    return ops.matmul(ops.unsqueeze(vec1, -2), ops.unsqueeze(vec2, -1)).squeeze(-1).squeeze(-1)
 
 
 def _mv(mat, vec):
-    return torch.matmul(mat, vec.unsqueeze(-1)).squeeze(-1)
+    return ops.matmul(mat, ops.unsqueeze(vec, -1)).squeeze(-1)
 
 
 def _trace_mm(x, y):
@@ -36,7 +36,7 @@ def _trace_mm(x, y):
     """
     assert x.dim() >= 2
     assert y.dim() >= 2
-    return (x * y).sum([-1, -2])
+    return (x * y).sum((-1, -2))
 
 
 def cholesky(u):
@@ -44,18 +44,14 @@ def cholesky(u):
     Like :func:`torch.cholesky` but uses sqrt for scalar matrices.
     Works around https://github.com/pytorch/pytorch/issues/24403 often.
     """
-    if u.size(-1) == 1:
-        return u.sqrt()
-    return u.cholesky()
+    return ops.cholesky(u)
 
 
 def cholesky_inverse(u):
     """
     Like :func:`torch.cholesky_inverse` but supports batching and gradients.
     """
-    if u.dim() == 2:
-        return u.cholesky_inverse()
-    return torch.eye(u.size(-1)).expand(u.size()).cholesky_solve(u)
+    return ops.cholesky_inverse(u)
 
 
 def _compute_offsets(inputs):
--- a/funsor/numpy.py
+++ b/funsor/numpy.py
@@ -306,6 +306,11 @@ def materialize(x):
 ################################################################################
 
 
+@ops.abs.register(np.ndarray)
+def _abs(x):
+    return abs(x)
+
+
 @ops.sigmoid.register(np.ndarray)
 def _sigmoid(x):
     try:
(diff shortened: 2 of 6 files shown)


funsor/numpy.py
```diff
@@ -299,3 +299,151 @@ def materialize(x):
...
+
+@ops.reciprocal.register(np.ndarray)
+def _reciprocal(x):
+    result = np.clip(np.reciprocal(x), a_max=np.finfo(x.dtype).max)
+    return result
+
+
+@ops.safesub.register((int, float), np.ndarray)
+def _safesub(x, y):
+    try:
```
fritzo: 1. It would be best to avoid slow try-except flow in every operation; maybe we can cache these.
2. Let's at least limit the scope of the try-except
```py
try:
    finfo = np.finfo(y.dtype)
except ValueError:
    finfo = np.iinfo(y.dtype)
return x + np.clip(-y, max=finfo.max)
```

funsor/gaussian.py
```diff
@@ -36,26 +36,22 @@ def _trace_mm(x, y):
...
 
 def cholesky(u):
     """
     Like :func:`torch.cholesky` but uses sqrt for scalar matrices.
     Works around https://github.com/pytorch/pytorch/issues/24403 often.
     """
-    if u.size(-1) == 1:
-        return u.sqrt()
-    return u.cholesky()
+    return ops.cholesky(u)
```
fritzo: You can now remove `cholesky` and `cholesky_inverse` from `gaussian.py`

funsor/numpy.py
```diff
@@ -299,3 +299,151 @@ def materialize(x):
...
+# Register Ops
+################################################################################
+
+
+@ops.sigmoid.register(np.ndarray)
+def _sigmoid(x):
+    try:
+        from scipy.special import expit
+        return expit(x)
+    except:
```
fritzo: `except ImportError:`

funsor/ops.py
```diff
@@ -303,6 +303,66 @@ def reciprocal(x):
...
+    raise NotImplementedError
+
+
+@Op
+def cholesky_inverse(x):
+    raise NotImplementedError
+
+
+@Op
+def triangular_solve_op(x, y, upper, transpose):
```
fritzo: It is probably cleaner to encode parameterized ops as classes, as e.g. `GetitemOp` and `ReshapeOp`. How about something like this:
```py
class TriangularSolveMeta(type):
    ...implement caching as in GetitemMeta...

class TriangularSolveOp(Op, metaclass=TriangularSolveMeta):
    def __init__(self, upper, transpose):
        assert isinstance(upper, bool)
        assert isinstance(transpose, bool)
        self.upper = upper
        self.transpose = transpose
    def __call__(self, x, y):
        raise NotImplementedError
```
then later you can, in torch.py
```py
for upper in [False, True]:
    for transpose in [False, True]:
        @TriangularSolveOp(upper, transpose).register(Tensor, Tensor)
        def triangular_solve(x, y):
            return  x.data.triangular_solve(y.data, upper, transpose)
```

test/test_numpy.py
```diff
@@ -219,3 +220,72 @@ def test_align():
...
+def test_unary(symbol, dims):
+    sizes = {'a': 3, 'b': 4}
+    shape = tuple(sizes[d] for d in dims)
+    inputs = OrderedDict((d, bint(sizes[d])) for d in dims)
+    dtype = 'real'
+    data = np.array(np.random.rand(*shape)) + 0.5
+    if symbol == '~':
+        data = data.astype(bool)
+        dtype = 2
+    expected_data = unary_eval(symbol, data)
```
fritzo: You could special case here
```py
if symbol == 'sigmoid':
    expected_data = scipy.special.expit(data)
else:
    expected_data = unary_eval(symbol, data)
```

funsor/gaussian.py
fehiepsi: I'll remove those in #296 PR. For now, it is required for internal math of gaussian and some other files too.

funsor/ops.py
fehiepsi: My intention is to use this op on `torch.Tensor` or `numpy.ndarray`, so it might be better to keep the current pattern as-is.

test/test_numpy.py
fehiepsi: In `unary_eval`, I added `if isinstance(x, np.ndarray): return getattr(funsor.ops, symbol)(x)` so the issue does not happen at this line. :(

fehiepsi:
> For all shape-changing ops like unsqueeze, you will need to add shaping logic to funsor.domains.find_domain().

This is required if we apply those ops to Funsor, but here I think it is enough to apply those ops to `torch.Tensor`/`np.ndarray`. WDYT?

fritzo:
OK, I guess we can omit `find_domain()` logic until we add eager registrations.

funsor/ops.py
fritzo: Ok, but we'll want to change to `TriangularSoveOp` eventually...

fehiepsi:
@fritzo I have marked failing unary tests as xfail to address it later. It is not important for Gaussian refactoring work I think.

PR #301, opened 2020-01-10
Switch to Apache 2.0 license

fritzo:
Addresses https://github.com/pyro-ppl/pyro/issues/2249

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -68,7 +68,7 @@ setup(
         'Intended Audience :: Developers',
         'Intended Audience :: Education',
         'Intended Audience :: Science/Research',
-        'License :: OSI Approved :: MIT License',
+        'License :: OSI Approved :: Apache License 2.0',
         'Operating System :: POSIX :: Linux',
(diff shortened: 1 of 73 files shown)


neerajprad:
Do we need to update the copyright info here too?

fritzo:
I think this is fine since Funsor was started after we joined the Linux Foundation.

setup.py
```diff
@@ -65,7 +68,7 @@
         'Intended Audience :: Developers',
         'Intended Audience :: Education',
         'Intended Audience :: Science/Research',
-        'License :: OSI Approved :: MIT License',
+        'License :: OSI Approved :: Apache License 2.0',
```
neerajprad: I am putting this in a license kwarg, but using `Apache Software License` here as per https://pypi.org/classifiers/, since this might not be recognizable by automated tools. We can have a more specific license here once https://github.com/pypa/pypi-support/issues/142 is addressed.

PR #302, opened 2020-01-13
Add tensor_ops.py

fehiepsi:
The file will contain several utilities such as: `align_tensor`, `align_tensors`, `arange`, `materialize`. In addition, a function Tensor will be constructed to fall back to torch.Tensor or numpy.Array.

Other changes in this PR:
+ use `tensor_ops.is_tensor` to check if a variable is a tensor
+ change `from funsor.torch import Tensor` to `from funsor.torch import TorchTensor`; we will use `from funsor.tensor_ops import Tensor` throughout.

TODO:
- [x] find bugs and make all tests pass

### DIFF
--- a/funsor/__init__.py
+++ b/funsor/__init__.py
@@ -5,8 +5,8 @@ from funsor.domains import Domain, bint, find_domain, reals
 from funsor.integrate import Integrate
 from funsor.interpreter import reinterpret
 from funsor.sum_product import MarkovProduct
-from funsor.terms import Cat, Funsor, Independent, Lambda, Number, Slice, Stack, Variable, of_shape, to_data, to_funsor
 from funsor.tensor_ops import Tensor, arange
+from funsor.terms import Cat, Funsor, Independent, Lambda, Number, Slice, Stack, Variable, of_shape, to_data, to_funsor
 from funsor.util import pretty, quote
 
 from . import (
--- a/funsor/__init__.py
+++ b/funsor/__init__.py
@@ -6,7 +6,7 @@ from funsor.integrate import Integrate
 from funsor.interpreter import reinterpret
 from funsor.sum_product import MarkovProduct
 from funsor.terms import Cat, Funsor, Independent, Lambda, Number, Slice, Stack, Variable, of_shape, to_data, to_funsor
-from funsor.torch import Tensor, arange
+from funsor.tensor_ops import Tensor, arange
 from funsor.util import pretty, quote
 
 from . import (
--- a/funsor/adjoint.py
+++ b/funsor/adjoint.py
@@ -5,6 +5,7 @@ from collections import OrderedDict, defaultdict
 
 import torch
 
+import funsor
 import funsor.interpreter as interpreter
 import funsor.ops as ops
 from funsor.cnf import Contraction, GaussianMixture, nullop
@@ -13,8 +14,8 @@ from funsor.gaussian import Gaussian, align_gaussian
 from funsor.interpreter import interpretation
 from funsor.ops import AssociativeOp
 from funsor.registry import KeyedRegistry
+from funsor.tensor_ops import Tensor, is_tensor, materialize
 from funsor.terms import Binary, Cat, Funsor, Number, Reduce, Slice, Subs, Variable, reflect, substitute, to_funsor
-from funsor.torch import Tensor, materialize
 
 
 def _alpha_unmangle(expr):
@@ -96,7 +97,7 @@ if interpreter._DEBUG:
     adjoint_ops.register = lambda *args: lambda fn: adjoint_ops_register(*args)(interpreter.debug_logged(fn))
 
 
-@adjoint_ops.register(Tensor, AssociativeOp, AssociativeOp, Funsor, torch.Tensor, tuple, object)
+@adjoint_ops.register(funsor.torch.Tensor, AssociativeOp, AssociativeOp, Funsor, torch.Tensor, tuple, object)
 def adjoint_tensor(adj_redop, adj_binop, out_adj, data, inputs, dtype):
     return {}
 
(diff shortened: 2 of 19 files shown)


fehiepsi:
@fritzo I'm pretty unhappy with the new `arange` pattern, which has an additional `prototype` arg. Do you have other suggestions for it?

funsor/tensor_ops.py
```diff
@@ -0,0 +1,149 @@
...
+def _Tensor(x, inputs, dtype):
+    return funsor.torch.Tensor(x, inputs, dtype)
+
+
+@dispatch(np.ndarray, (type(None), tuple, OrderedDict), (int, str))
+def _Tensor(x, inputs, dtype):
+    return funsor.numpy.Array(x, inputs, dtype)
+
+
+def Tensor(x, inputs=None, dtype="real"):
```
fritzo: This is very confusing and should be renamed.

funsor/__init__.py
```diff
@@ -5,8 +5,8 @@
 from funsor.integrate import Integrate
 from funsor.interpreter import reinterpret
 from funsor.sum_product import MarkovProduct
+from funsor.tensor_ops import Tensor, arange
```
fritzo: We definitely want to avoid two distinct concepts named `Tensor`

funsor/adjoint.py
```diff
@@ -13,8 +13,9 @@
 from funsor.interpreter import interpretation
 from funsor.ops import AssociativeOp
 from funsor.registry import KeyedRegistry
+from funsor.tensor_ops import Tensor, is_tensor, materialize
```
fritzo: I think `is_tensor` is a bad idea. Could we either use `isinstance(-, (torch.Tensor, numpy.ndarray))` or define a tuple
```py
# in tensor_ops
numeric_array = (torch.Tensor, numpy.ndarray)
```
to be used the same way we used to use `numbers.Number` back in Python 2?

funsor/distributions.py
```diff
@@ -225,7 +227,7 @@ def __init__(self, total_count, probs, value=None):
         super(Binomial, self).__init__(total_count, probs, value)
 
 
-@eager.register(Binomial, Tensor, Tensor, Tensor)
+@eager.register(Binomial, TorchTensor, TorchTensor, TorchTensor)
```
fritzo: Let's keep this as `torch.Tensor`

fritzo:
Can you explain more the motivation and aim of this PR? I feel that there is so much indirection already in Funsor, that any additional indirection comes at high cost, a cost that often outweighs code duplication. I had expected a PR deduplicating code in `align_tensor`, but if that is the only aim of this PR, then it seems maybe not worth the abstraction overhead.

funsor/tensor_ops.py
fehiepsi: Initially, I wanted to change torch.Tensor to torch.TorchTensor so less code will be needed to change. But I have less opinion on naming, so I would like to incorporate your suggestion. How about NumericArray? And we use NumericArray throughout the repo...

funsor/adjoint.py
fehiepsi: good idea! using `numeric_array` is simpler. I didn't know about this usage pattern. :D

fehiepsi:
@fritzo The topic-comment already covers changes in this PR. I guess your most worrisome stuff is about the duplicated names torch.Tensor and tensor_ops.Tensor? If so, we can just discuss about appropriate names. I don't think it will lead to any indirection at the moment (but maybe it will in the future, when more stuffs come - though I think that it won't cause much trouble for us).

About Tensor, my motivation is to have the same statement
```
Tensor(a, b, ...)
```
to create a Tensor/Array (so I don't have to change Tensor...). Before this PR, in `gaussian.py`, I used `ops.Tensor` for this purpose. However, as you pointed out in the last PR, `ops` is reserved for numeric operators, so I move the implementation to `tensor_ops`. The ultimate goal is to not have "import torch; import funsor.torch" in the main files. As you can observe from this PR, we don't need to use `import funsor.torch` in most of the files. `import torch` is almost not required in `gaussian.py` and I intend to remove it in other files when more tests for `numpy` backend are added in later PRs. Does this align with your intentions?

Anyway, if the issue is just "naming", then as I said, I have less opinion on it (though personally, I would prefer to use the name Tensor for funsor of numeric arrays and to have TorchTensor for the specific implementation with torch.Tensor). Whatever you prefer, I will follow.

fritzo:
@fehiepsi I am still missing your high level goal. Could you instead start a design doc for the proposed refactorings? I am guessing your end goal is to make all of funsor agnostic to either the `numpy` or `torch` backend, and that the nearer term goal is to make at least `Gaussian` agnostic. But I do not understand how this PR fits in, and I am not convinced this PR moves us in the best direction.

There are multiple ways to may funsor backend-agnostic, e.g. we could make `Gaussian` use {`funsor.torch.Tensor`, `funsor.numpy.Array`}, or we could make `Gaussian` use {`torch.Tensor`, `numpy.ndarray`}. It appears that this PR adds abstraction mechanisms for the latter, whereas we already have many abstraction mechanisms for the former. This deserves more detailed discussion.

Because either design solution would require a large-scale refactoring (even if only rename refactoring), it is difficult to discuss these broad changes in a PR. I suggest we close or pause this PR and discuss strategy in a design doc.

fritzo:
As discussed offline, we decided the ideas in this PR move in a good direction, and that we should follow a similar refactoring to combine `funsor.torch.Tensor` and `funsor.numpy.Array`. See "approach 4" in the [design doc](https://docs.google.com/document/d/1HStc-F668gU8PIYfNiv2jHWeYzz5E_7K7w4c-gRb2es/edit).

funsor/adjoint.py
```diff
@@ -225,9 +226,9 @@ def _scatter(src, res, subs):
                 index.append(int(v.data))
             else:
                 # Permute and expand v.data to end up at new_dims.
-                assert isinstance(v, Tensor)
+                assert is_tensor(v)
```
fritzo: As discussed offline, I think it is better to refactor `Tensor+Array` before refactoring `Gaussian`.

PR #303, opened 2020-01-21
make Tensor backend agnostic

fehiepsi:
This PR attempts to make Tensor backend agnostic, which will subsume efforts to write non-numeric code for each backend. In principle, we only require adding dispatches to numeric ops for a new backend.

### Tasks

- [x] fork `funsor.torch.Tensor` to a new class `tensor.Tensor` and change all torch ops to funsor.ops

- [x] fork all eager op on `funsor.torch.Tensor` to `tensor.Tensor`

- [x] add tests for the change

- [x] remove unnecessary code in `torch.py`, `numpy.py`; only keep dispatching implementation on those files

- [x] Fix `test_arange_simple`

- [x] Fix a weird bug of at the assertation `assert type(actual) == type(expected)`, where actual and expected are two `Tensor`s with numpy scalar data.

- [x] Update tensor docs

### Future PRs

- add backend-agnostic tests to `test_gaussian`.
- create `torch` and `numpy` folders that hold implementations of dispatching numeric ops and dispatching distributions.


funsor/tensor.py
```diff
@@ -0,0 +1,973 @@
...
+    Variable,
+    eager,
+    substitute,
+    to_data,
+    to_funsor
+)
+from funsor.util import getargspec, quote
+
+
+numeric_array = (torch.Tensor, np.ndarray, np.generic)
```
fehiepsi: NumPy scalars belong to the class [np.generic](https://docs.scipy.org/doc/numpy/reference/generated/numpy.generic.html) rather than `np.ndarray`, so we need to include them here.

funsor/tensor.py
```diff
@@ -0,0 +1,973 @@
...
+        tensors.append(ops.expand(align_tensor(inputs, part), shape))
+    del inputs[part_name]
+
+    dim = 0
+    tensor = ops.cat(dim, *tensors)
+    inputs = OrderedDict([(name, bint(tensor.shape[dim]))] + list(inputs.items()))
+    return Tensor(tensor, inputs, dtype=output.dtype)
+
+
+def arange(prototype, name, *args, **kwargs):
```
fehiepsi: @fritzo As mentioned in a previous PR, I need `prototype` here to dispatch to `torch.arange` or `np.arange`. IMO, it is better to make `arange` and `materialize` Tensor methods, rather than standalone functions. WDYT?

funsor/tensor.py
```diff
@@ -0,0 +1,973 @@
...
+        raise ValueError
+    if step <= 0:
+        raise ValueError
+    stop = min(dtype, max(start, stop))
+    data = ops.new_arange(prototype, start, stop, step)
+    inputs = OrderedDict([(name, bint(len(data)))])
+    return Tensor(data, inputs, dtype=dtype)
+
+
+def materialize(prototype, x):
```
fehiepsi: ditto: need `prototype` here

funsor/tensor.py
```diff
@@ -0,0 +1,973 @@
...
+        out = ''.join(new_symbols[k] for k in inputs) + out
+        equation = ','.join(ins) + '->' + out
+
+        data = ops.einsum(equation, *[x.data for x in operands])
+        return Tensor(data, inputs)
+
+    return None  # defer to default implementation
+
+
+def numeric_tensordot(x, y, dims):
```
fehiepsi: `torch_tensordot` is renamed to `numeric_tensordot`. Alternatives are welcomed!

funsor/tensor.py
```diff
@@ -0,0 +1,973 @@
...
+    y_start = x_end - dims
+    y_end = y_start + len(y.output.shape)
+    symbols = 'abcdefghijklmnopqrstuvwxyz'
+    equation = '{},{}->{}'.format(symbols[x_start:x_end],
+                                  symbols[y_start:y_end],
+                                  symbols[x_start:y_start] + symbols[x_end:y_end])
+    return Einsum(equation, (x, y))
+
+
+def numeric_stack(parts, dim=0):
```
fehiepsi: ditto: torch_stack -> numeric_stack

funsor/tensor.py
```diff
@@ -0,0 +1,973 @@
...
+        dim = dim - len(shape) - 1
+    assert dim < 0
+    split = dim + len(shape) + 1
+    shape = shape[:split] + (len(parts),) + shape[split:]
+    output = Domain(shape, parts[0].dtype)
+    fn = functools.partial(ops.stack, dim)
+    return Function(fn, output, parts)
+
+
+REDUCE_OP_TO_NUMERIC = {
```
fehiepsi: Because we are making Tensor backend-agnostic, `torch.foo` is replaced by `ops.foo`. We already have `ops.min` and `ops.max`, so I used numpy [amin](https://docs.scipy.org/doc/numpy/reference/generated/numpy.amin.html), `amax` names here.

test/test_numpy.py
```diff
@@ -1,296 +0,0 @@
...
-import funsor
-from funsor import Number, Variable, bint, reals
-from funsor.domains import Domain
-from funsor.interpreter import _USE_TCO, interpretation
-from funsor.numpy import Array, align_arrays
-from funsor.terms import lazy
-from funsor.testing import assert_equiv, check_funsor, random_array
-
-# FIXME rewrite stack-based interpreter to be compatible with unhashable data
-xfail_with_tco = pytest.mark.xfail(
```
fehiepsi: @eb8680 It seems that NumPy tests passed with the Tensor class. Do I need to pay attention to this mark?

test/test_tensor.py
```diff
@@ -134,43 +148,50 @@ def test_slice_1(stop):
...
+    t = randn((10, 2), backend)
     actual = Tensor(t)["i"](i=Slice("j", start, stop, step, dtype=10))
     expected = Tensor(t[start: stop: step])["j"]
     assert_close(actual, expected)
 
 
-def test_arange_simple():
-    t = torch.randn(3, 4, 5)
+@pytest.mark.parametrize("backend", ["torch", "numpy"])
+# FIXME: this test is failing at the second `assert_close`?
```
fehiepsi: I am not sure why this simple test is failing while most of remaining tests passed. :(

funsor/tensor.py
fritzo: Sure, how about naming these methods `.new_arange()` and `.materialize()`?

funsor/tensor.py
```diff
@@ -0,0 +1,977 @@
...
+
+
+class LazyTuple(tuple):
+    def __call__(self, *args, **kwargs):
+        return LazyTuple(x(*args, **kwargs) for x in self)
+
+
+class Function(Funsor):
+    r"""
+    Funsor wrapped by a PyTorch function.
```
fritzo: "PyTorch" -> "native PyTorch or NumPy" here and below

funsor/tensor.py
fritzo: I think simply `tensordot()` should work now? The previous `torch_` prefix was simply to distinguish it from `numpy_tensordot()`.

funsor/tensor.py
fritzo: ditto, I think simply `stack()` or `tensor_stack()` makes sense now. We have a different `Stack()` for the funsor version.

test/examples/test_bart.py
```diff
@@ -32,8 +32,8 @@ def bounded_exp(x, bound):
 call_count = 0
 
 
-@funsor.torch.function(reals(2 * num_origins * num_destins),
-                       (reals(num_origins, num_destins, 2),
+@funsor.tensor.function(reals(2 * num_origins * num_destins),
```
fritzo: ditto: It would be nice to simplify to `@funsor.function`

test/test_distributions.py
```diff
@@ -26,7 +26,7 @@ def test_beta_density(batch_shape, eager):
     batch_dims = ('i', 'j', 'k')[:len(batch_shape)]
     inputs = OrderedDict((k, bint(v)) for k, v in zip(batch_dims, batch_shape))
 
-    @funsor.torch.function(reals(), reals(), reals(), reals())
+    @funsor.tensor.function(reals(), reals(), reals(), reals())
```
fritzo: ditto: `@funsor.function` everywhere in this file :smile:

Makefile
```diff
@@ -20,7 +20,6 @@ test: lint FORCE
 	FUNSOR_DEBUG=1 pytest -v test/test_gaussian.py
 	FUNSOR_USE_TCO=1 pytest -v test/test_terms.py
 	FUNSOR_USE_TCO=1 pytest -v test/test_einsum.py
-	FUNSOR_USE_TCO=1 pytest -v test/test_numpy.py
```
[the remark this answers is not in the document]
eb8680: Yeah, that makes sense after this PR

test/test_numpy.py
eb8680: I believe the failures here were caused by the fact that `numpy.ndarray`s were unhashable and hence couldn't be used as dictionary keys.  I'm surprised it works now, I'd have to dig into why.

fehiepsi: Oops, it doesn't work...

eb8680: In that case you can just leave the xfail, it's not an important issue.

eb8680:
Looks great once @fritzo's nits are addressed!  Further changes can be left to followup PRs and tracked in #207

funsor/tensor.py
```diff
@@ -924,7 +920,7 @@ def numeric_tensordot(x, y, dims):
     return Einsum(equation, (x, y))
 
 
-def numeric_stack(parts, dim=0):
+def stack(parts, dim=0):
```
fritzo: Interesting, I think we could later replace this with an eager evaluation rule for `ops.stack` and eliminate `funsor.tensor.stack` entirely. Similarly I think we'll be able to replace `tensordot` with an `ops.tensordot` + an `eager` rule. But that's out of scope for this PR :smile:

fehiepsi:
@fritzo @eb8680 Thanks for reviewing! I have addressed most of your comments. I have added FUNSOR_USE_TCO=1 task to #1 to address it in a follow-up PR. #1 will contain tasks for this numpy-backend work. Please feel free to adjust it as necessary.

Issue #304, opened 2020-01-24
FR GaussianHMM.rsample() or MarkovProduct(Gaussian).sample()


fritzo:
Addresses https://github.com/pyro-ppl/pyro/issues/2182

A powerful class of guides would be enabled if we could support posterior sampling from a `GaussianHMM` class, with an interface like
```py
class GaussianGMM(TorchDistribution):
    def posterior(self, likelihood):
        assert isinstance(likelihood, MultivariateNormal)
        ...
        return GaussianHMM(...)  # the posterior
```
It should be easy in Funsor to contruct the `GaussianHMM`'s `.funsor_dist` using only a `+` operator. Then all we would need would be a `GaussianHMM.rsample()` method similar to https://github.com/pyro-ppl/pyro/pull/2238, requiring (I believe) a `MarkovProduct.sample()` implementation.

## Tasks
TODO(@eb8680) suggest a work plan

PR #306, opened 2020-01-26
Add ops.sample

fritzo:
Addresses #304 
Pair coded with @eb8680 

This pulls the `ops.sample` changes out of #305 

@eb8680 I hope this does not conflict with any of your work. I'll leave the remaining `Argreduce` and adoint changes to you, and then port Pyro's https://github.com/pyro-ppl/pyro/pull/2285 to Funsor.

## Tested
- added `ops.sample` to some Tensor tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ cache:
 
 install:
     - pip install -U pip
-    - pip install torch==1.3.0+cpu torchvision==0.4.0+cpu -f https://download.pytorch.org/whl/torch_stable.html
+    - pip install torch==1.4.0+cpu torchvision==0.5.0+cpu -f https://download.pytorch.org/whl/torch_stable.html
 
     # Keep track of Pyro dev branch
     - pip install https://github.com/pyro-ppl/pyro/archive/dev.zip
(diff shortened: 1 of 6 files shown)


eb8680:
Looks like the test failure is a bug in dependent type inference for negative numbers that was being ignored via `pytest.xskip` by `test_terms.py::test_reduce_all`.  This failure should really never be triggered, though, since op implementations are meant to be opaque.  One possible fix is to ensure that the things being passed into the actual implementation `_logaddexp` are always actual data types rather than funsors.  This is handled implicitly by our method resolution order for ops that involve Python operator overloading via pattern `funsor.terms.eager_binary_number_number`, but not for ops like `logaddexp`.

fritzo:
@eb8680 as I see it, the failure was due to a missed `pytest.skip()` in `test_reduce_all()` which is needed because that test is written with integer data, and integer data only supports add, mul ops (not logaddexp). I've extended the `.skip()`.

eb8680:
> integer data only supports add, mul ops (not logaddexp)

Ok, we should make that explicit in `find_domain` in a separate PR

Issue #1, opened 2019-02-14
Add funsor.numpy backend


fritzo:
Currently `funsor.torch` provides a PyTorch backend via `Tensor`, `@function` and a few other helpers. It should be straight-forward to provide similar adaptors for numy and jax., say as a new module `funsor.numpy` and class `Array`.

## Outline

- [x] #74 support numpy in `pyro.ops`. e.g. we will need to call `np.log(x)` rather than `x.log()` based on type.
- [x] implement an `align_arrays` function similar to `align_tensors`
- [x] implement a `funsor.numpy.Array` class
- [x] register `np.array` with `to_funsor`
- [x] fork `test_torch.py` to a new `test_numpy.py`
- [x] Add compatibility with tail-call optimized interpreter from #145 and remove conditional xfail added to `test_numpy.py` added in #155 
- [x] implement `Array.eager_reduce()`
- [x] optionally add a `Function` wrapper and `@funsor.numpy.function` decorator
- [x] refactor (`Tensor`,`Array`) to share a superclass, e.g. `Tensor` -> {`TorchTensor`, `NumpyTensor`)
- [x] refactor `Gaussian` to `Tensor` rather than torch.Tensor internally.
    This requires wrapping many linear algebra operations as functions, e.g. `cholesky_solve`.
- [x] Make unary ops work with Array input. (#303)
- [ ] Enable `test_tensor.py` with FUNSOR_USE_TCO=1
- [x] Support numpy docstring in `funsor.util.getargspec`
- [x] Add numpy backend tests for `test_gaussian`.
- [x] Add numpy backend for `einsum`.
- [ ] Consider disabling some of the overhead caused by `normalize`.
- [x] Make `funsor.distributions` backend-agnostic
- [x] Add support for `jax.numpy`.
- [ ] update `funsor.minipyro` to support numpy backend
- [x] add inference examples using numpy

eb8680:
> Enable test_tensor.py with FUNSOR_USE_TCO=1

@fehiepsi this should currently work with PyTorch but not NumPy.  Unless there was a regression in #303 and the PyTorch version is failing too, there's no need to prioritize this.

eb8680:
I'm going to close this as resolved. We can track progress on `normalize` in #411.

PR #311, opened 2020-01-30
Add numpy backend tests for Gaussian

fehiepsi:
Follow up  #303, this PR adds numpy backend tests for gaussian.

### Tasks

- [x] fix eager_subs_affine test; currently, the numpy test failed due to torch tensor is created during the process...
- [x] fix integrate tests


funsor/affine.py
```diff
@@ -135,10 +134,12 @@ def extract_affine(fn):
         ``(coefficient, eqn)`` pair in einsum form.
     :rtype: tuple
     """
+    # FIXME: does this make sense?
+    prototype = fn.terms[1].data
```
fehiepsi: @fritzo I need prototype value for the below `ops.new_zeros`. This line passes all gaussian tests but does not work with general `fn` (of course). Do you have an idea to resolve this issue? It seems to me that `extract_affine` is only used in Gaussian. Should we make it a method of gaussian (to use backend information from `info_vec` or `precision`)?

funsor/cnf.py
```diff
@@ -231,14 +231,14 @@ def eager_contraction_to_binary(red_op, bin_op, reduced_vars, lhs, rhs):
 def eager_contraction_tensor(red_op, bin_op, reduced_vars, *terms):
     if not all(term.dtype == "real" for term in terms):
         raise NotImplementedError('TODO')
-    return _eager_contract_tensors(reduced_vars, terms, backend="torch")
+    return _eager_contract_tensors(reduced_vars, terms, backend=terms[0].backend)
```
fehiepsi: `Tensor.backend` is a string of `"torch"` or `"numpy"`.

funsor/cnf.py
```diff
@@ -231,14 +231,14 @@ def eager_contraction_to_binary(red_op, bin_op, reduced_vars, lhs, rhs):
...
 
 
 @eager.register(Contraction, ops.LogAddExpOp, ops.AddOp, frozenset, Tensor, Tensor)
 def eager_contraction_tensor(red_op, bin_op, reduced_vars, *terms):
     if not all(term.dtype == "real" for term in terms):
         raise NotImplementedError('TODO')
-    return _eager_contract_tensors(reduced_vars, terms, backend="pyro.ops.einsum.torch_log")
+    return _eager_contract_tensors(reduced_vars, terms, backend=terms[0].logsumexp_backend)
```
fehiepsi: Tensor.logsumexp_backend is either `pyro.ops.einsum.torch_log` or `funsor.numpy`

funsor/numpy.py
```diff
@@ -7,6 +7,45 @@
 from funsor.util import quote
 
 
+def einsum(equation, *operands):
```
fehiepsi: adapted from Pyro [torch_log](https://github.com/pyro-ppl/pyro/blob/dev/pyro/ops/einsum/torch_log.py)

funsor/testing.py
```diff
@@ -241,18 +241,18 @@ def random_tensor(inputs, output=reals(), backend="torch"):
...
     batch_shape = tuple(d.dtype for d in inputs.values() if d.dtype != 'real')
     event_shape = (sum(d.num_elements for d in inputs.values() if d.dtype == 'real'),)
-    prec_sqrt = torch.randn(batch_shape + event_shape + event_shape)
-    precision = torch.matmul(prec_sqrt, prec_sqrt.transpose(-1, -2))
-    precision = precision + 0.05 * torch.eye(event_shape[0])
-    loc = torch.randn(batch_shape + event_shape)
-    info_vec = precision.matmul(loc.unsqueeze(-1)).squeeze(-1)
+    prec_sqrt = randn(batch_shape + event_shape + event_shape, backend)
+    precision = ops.matmul(prec_sqrt, ops.transpose(prec_sqrt, -1, -2))
+    precision = precision + 0.5 * ops.new_eye(precision, event_shape[:1])
```
fehiepsi: Here I change `0.05` to `0.5` to pass numpy tests. With `0.05`, even with torch backend, gaussian integrate tests will fail with some random seed.

funsor/affine.py
fehiepsi: A simple solution is to introduce an arg `backend` here, and use Tensor(...).backend to dispatch.

funsor/cnf.py
fritzo: I think it is better to put this in a new file funsor/einsum/numpy_log.py. Note the `opt_einsum` requirements are a little subtle: they use a `tensordot`, an `einsum` and a `transpose` implementation. Since you've already implemented a logsumexp `einsum`, you can create a `tensordot` via
```py
from pyro.ops.einsum.util import Tensordot
tensordot = Tensordot(einsum)
```

funsor/numpy.py
fritzo: As I mentioned above, this should go in a separate file like funsor/einsum/numpy_log.py, in particular because we might want to add a linear (sum-product) or a tropical (max,sum) version in the future, and we'll want to set e.g. `backend=funsor.einsum.numpy_tropical` for those cases. Note these are not user-facing; they are only consumed by `opt_einsum`.

funsor/numpy.py
```diff
@@ -7,6 +7,45 @@
...
+    """
+    Log-sum-exp implementation of einsum.
+    """
+    # rename symbols to support PyTorch 0.4.1 and earlier,
```
fritzo: nit: Does this still apply to NumPy?

funsor/pyro/convert.py
```diff
@@ -82,12 +82,12 @@ def tensor_to_funsor(tensor, event_inputs=(), event_output=0, dtype="real"):
 
 def funsor_to_tensor(funsor_, ndims, event_inputs=()):
     """
-    Convert a :class:`funsor.torch.Tensor` to a :class:`torch.Tensor` .
+    Convert a :class:`funsor.tensor.Tensor` to a :class:`torch.Tensor` .
```
fritzo: I guess we'll eventually want to use this same helper for `np.array`, so we can convert to a raw `numeric_array` of underlying type?

funsor/tensor.py
```diff
@@ -381,6 +381,20 @@ def materialize(self, x):
...
 
+    @property
+    def backend(self):
+        if torch.is_tensor(self.data):
+            return "torch"
+        else:
+            return "numpy"
+
+    @property
+    def logsumexp_backend(self):
```
fritzo: nit: I think this is better as a global dict accessible to `_eager_contract_tensors()`:
```py
# in funsor/cnf.py
BACKEND_TO_LOGSUMEXP_BACKEND = {
    "torch": "pyro.ops.einsum.torch_log",
    "numpy": "funsor.einsum.numpy_log",
}

@eager.register(Contraction, ops.LogAddExpOp, ops.AddOp, frozenset, Tensor, Tensor)
def eager_contraction_tensor(red_op, bin_op, reduced_vars, *terms):
    if not all(term.dtype == "real" for term in terms):
        raise NotImplementedError('TODO')
    backend = BACKEND_TO_LOGSUMEXP_BACKEND[terms[0].backend]
    return _eager_contract_tensors(reduced_vars, terms, backend=backend)
```

funsor/affine.py
fritzo: This seems reasonable, and preferable to a global option.

funsor/pyro/convert.py
fehiepsi: Yes, I think we should move these utilities to e.g. `tensor.py` file. I would like to defer this to the relevant PR (where we can discuss which file those utilities should be in).

funsor/affine.py
```diff
@@ -135,10 +135,12 @@ def extract_affine(fn):
         ``(coefficient, eqn)`` pair in einsum form.
     :rtype: tuple
     """
+    # FIXME: does this make sense?
+    prototype = prototype if prototype is not None else torch.tensor([])
```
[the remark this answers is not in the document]
fehiepsi: I think having a global setting is a good idea. Currently, we don't need a global setting for remaining dispatching code (except `arange` and `materialize`, which I have converted them to Tensor methods), so I made a temporary implementation to only use this global setting in Gaussian affine subs.

funsor/numpy.py
fehiepsi: It applies to NumPy, but in JAX we don't need this.

funsor/tensor.py
fehiepsi: Thanks!

funsor/einsum/numpy_log.py
```diff
@@ -0,0 +1,40 @@
+import numpy as np
```
fritzo: I believe [opt_einsum requires](https://optimized-einsum.readthedocs.io/en/stable/backends.html) also `tensordot` and `transpose` implementations, which you can get via
```py
from pyro.ops.einsum.util import Tensordot

tensordot = Tensordot(einsum)
transpose = np.transpose
```
These may not be exercised in existing tests, but I believe they will be required for more complex contractions in e.g. NumPyro's enumeration machinery.

fehiepsi:
Yes, I have made a separate issue for `test_einsum` at https://github.com/pyro-ppl/funsor/issues/1#issue-410457128. Thanks for your suggestion!!

PR #313, opened 2020-02-07
Switch numpy op to jax op

fehiepsi:
While testing numpy backend for einsum, I realize that the [einsum_categorical](https://github.com/pyro-ppl/funsor/blob/master/test/test_einsum.py#L65) test requires us to have wrappers for numpyro distribution first. This PR changes `numpy.foo` to `jax.numpy.foo`.

- [x] Currently, I got the hashing issue "TypeError: JAX DeviceArray, like numpy.ndarray, is not hashable." This is a bit strange because I didn't face that hashing issue in numpy. ~Debugging...~ The issue is Device Array not hashable but isinstance(device_array, Hashable) is True.
- [x] ~Add ops.index_update because JAX does not allow in-place modifying data.~ no need
- [x] Fix matmul test


funsor/numpy.py
```diff
@@ -11,17 +17,11 @@
...
 
-try:
-    from scipy.special import expit as _sigmoid
-except ImportError:
-    def _sigmoid(x):
-        return 1 / (1 + np.exp(-x))
-
 # take care of scalar numpy objects
-array = (np.ndarray, np.generic)
+array = (onp.ndarray, onp.generic, UnshapedArray, DeviceArray)
```
fehiepsi: I would like to defer tests for UnshapedArray to a later PR (if there is some problems when applying `jax.jit`).

funsor/numpy.py
```diff
@@ -125,36 +131,60 @@ def _cholesky_inverse(x):
...
-    return ans.reshape(batch_shape + ans.shape[-2:])
+    assert np.ndim(x) >= 2 and np.ndim(y) >= 2
+    n, m = x.shape[-2:]
+    assert y.shape[-2:] == (n, n)
+    # NB: JAX requires x and y have the same batch_shape
+    batch_shape = lax.broadcast_shapes(x.shape[:-2], y.shape[:-2])
+    x = np.broadcast_to(x, batch_shape + (n, m))
+
+    # The following procedure handles the case: y.shape = (i, 1, n, n), x.shape = (..., i, j, n, m)
+    # because we don't want to broadcast y to the shape (i, j, n, n).
```
fehiepsi: This is to save computation. In principle, with A.shape = (3, 3), b.shape=(100, 3, 1), solving a single linear equation `Ax = b.swapaxis(0, 1).reshape(3, 100)` would be much faster than solving 100 linear equations `Ax_i = b[i]` for `i in range(100)`.

fehiepsi: I'll move this logic upstream.

funsor/tensor.py
```diff
@@ -70,8 +72,11 @@ class TensorMeta(FunsorMeta):
     Wrapper to fill in default args and convert between OrderedDict and tuple.
     """
     def __call__(cls, data, inputs=None, dtype="real"):
+        # FIXME: how to make the following assertation pass?
+        #   type(Tensor(np.ones(3))) is type(Tensor(onp.ones(3)))
```
fehiepsi: @fritzo Do you have a solution for this issue? This is not important I guess...

eb8680: @fehiepsi why do you need this to be true when `type(np.ones(3)) is type(onp.ones(3))` is false?

fehiepsi: @eb8680 This is just to cleanup the assert_close test. :D

eb8680: To clarify, types in Funsor work a little differently than in regular Python to support pattern matching - `type(Tensor(np.ones(3)))` [evaluates](https://github.com/pyro-ppl/funsor/blob/master/funsor/terms.py#L222) via [`reflect`](https://github.com/pyro-ppl/funsor/blob/master/funsor/terms.py#L76) `Tensor[type(np.ones(3))]`, not `Tensor`.  If you wanted to change this behavior for `Tensor`, you'd need to write a custom `TensorMeta.__getitem__` that made your assertion true by construction:
```py
def __getitem___(cls, arg_types):
    arg_types = (arg_types,) if not isinstance(arg_types, tuple) else arg_types
    arg_types = (GENERIC_TENSOR_TYPE,) + arg_types[1:]  # e.g. onp.generic
    return super().__getitem__(cls, arg_types)
```
However, it's not clear to me why the above should be true - I don't think `Tensor` should be converting data between different backends or asserting that they're the same.

fehiepsi: Understood. I just remove those comments and the cast from onp.generic to onp.ndarray too. Thanks for explaining!

funsor/interpreter.py
```diff
@@ -10,6 +10,7 @@
 from contextlib import contextmanager
 from functools import singledispatch
 
+import jax
```
fritzo: I see this now forces Funsor users to install both PyTorch and JAX. How difficult would it be to make those requirements optional (as many users would have either PyTorch or JAX but not both)?

funsor/testing.py
```diff
@@ -72,7 +75,15 @@ def allclose(a, b, rtol=1e-05, atol=1e-08):
...
-    assert type(actual) == type(expected), msg
+    if isinstance(actual, array):
+        assert isinstance(expected, array), msg
+    elif isinstance(actual, Tensor) and isinstance(actual.data, array):
+        assert isinstance(expected, Tensor) and isinstance(expected.data, array)
+    elif isinstance(actual, Contraction) and isinstance(actual.terms[0], Tensor) \
+            and isinstance(actual.terms[0].data, array):
+        assert isinstance(expected, Contraction) and isinstance(expected.terms[0].data, array)
+    else:
+        assert type(actual) == type(expected), msg
```
fritzo: nit: a newline after this final `else` clause would help separate the various `if-elif-sections`

setup.py
```diff
@@ -42,6 +42,12 @@
         'torch>=1.3.0',
     ],
     extras_require={
+        'numpy': [
```
fritzo: IIUC this is not really an `extras` require because jax is imported in core funsor modules. I think for now it's fine to move this to a core dependency, but I'd like to make Funsor not rely on having all libraries installed. In particular, I've had some issues installing jax in the past, and I'd like to make sure we can develop Funsor with only one of JAX/PyTorch installed.

fehiepsi: Good point! I think we should have a simple solution here, let me think more about it...

funsor/terms.py
```diff
@@ -65,7 +66,9 @@ def reflect(cls, *args, **kwargs):
         assert len(new_args) == len(cls._ast_fields)
         _, args = args, new_args
 
-    cache_key = tuple(id(arg) if not isinstance(arg, Hashable) else arg for arg in args)
+    # JAX DeviceArray has .__hash__ method but raise the unhashable error there.
```
neerajprad: I think the issue is that `__hash__` is defined but raises a `TypeError`. Instead if it set `__hash__` to `None`, Python would raise an appropriate `TypeError`, we should probably fix this in JAX.

fehiepsi: @neerajprad @eb8680 In JAX, they consider DeviceArray, ndarray as [Hashably](https://github.com/google/jax/blob/master/jax/util.py#L186), which [seems to do](https://github.com/google/jax/blob/master/jax/api.py#L1393) what we do here. I think having `__hash__` returns `TypeError` is their intention.

neerajprad: @fehiepsi - This isn't a big deal, it just might be confusing to new users, and I think the pythonic way would be to just set `__hash__ = None`, which will throw a TypeError by default (I don't think the error message is particularly informative that we need to override it). Quoting from the [python docs](https://docs.python.org/3/reference/datamodel.html):

> A class which defines its own __hash__() that explicitly raises a TypeError would be incorrectly identified as hashable by an isinstance(obj, collections.abc.Hashable) call.

Anyways, this is a really minor point, and we can bring it up with the jax devs if needed later.

setup.py
fehiepsi: @fritzo It seems that [pypi does not support default extra_requires](https://github.com/pypa/setuptools/issues/1139) (I intended to make `pip install funsor` install `torch` by default and `pip install funsor[numpy]` skip `torch` and install `jax`). So my current solution is `pip install funsor[numpy]` will install both `torch` and `jax`. I hope it is fine for now.

eb8680:
> is it ok with you if we add JAX as a required dependency (at least temporarily), or would you prefer to add optional logic in this PR?

Either is fine.  I think in the future we'd ideally have a default numpy backend (i.e. regular numpy, not JAX, and only supporting a subset of ops), and make the other backends optional installs.

fehiepsi:
@eb8680 @neerajprad @fritzo Thanks for your reviewing and suggestions!! I just made some changes to make sure that if `jax` or `scipy` is not available, all tests still pass with the `numpy` backend. This is pretty easy for now, but things might get complicated when we start to use `numpyro` with a lot of numpyro import (I guess pyro-api will help here). I leave it for some future PRs to not block our development.

setup.py
fritzo: This looks fine for now, thanks for getting both numpy+jax backends to work!

PR #314, opened 2020-02-07
add numpy einsum backend

fehiepsi:
This PR ports pyro einsum MAP backend to funsor. Here are changes in this PR:

+ Make `funsor.einsum.numpy_log`, `funsor.einsum.numpy_map` backend-agnostic.
+ Address comments by @fritzo  in the previous PR. However, I still keep `set_backend(get_backend())` in `funsor.__init__` file due to a circular import issue (I'll revisit this issue later).

There are some issues with numpy backend:

+ We can't set attributes `._pyro_dims` (and `._backward` stuff) to ndarray/DeviceArray.
+ `memoize` tests fail with `numpy.generic` data. As a workaround, I cast `numpy.generic` to `numpy.ndarray` at Tensor construction. Because this issue does not happen with JAX backend, I didn't attempt to resolve it.


fritzo:
This should be unblocked now, right?

funsor/einsum/numpy_map.py
```diff
@@ -0,0 +1,55 @@
...
+    Forward-max-sum backward-argmax implementation of einsum.
+    This assumes all operands have a ``._pyro_dims`` attribute set.
+    """
+    equation = packed.rename_equation(equation, *operands)
+    inputs, output = equation.split('->')
+    any_requires_backward = any(hasattr(x, '_pyro_backward') for x in operands)
+
+    contract_dims = ''.join(sorted(set().union(*(x._pyro_dims for x in operands)) - set(output)))
+    dims = output + contract_dims
+    result = reduce(operator.add, packed.broadcast_all(*operands, dims=dims))
```
eb8680: It looks like line 29 (`packed.rename_equation`) and line 35 (`packed.broadcast_all`) are the only places in this code where `ops.packed` is used, at least outside of the adjoint methods.  I suggest copying those two functions into this file or `util.py` and deleteting all of the adjoint-related code in this PR, including `adjoint.py`, `packed.py`, all of the `Backward` implementations and any `_pyro_backward` properties.  Don't worry about missing adjoint functionality for now; just delete or xfail any tests as necessary.

fehiepsi: Thanks! I'll add comments for missing points. :)

funsor/einsum/util.py
```diff
@@ -56,3 +60,48 @@ def __call__(self, x, y, axes=2):
...
+        assert set(dims) == set(sizes)
+    shape = tuple(sizes[dim] for dim in dims)
+    values = list(values)
+    for i, x in enumerate(values):
+        old_dims = x._pyro_dims
+        if old_dims != dims:
+            x = ops.permute(x, tuple(old_dims.index(dim) for dim in dims if dim in old_dims))
+            x = x.reshape(tuple(sizes[dim] if dim in old_dims else 1 for dim in dims))
+            x = ops.expand(x, shape)
+            # workaround: ndarray does not allow setting attribute "_pyro_dims"
```
fehiepsi: I think this is an important point: we can't set new attributes to ndarray, DeviceArray directly.

funsor/einsum/numpy_map.py
```diff
@@ -0,0 +1,38 @@
...
+import funsor.ops as ops
+from funsor.einsum.util import Tensordot, broadcast_all, rename_equation
+
+
+def einsum(equation, *operands):
+    """
+    Forward-max-sum backward-argmax implementation of einsum.
+    This assumes all operands have a ``._pyro_dims`` attribute set.
+    """
+    # TODO: the implementation skips "backward" logic
```
eb8680: The "backward" logic will be handled by `funsor.adjoint`, so you can remove this TODO.

funsor/einsum/numpy_map.py
```diff
@@ -0,0 +1,38 @@
...
+    equation = rename_equation(equation, *operands)
```
eb8680: Is this actually necessary now?  What happens if you remove this line?

funsor/einsum/numpy_map.py
```diff
@@ -0,0 +1,38 @@
...
+    inputs, output = equation.split('->')
+
+    contract_dims = ''.join(sorted(set().union(*(x._pyro_dims for x in operands)) - set(output)))
+    dims = output + contract_dims
+    result = reduce(operator.add, broadcast_all(*operands, dims=dims))
```
eb8680: Is this the only place where `_pyro_dims` are actually used and not just passed along?

fritzo:
@eb8680 has additional comments, please don't merge yet

funsor/einsum/numpy_map.py
fehiepsi: Sorry I missed your comments. I'll address those soon.

fehiepsi: Yes, it is necessary if we keep `_pyro_dims` attribute. One of the `test_einsum` fail: `test_einsum[funsor.einsum.numpy_map-ab,bc,cd->da]`.

funsor/einsum/numpy_map.py
fehiepsi: This and `rename_equation`, `broadcast_all` utility require `pyro_dims`. I am trying to remove `pyro_dims`. If things pass, then I think we can simplify much code here.

fehiepsi:
@eb8680 To remove `pyro_dims`, I intend to remove `rename_equation` and reimplement `broadcast_all` so that it takes args as `values`, names of dimensions of each value, and the output `dims`, then `permute`, `reshape`, and possibly `expand` as in current implementation. Does it sound reasonable to you?

eb8680:
@fehiepsi sure, that sounds easy enough

fehiepsi:
@eb8680 It seems that it is actually easy. Thanks for guiding!

PR #316, opened 2020-02-12
Add optional input arguments to to_funsor and to_data

eb8680:
## Overview

Based on our design discussion on Monday, this PR is a first step toward replacing much of Pyro's internals with Funsor and supporting enumeration in NumPyro.  It adds a `dim_to_name` mapping argument to `to_funsor` and a `name_to_dim` mapping argument to `to_data` which provide sufficient information for uniquely converting PyTorch tensors and distributions to Funsors with free variables and vice versa.  

The changes in this PR only support `Tensor`s.  In followup PRs, I will implement versions of `to_funsor` and `to_data` that apply to `Distribution`s.

This code is ready for review, but before changing the design too much I'd like to have first versions of the plate/enum-aware wrappers (discussed below) implemented as well (update: https://github.com/pyro-ppl/pyro/pull/2307)

## Design

`dim_to_name` is a dictionary that maps batch dimensions to (name, domain) pairs.  `name_to_dim` is a dictionary that maps input names to batch dimensions.  

This extra information (beyond an ordered list of names) is necessary for unique conversion because in Pyro we often have empty batch dimensions in our unpacked tensor shapes that do not mean anything, and eventually in Funsor we will support non-vector `bint` domains.

These arguments are optional, and `to_funsor` and `to_data` behave the same as before when they are not provided.

## Rationale 

The primary use case for this new functionality is upstream in Pyro and NumPyro, where we would like to convert distributions and lazy sample values to and from funsors with shapes consistent with the global plate and enumeration dimension state.  This code could plausibly live in Pyro, but I've put it here since it is intended to support an implementation of `EnumMessenger` and tensor variable elimination in NumPyro.

I am implementing wrappers of these two functions in Pyro that construct `name_to_dim`/`dim_to_name` automatically using the information in `MarkovMessenger`, and I do not expect that Pyro users would ever construct these manually or even know that they exist.

This design is very similar to the `dim_to_symbol` and `symbol_to_dim` arguments used in `pyro.ops.packed.pack` and `pyro.ops.packed.unpack`, but generalized to support event dimensions.  As a consequence of this similarity, we should easily be able to replace `pyro.ops.packed` with Funsor.

## Other changes

This design is also very similar to the conversion functions `funsor_to_tensor` and `tensor_to_funsor` in `pyro/ops/convert.py`, which use a particular convention for constructing instances of `dim_to_name` and `name_to_dim`.  I have modified the implementations of these functions to use `to_data` and `to_funsor` to reflect this similarity and to reuse their tests, and eventually I expect that much of `funsor.pyro.convert` will be deprecated in favor of the plate/enum-aware wrappers being implemented in Pyro.

I have also switched `to_funsor` to use `functools.singledispatch` rather than `multipledispatch`, which may provide a small performance boost and also avoids introducing a `multipledispatch` dependency in the event that we move this code out of Funsor.

## Tested
- New design exercised by existing tests in `test/pyro/test_convert.py`
- Ground `to_funsor` exercised by existing tests

### DIFF
--- a/funsor/pyro/convert.py
+++ b/funsor/pyro/convert.py
@@ -28,12 +28,12 @@ from pyro.distributions.util import broadcast_shape
 from funsor.cnf import Contraction
 from funsor.delta import Delta
 from funsor.distributions import BernoulliLogits, MultivariateNormal, Normal
-from funsor.domains import bint, reals
+from funsor.domains import Domain, bint, reals
 from funsor.gaussian import Gaussian
 from funsor.interpreter import gensym
 from funsor.ops import cholesky
 from funsor.tensor import Tensor, align_tensors
-from funsor.terms import Funsor, Independent, Variable, eager
+from funsor.terms import Funsor, Independent, Variable, eager, to_data, to_funsor
 
 # Conversion functions use fixed names for Pyro batch dims, but
 # accept an event_inputs tuple for custom event dim names.
@@ -63,21 +63,12 @@ def tensor_to_funsor(tensor, event_inputs=(), event_output=0, dtype="real"):
     assert isinstance(event_output, int) and event_output >= 0
     inputs_shape = tensor.shape[:tensor.dim() - event_output]
     output_shape = tensor.shape[tensor.dim() - event_output:]
-    dim_to_name = DIM_TO_NAME + event_inputs if event_inputs else DIM_TO_NAME
-
-    # Squeeze shape of inputs.
-    inputs = OrderedDict()
-    squeezed_shape = []
-    for dim, size in enumerate(inputs_shape):
-        if size > 1:
-            name = dim_to_name[dim - len(inputs_shape)]
-            inputs[name] = bint(size)
-            squeezed_shape.append(size)
-    squeezed_shape = torch.Size(squeezed_shape)
-    if squeezed_shape != inputs_shape:
-        tensor = tensor.reshape(squeezed_shape + output_shape)
-
-    return Tensor(tensor, inputs, dtype)
+    dim_to_name_list = DIM_TO_NAME + event_inputs if event_inputs else DIM_TO_NAME
+    dim_to_name = OrderedDict(zip(
+        range(-len(inputs_shape), 0),
+        zip(dim_to_name_list[len(dim_to_name_list) - len(inputs_shape):],
+            map(bint, inputs_shape))))
+    return to_funsor(tensor, Domain(dtype=dtype, shape=output_shape), dim_to_name)
 
 
 def funsor_to_tensor(funsor_, ndims, event_inputs=()):
@@ -99,16 +90,8 @@ def funsor_to_tensor(funsor_, ndims, event_inputs=()):
     if event_inputs:
         dim_to_name = DIM_TO_NAME + event_inputs
         name_to_dim = dict(zip(dim_to_name, range(-len(dim_to_name), 0)))
-    names = tuple(sorted(funsor_.inputs, key=name_to_dim.__getitem__))
-    tensor = funsor_.align(names).data
-    if names:
-        # Unsqueeze shape of inputs.
-        dims = list(map(name_to_dim.__getitem__, names))
-        inputs_shape = [1] * (-dims[0])
-        for dim, size in zip(dims, tensor.shape):
-            inputs_shape[dim] = size
-        inputs_shape = torch.Size(inputs_shape)
-        tensor = tensor.reshape(inputs_shape + funsor_.output.shape)
+    tensor = to_data(funsor_, name_to_dim)
+
     if ndims != tensor.dim():
         tensor = tensor.reshape((1,) * (ndims - tensor.dim()) + tensor.shape)
     assert tensor.dim() == ndims
--- a/funsor/pyro/convert.py
+++ b/funsor/pyro/convert.py
@@ -66,8 +66,7 @@ def tensor_to_funsor(tensor, event_inputs=(), event_output=0, dtype="real"):
     dim_to_name_list = DIM_TO_NAME + event_inputs if event_inputs else DIM_TO_NAME
     dim_to_name = OrderedDict(zip(
         range(-len(inputs_shape), 0),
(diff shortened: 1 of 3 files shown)


funsor/tensor.py
```diff
@@ -405,18 +405,40 @@ def backend(self):
...
+            name, domain = dim_to_name[dim + len(output.shape) - len(x.shape)]
+            packed_inputs[name] = domain if domain.dtype > 1 else bint(size)
+        if any(size > 1 for size in output.shape):
+            # pack outputs into a single dimension
+            x = x.reshape(x.shape[:-len(output.shape)] + (-1,))
+        x = x.squeeze()
+        if all(size == 1 for size in output.shape):
+            # handle special case: all output dims are 1
+            x = x.unsqueeze(-1)
+        x = x.reshape(x.shape[:-1] + output.shape)
```
fritzo: Can you simplify these 8 lines to
```py
shape = tuple(d.size for k, d in packed_inputs.values()) + output.shape
if x.shape != shape:
    x = x.reshape(shape)
```

funsor/tensor.py
```diff
@@ -405,18 +405,40 @@ def backend(self):
...
+    if not dim_to_name:
+        output = output if output is not None else reals(*x.shape)
+        result = Tensor(x, dtype=output.dtype)
+        if result.output != output:
+            raise ValueError("Invalid shape: expected {}, actual {}"
+                             .format(output.shape, result.output.shape))
+        return result
+    else:
+        assert output is not None  # TODO attempt to infer output
+        assert all(isinstance(k, int) and isinstance(v[0], str) and isinstance(v[1], Domain)
```
fritzo: nit: could you add either `and k >= 0` or `and k < 0` to remind readers whether dims are negative?

funsor/tensor.py
```diff
@@ -484,10 +506,26 @@ def align_tensors(*args, **kwargs):
...
-def _to_data_tensor(x):
-    if x.inputs:
-        raise ValueError(f"cannot convert Tensor to data due to lazy inputs: {set(x.inputs)}")
-    return x.data
+def _to_data_tensor(x, name_to_dim=None):
+    if not name_to_dim or not x.inputs:
+        if x.inputs:
+            raise ValueError(f"cannot convert Tensor to data due to lazy inputs: {set(x.inputs)}")
+        return x.data
+    else:
```
fritzo: nit: could you add an assertion about `name_to_dim` similar to the assertion above in `tensor_to_funsor()`?

funsor/tensor.py
```diff
@@ -405,18 +405,40 @@ def backend(self):
...
+                   for k, v in dim_to_name.items())
+        # logic very similar to pyro.ops.packed.pack
+        # this should not touch memory, only reshape
+        # pack the tensor according to the dim => (name, domain) mapping in inputs
+        packed_inputs = OrderedDict()
+        for dim, size in zip(range(len(x.shape) - len(output.shape)), x.shape):
+            if size == 1:
+                continue  # TODO broadcast domain and shape here
+            name, domain = dim_to_name[dim + len(output.shape) - len(x.shape)]
+            packed_inputs[name] = domain if domain.dtype > 1 else bint(size)
```
fritzo: I would expect this to infer size from the underlying array:
```py
if size == 1:
    continue
name = dim_to_name[dim + len(output.shape) - len(x.shape)]
packed_inputs[name] = bint(size)
```
Can you explain the TODO and why `dim_to_name` appears to have not only a name but also a domain?

funsor/terms.py
```diff
@@ -756,51 +756,40 @@ def _(arg, indent, out):
...
-def to_funsor(x):
+@singledispatch
+def to_funsor(x, output=None, dim_to_name=None):
     """
     Convert to a :class:`Funsor` .
     Only :class:`Funsor` s and scalars are accepted.
 
     :param x: An object.
     :param funsor.domains.Domain output: An optional output hint.
+    :param OrderedDict dim_to_name: An optional inputs hint.
```
fritzo: Could you add a description of the keys and values? In particular:
1. are dims negative or nonnegative or either?
2. are names `str` (as I would expect) or `Tuple[str, Domain]` as commented above?

fritzo:
Thanks for the clear PR description!

funsor/tensor.py
eb8680: I think you're right that this can be relaxed to just a name with a domain inferred from `size`.  If we decide to support higher-rank `bint` domains in Funsor we'd need to provide a domain explicitly.

funsor/pyro/convert.py
```diff
@@ -63,21 +63,12 @@ def tensor_to_funsor(tensor, event_inputs=(), event_output=0, dtype="real"):
...
-    if squeezed_shape != inputs_shape:
-        tensor = tensor.reshape(squeezed_shape + output_shape)
-
-    return Tensor(tensor, inputs, dtype)
+    dim_to_name_list = DIM_TO_NAME + event_inputs if event_inputs else DIM_TO_NAME
+    dim_to_name = OrderedDict(zip(
+        range(-len(inputs_shape), 0),
+        zip(dim_to_name_list[len(dim_to_name_list) - len(inputs_shape):],
+            map(bint, inputs_shape))))
+    return to_funsor(tensor, Domain(dtype=dtype, shape=output_shape), dim_to_name)
```
eb8680: > could you point out how the `dim_to_name` kwarg is exercised by existing tests

Because of this refactoring of `funsor.pyro.convert.tensor_to_funsor`, the tests in `test/pyro/test_convert.py` that exercise `funsor.pyro.convert.tensor_to_funsor` also exercise `to_funsor`, and similarly for `to_data` via `funsor.pyro.convert.funsor_to_tensor` below.  `test/pyro/test_convert.py::test_tensor_funsor_tensor` does so most explicitly.

fritzo: Thanks for explaining!

fritzo:
Sorry I missed this last week. Feel free to "bump" if I take longer than a couple days to review.

PR #317, opened 2020-02-14
Fix numpy expand and make numpy backend jittable

fehiepsi:
Following fixes are detected while I did profiling GaussianHMM in #315.
- Support `-1` in expand shape for numpy backend using @fritzo 's suggestion
- `UnshapedArray` is replaced by `jax.core.Tracer`. A device array `x` will become `Tracer` under jit/vmap/pmap. This tracer has an attribute `x.aval`, which is UnshapedArray.
- Make `eager_cat_homogeneous` backend agnostic

Follow-up PR:
- enhance the speed of Gaussian

### DIFF
--- a/funsor/tensor.py
+++ b/funsor/tensor.py
@@ -326,7 +326,7 @@ class Tensor(Funsor, metaclass=TensorMeta):
         #   Then g is an unbiased estimator of f in value and all derivatives.
         #   In the special case f = detach(f), we can simplify to
         #       g = delta(x=x0) |f|.
-        if flat_logits.requires_grad:
+        if hasattr(flat_logits, "requires_grad") and flat_logits.requires_grad:
             # Apply a dice factor to preserve differentiability.
             index = [ops.new_arange(self.data, n).reshape((n,) + (1,) * (len(flat_logits.shape) - i - 2))
                      for i, n in enumerate(flat_logits.shape[:-1])]
@@ -337,7 +337,7 @@ class Tensor(Funsor, metaclass=TensorMeta):
                                   (log_prob - log_prob.detach()), sb_inputs))
         else:
             # This is the special case f = detach(f).
-            results.append(Tensor(flat_logits.logsumexp(-1), batch_inputs))
+            results.append(Tensor(ops.logsumexp(flat_logits, -1), batch_inputs))
 
         return reduce(ops.add, results)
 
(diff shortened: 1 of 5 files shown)


funsor/tensor.py
```diff
@@ -326,7 +326,7 @@ def unscaled_sample(self, sampled_vars, sample_inputs):
         #   Then g is an unbiased estimator of f in value and all derivatives.
         #   In the special case f = detach(f), we can simplify to
         #       g = delta(x=x0) |f|.
-        if flat_logits.requires_grad:
+        if hasattr(flat_logits, "requires_grad") and flat_logits.requires_grad:
```
fehiepsi: I am not sure if for numpy backend, we should exercise `if` or `else` branch here.

fritzo: 1. Does JAX have a `.detach()` method, as needed in the DiCE branch? Does jax have a `.requires_grad` checker?
2. Should we have three backends, "numpy", "jax", and "torch"? In that case, "numpy" should use the second branch and jax should use the first (i.e. DiCE) branch.

eb8680: > Should we have three backends, "numpy", "jax", and "torch"?

+1 to this, with "numpy" as the default backend - that way dependencies heavier than numpy are optional and backend-specific

neerajprad: > Does JAX have a .detach() method, as needed in the DiCE branch? Does jax have a .requires_grad checker?

The analog of `detach` is [lax.stop_gradient](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.stop_gradient.html#jax.lax.stop_gradient), but I don't think there is a way to check for `requires_grad` flag (and probably wouldn't work under jit anyways). I suspect that we may have to write this as a JAX custom primitive and specify the behavior under ad.

fehiepsi: Thank all! I can understand the situation now. I'll address this in an upcoming PR, where I add global BACKEND "numpy"/"torch"/"jax". As @neerajprad pointed out, we can use `lax.stop_gradient` in this branch.

About `requires_grad`, I think we don't need to worry about it. If JAX users don't want to take `grad` here, the overhead of `if` branch is pretty small. Please let me know if I miss crucial points here (e.g. `sb_inputs` is different from `batch_inputs`...)

fritzo: OK, so as I understand, in this branch you will
```py
if backend == "torch" and not flat_logits.requires_grad:
    # use the shorcut
else:  # numpy, jax, and torch with grad
    # use the DiCE version
```
and in a follow-up PR you can add "numpy"/"jax"/"torch" and optimize the numpy branch?
```py
if backend == "numpy" or (backend == "torch" and not flat_logits.requires_grad):
    # use the shorcut
else:  # jax, and torch with grad
    # use the DiCE version
```

fehiepsi: Yes, but let's keep this PR as-is because the logic here (for backends other than `torch`) is not exercised in any test yet.

PR #319, opened 2020-03-03
Refactor funsor.distributions

eb8680:
This PR adds a replacement for the `funsor.distributions` module in which PyTorch distributions are wrapped programmatically and `to_funsor` and `to_data` are defined generically.  The main API change in this version is that distributions now take a `name` argument rather than a `value`, making them behave a bit more like `Gaussian` or `Delta`, but I'm open to reverting that.

Remaining tasks:
- [ ] Add output shape inference to `tensor_to_funsor` so that distribution shape inference works
- [ ] Add more tests for conversion
- [ ] Make tests more generic
- [ ] Port any non-trivial `eager` rules for distribution evaluation to the new module
- [ ] Deprecate the existing `distributions` code


funsor/distributions2.py
```diff
@@ -0,0 +1,194 @@
...
+    @classmethod
+    def _infer_value_shape(cls, **kwargs):
+        # rely on the underlying distribution's logic to infer the event_shape
+        instance = cls.dist_class(**{k: _dummy_tensor(v.output) for k, v in kwargs.items()})
+        out_shape = instance.event_shape
+        if isinstance(instance.support, torch.distributions.constraints._IntegerInterval):
+            out_dtype = instance.support.upper_bound + 1
+        else:
+            out_dtype = 'real'
+        return Domain(dtype=out_dtype, shape=out_shape)
```
eb8680: This method is the main innovation in this PR: it turns out we can reuse the distributions' existing `event_shape` and `support` inference logic to generically infer the shape of the fresh `value` input, rather than reimplementing that logic by hand for every distribution we add.

funsor/distributions2.py
```diff
@@ -0,0 +1,194 @@
...
+
+    dist_class = DistributionMeta2(pyro_dist_class.__name__, (Distribution2,), {
+        'dist_class': pyro_dist_class,
+        '__init__': dist_init,
+    })
+
+    return dist_class
+
+
+class BernoulliProbs(dist.Bernoulli):
```
eb8680: I added a few of these wrappers for distributions with multiple parametrizations, but these should probably be moved upstream to Pyro or PyTorch.

eb8680:
Closing in favor of #320 which is similar but narrower in scope

PR #320, opened 2020-03-03
Refactor funsor.distributions without changing API

### LINKED ISSUE #159: Generate Distribution and RandomVariable wrapper classes programmatically

eb8680:
Resolves #159 

This PR refactors the `funsor.distributions` module so that PyTorch distributions are wrapped programmatically and `to_funsor` and `to_data` can be defined generically.  It is similar to #319 except that it does not change the `funsor.distributions` API and does not contain new `to_funsor`/`to_data` logic.

Necessary to support the broader refactoring around funsor integration and backend independence in https://github.com/pyro-ppl/pyro/pull/2307.  In particular, this PR is a prerequisite for Funsor support in NumPyro.


funsor/distributions.py
```diff
@@ -88,68 +88,98 @@ def eager_reduce(self, op, reduced_vars):
...
+
+    @classmethod
+    def _infer_value_domain(cls, **kwargs):
+        # rely on the underlying distribution's logic to infer the event_shape
+        instance = cls.dist_class(**{k: _dummy_tensor(v.output) for k, v in kwargs.items()}, validate_args=False)
+        out_shape = instance.event_shape
+        if isinstance(instance.support, torch.distributions.constraints._IntegerInterval):
+            out_dtype = int(instance.support.upper_bound + 1)
+            # this is a hack, but we don't really care about precise dtypes except for Categorical
+            out_dtype = 'real' if out_dtype == 1 else out_dtype
```
fehiepsi: my curiousity: which discrete distribution has upper_bound = 0?

funsor/distributions.py
```diff
@@ -88,68 +88,98 @@ def eager_reduce(self, op, reduced_vars):
...
 
+    def __getattribute__(self, attr):
+        if attr in type(self)._ast_fields and attr != 'name':
+            return self.params[attr]
+        return super().__getattribute__(attr)
+
+    @classmethod
+    def _infer_value_domain(cls, **kwargs):
+        # rely on the underlying distribution's logic to infer the event_shape
+        instance = cls.dist_class(**{k: _dummy_tensor(v.output) for k, v in kwargs.items()}, validate_args=False)
```
fehiepsi: I think this is costly. Can we store `event_shape` information in `_wrapped_pyro_dists`?

funsor/distributions.py
eb8680: This is a hack to make some `dist.Binomial` tests pass.  It would probably be better to remove this and add a custom `Binomial._infer_value_domain`, as with `dist.Delta` below.

funsor/distributions.py
eb8680: I'm not sure I understand your proposal, but I don't think this is unreasonably costly (note `_dummy_tensor` is always an expanded scalar), and I don't see how to implement `_infer_value_domain` generically without either using this trick or reimplementing `event_shape` inference for every underlying distribution by hand, which is what the previous version did.

fehiepsi: Understood! (I was confused about `event_shape` and `event_dim`).

Some distributions have additional numeric ops in their constructions. But I don't have a solution for this issue.

funsor/distributions.py
```diff
@@ -88,68 +88,98 @@ def eager_reduce(self, op, reduced_vars):
...
-    """
-    Wraps :class:`pyro.distributions.Bernoulli` .
+def make_dist(pyro_dist_class, param_names=()):
 
-    :param Funsor probs: Probability of 1.
-    :param Funsor value: Optional observation in ``{0,1}``.
-    """
-    dist_class = dist.Bernoulli
+    if not param_names:
+        param_names = tuple(pyro_dist_class.arg_constraints.keys())
```
fritzo: This is nondeterministic. We should either add an assertion
```py
assert len(pyro_dist_class.arg_constraints) <= 1
```
or parse the `.__init__()` method to determine order.

funsor/distributions.py
```diff
@@ -88,68 +88,98 @@ def eager_reduce(self, op, reduced_vars):
...
 
-    @staticmethod
-    def _fill_defaults(probs, value='value'):
-        probs = to_funsor(probs)
-        assert probs.dtype == "real"
-        value = to_funsor(value, reals())
-        return probs, value
+    @makefun.with_signature(f"__init__(self, {', '.join(param_names)}, value='value')")
+    def dist_init(self, *args, **kwargs):
+        return Distribution.__init__(self, *tuple(kwargs.values()))
```
fritzo: Isn't this nondeterministic? Why not simply pass through `**kwargs`?

funsor/distributions.py
```diff
@@ -88,68 +88,98 @@ def eager_reduce(self, op, reduced_vars):
...
-@eager.register(BernoulliProbs, Tensor, Tensor)
-def eager_bernoulli(probs, value):
-    return BernoulliProbs.eager_log_prob(probs=probs, value=value)
+    return dist_class
 
 
-class BernoulliLogits(Distribution):
-    """
-    Wraps :class:`pyro.distributions.Bernoulli` .
+class BernoulliProbs(dist.Bernoulli):
```
fritzo: I'm pretty sure you'll need to add underscores to these local classes for serializability, because otherwise you're overwriting them in the loop below.

funsor/distributions.py
```diff
@@ -42,17 +48,11 @@ class DistributionMeta(FunsorMeta):
...
-        # If value was explicitly specified, evaluate under current interpretation.
-        if 'value' in kwargs:
-            return super(DistributionMeta, cls).__call__(*args)
-
-        # Otherwise lazily construct a distribution instance.
-        # This makes it cheaper to construct observations in minipyro.
-        with interpretation(lazy):
-            return super(DistributionMeta, cls).__call__(*args)
+        value = kwargs.pop('value', 'value')
+        kwargs = OrderedDict((k, to_funsor(v)) for k, v in kwargs.items())
```
fritzo: Did I miss a major change in Python 3.6 that makes kwargs deterministically ordered? Can you point me to any reading material about this change?

eb8680: Insertion order in dictionaries was [made deterministic in Python 3.6](https://docs.python.org/3/whatsnew/3.6.html#new-dict-implementation) and this was [added to the language spec in 3.7](https://docs.python.org/3/whatsnew/3.7.html) ("Python data model improvements").  I guess that's why this code isn't breaking.

funsor/distributions.py
fritzo: I also think this is costly, triggering needless tensor ops and Python overhead and GPU-CPU communication each time a distribution is constructed. Can we enter this in a table instead? Maybe you could you move this probing logic to an on-library-load helper that populates the table of shaping metadata?

test/test_distributions.py
```diff
@@ -166,7 +166,6 @@ def categorical(probs, value):
 def test_delta_defaults():
     v = Variable('v', reals())
     log_density = Variable('log_density', reals())
-    assert isinstance(dist.Delta(v, log_density), dist.Delta)
```
fritzo: what changed here?

funsor/distributions.py
eb8680: > Can we enter this in a table instead?

The result of `_infer_value_domain` depends on the shapes and sometimes even the values (in the case of `Binomial` or `Multinomial`) of distribution parameters.  I could certainly memoize it but I'm not sure I see how to avoid constructing a new distribution at runtime for new parameter shapes - what did you have in mind?

funsor/distributions.py
fritzo: > The order-preserving aspect of this new implementation is considered an implementation detail and should not be relied upon

I'd feel safer if we avoided reliance on deterministic order for now (there being so many other potential pitfalls). How much extra work would it be to avoid relying on this?

eb8680: Shouldn't be much, I'll make all the suggested determinism-related changes

fritzo: Oh look at this, [PEP 468](https://www.python.org/dev/peps/pep-0468/) was [accepted](https://docs.python.org/3/whatsnew/3.6.html#pep-468-preserving-keyword-argument-order) in Python 3.6. So I guess we're safe :smile:

> **kwargs in a function signature is now guaranteed to be an insertion-order-preserving mapping.

funsor/distributions.py
fritzo: Could you change this one for now, as it is only safe in Python 3.7+?

funsor/distributions.py
fritzo: This should be safe in Python 3.6+ but might as well pass `**kwargs` as it's clearer.

funsor/distributions.py
```diff
@@ -88,69 +91,111 @@ def eager_reduce(self, op, reduced_vars):
...
-@eager.register(BernoulliProbs, Tensor, Tensor)
-def eager_bernoulli(probs, value):
-    return BernoulliProbs.eager_log_prob(probs=probs, value=value)
+    return dist_class
 
 
-class BernoulliLogits(Distribution):
-    """
-    Wraps :class:`pyro.distributions.Bernoulli` .
+class __BernoulliProbs(dist.Bernoulli):
```
fritzo: Is there a reason you're using double underscore here? I try to avoid double underscore because it makes debugging difficult due to Python's name mangling (double underscores are generally avoided for that reason).

funsor/distributions.py
```diff
@@ -88,69 +91,111 @@ def eager_reduce(self, op, reduced_vars):
...
+    (dist.VonMises, ()),
+    (dist.Dirichlet, ()),
+    (dist.DirichletMultinomial, ()),
+    (dist.Normal, ()),
+    (dist.MultivariateNormal, ('loc', 'scale_tril')),
+    (dist.Delta, ()),
+]
+
+for pyro_dist_class, param_names in _wrapped_pyro_dists:
+    locals()[pyro_dist_class.__name__.split("__")[-1].split(".")[-1]] = make_dist(pyro_dist_class, param_names)
```
fritzo: ditto, could you use a single underscore and `.lstrip("_")`?

funsor/distributions.py
```diff
@@ -88,69 +91,111 @@ def eager_reduce(self, op, reduced_vars):
...
+
+# Delta has to be treated specially because of its weird shape inference semantics
+Delta._infer_value_domain = classmethod(lambda cls, **kwargs: kwargs['v'])
+
+
+# Multinomial and related dists have dependent bint dtypes, so we just make them 'real'
```
fritzo: Good point, would you mind pointing to this new issue so we can start planning refactoring?
https://github.com/pyro-ppl/funsor/issues/322

funsor/distributions.py
```diff
@@ -63,7 +66,7 @@ class Distribution(Funsor, metaclass=DistributionMeta):
         funsors or objects that can be coerced to funsors via
         :func:`~funsor.terms.to_funsor` . See derived classes for details.
     """
-    dist_class = "defined by derived classes"
+    dist_class = dist.Distribution
```
fritzo: Is this value actually used, or are you merely tidying up to be well-typed?

funsor/distributions.py
fritzo: memoization based on shape seems fine.

funsor/distributions.py
eb8680: I changed it to avoid confusion in the wrapping logic with any custom distribution types (like Pyro's `_Subsample`) that have a `_` prefix for some reason, but I'll change it to something other than double underscore.

funsor/distributions.py
eb8680: Just tidying up to be well-typed.

funsor/distributions.py
fritzo: @eb8680 maybe I'm overreacting about the effort required to resume support for Python 3.5. This comment thread illustrates my main concern, that we would need to pay special attention to order in kwargs and other use of dictionaries.

eb8680: Python 3.5 compatibility in Funsor [has always been nominal at best](https://github.com/pyro-ppl/funsor/commit/986e75af9a3e32b65c2d2b5d4774ebcdbff5b9ab#diff-354f30a63fb0907d4ad57269548329e3R31).  I am happy to reopen and continue working on #329 but that only makes sense if we can agree beforehand on specific, objective criteria that would give us enough confidence to go ahead with adding Funsor as an optional dependency to Pyro.

fritzo: > nominal at best

Well early on we support Python 2.7 and we took care to ensure determinism. I think this PR on March 25 was the first where I stopped reviewing for dict order determinism.

> if we can agree beforehand on specific, objective criteria

Sounds reasonable. I think there are two concerns: ensuring determinism and ensuring correctness. I'm willing to give up determinism (at least across python processes; see [PYTHONHASHSEED](https://docs.python.org/3.3/using/cmdline.html#envvar-PYTHONHASHSEED) in Python 3.3+). But we'll need to avoid dangerous nondeterminism, e.g. 
```py
# This is correct only in Python 3.6+:
Tensor(torch.randn(2, 3), OrderedDict(a=bint(2), b=bint(3)))
```
Since this is difficult to test, I can only offer subjective criteria. To make this as objective as possible, I suggest we split your 3.5 PR #329 into two parts:
1. #329 objective: get existing tests to pass; followed by
2. #xxx subjective: we meet for 1 hour to decide subjective criteria in the form of say regular expressions we can run agains source code (e.g. `OrderedDict(\w+(\w\d)?=`). We code those up in a `test_source_code.py` and make a second PR that changes only enough code to satisfy those properties.

wdyt?

eb8680: Those are good suggestions, but I would prefer to avoid any further speculative coding.  To that end I suggest we reverse the order of operations and start a PR with a failing `test_source_code.py` and a list of any other show-stopping compatibility problems, ideally expressed as failing unit tests.  

If we can agree that fixing the problems in the PR is *both necessary and sufficient* to add Funsor as an optional dependency to Pyro, I can make the actual fixes including getting existing tests to pass in #329.  If we cannot agree, or the list appears to require too much engineering effort, then there's no reason to proceed with #329 or any related work.

If this plan is acceptable, I will put up a PR in the next few days with a seed version of `test_source_code.py`.

fritzo: That sounds like a reasonable plan. I'm happy to chat about `test_source_code.py` or just see what you come up with 🙂

It might help us discover dangerous patterns to simply run the tests under Python 3.5.

PR #321, opened 2020-03-04
Add generic to_funsor conversion methods for funsor.distributions

eb8680:
This PR adds implementations of `to_funsor` and `to_data` for the revamped `funsor.distributions.Distribution` from #320.  These implementations make use of a generic `Distribution._infer_param_domain` method that uses the `arg_constraints` `TorchDistribution` metadata to infer event shapes.

The git history is kind of gross, but there's nothing too complicated that's new here.

Tested:
- Like #316, the new `to_funsor`/`to_data` implementations are exercised by existing tests in `test/pyro/test_convert.py` via the changes in this PR to `funsor.pyro.convert`


funsor/distributions.py
```diff
@@ -194,6 +215,75 @@ def _multinomial_infer_value_domain(cls, **kwargs):
...
+# Converting PyTorch Distributions to funsors
+###############################################
+
+@to_funsor.register(torch.distributions.Distribution)
+def torchdistribution_to_funsor(pyro_dist, output=None, dim_to_name=None):
+    import funsor.distributions  # TODO find a better way to do this lookup
+    funsor_dist_class = getattr(funsor.distributions, type(pyro_dist).__name__.split("__")[-1])
+    params = [to_funsor(
+            getattr(pyro_dist, param_name),
+            output=funsor_dist_class._infer_param_domain(
```
eb8680: `Distribution._infer_param_domain` is used here to enable a generic `torchdistribution_to_funsor` conversion that works by converting `TorchDistribution` parameters to funsors and constructing `Distribution` funsors from the results.

funsor/distributions.py
```diff
@@ -104,7 +110,7 @@ def __getattribute__(self, attr):
         return super().__getattribute__(attr)
 
     @classmethod
-    @functools.lru_cache(maxsize=None)
+    @functools.lru_cache(maxsize=5000)
```
fritzo: Just curious, were you seeing unbounded growth?

funsor/distributions.py
fritzo: Do I understand correctly that this assumes all parameters are batched scalars? So e.g. `MultivariateNormal` and `LowRankMultivariateNormal` and `Dirichlet` will need special casing? If so do you think we could add an `assert .event_dim == 0`?

eb8680: `_infer_param_domain` looks at the `arg_constraint` for each parameter to infer their `output`s.  In the case of `MultivariateNormal`, `arg_constraints["loc"]` is a `constraints.RealVector` object, so `_infer_param_domain` knows that its output shape is `loc.shape[-1:]`.

funsor/distributions.py
eb8680: Not in any existing tests, I just figured it would be good to add a ceiling.

eb8680:
> Is there any logic you'd like reviewed with extra scrutiny?

Nope, this is mostly boring refactoring

> Can you confirm these changes are pure refactoring and thus covered by existing tests?

Everything here should be covered by existing tests in `test/pyro/test_convert.py`.

PR #323, opened 2020-03-25
Add generic unscaled_sample method to Distribution

eb8680:
This PR adds a `Distribution.unscaled_sample` method for rewriting `Distribution` funsors to `Delta` funsors with correct Dice factors computed generically from Pyro's `TorchDistribution.score_parts` methods.

Questions for review:
- Is there a better way to test gradients generically?
- Can we make the tests less expensive and more reliable?

### DIFF
--- a/funsor/distributions.py
+++ b/funsor/distributions.py
@@ -216,6 +216,7 @@ _wrapped_pyro_dists = [
     (dist.Normal, ()),
     (dist.MultivariateNormal, ('loc', 'scale_tril')),
     (dist.Delta, ()),
+    (fakes.NonreparameterizedBeta, ()),
     (fakes.NonreparameterizedGamma, ()),
     (fakes.NonreparameterizedNormal, ()),
     (fakes.NonreparameterizedDirichlet, ()),
--- a/funsor/distributions.py
+++ b/funsor/distributions.py
@@ -113,7 +113,7 @@ class Distribution(Funsor, metaclass=DistributionMeta):
         inputs_, tensors = align_tensors(*params.values())
         inputs = OrderedDict(sample_inputs.items())
         inputs.update(inputs_)
-        sample_shape = tuple(v.dtype for v in sample_inputs.values())
+        sample_shape = tuple(v.size for v in sample_inputs.values())
 
         raw_dist = self.dist_class(**dict(zip(self._ast_fields[:-1], tensors)))
         if getattr(raw_dist, "has_rsample", False):
--- a/funsor/distributions.py
+++ b/funsor/distributions.py
@@ -111,18 +111,21 @@ class Distribution(Funsor, metaclass=DistributionMeta):
         assert all(isinstance(v, (Number, Tensor)) for v in params.values())
         assert isinstance(value, Variable) and value.name in sampled_vars
         inputs_, tensors = align_tensors(*params.values())
+        inputs = OrderedDict(sample_inputs.items())
+        inputs.update(inputs_)
+        sample_shape = tuple(v.dtype for v in sample_inputs.values())
+
         raw_dist = self.dist_class(**dict(zip(self._ast_fields[:-1], tensors)))
         if getattr(raw_dist, "has_rsample", False):
-            raw_sample = raw_dist.rsample(tuple(v.dtype for v in sample_inputs.values()))
+            raw_sample = raw_dist.rsample(sample_shape)
         else:
-            raw_sample = raw_dist.sample(tuple(v.dtype for v in sample_inputs.values()))
-        raw_score_function = raw_dist.score_parts(raw_sample).score_function
-        inputs = OrderedDict((k, v) for k, v in sample_inputs.items())
-        inputs.update(inputs_)
+            raw_sample = raw_dist.sample(sample_shape)
+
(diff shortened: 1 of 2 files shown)


funsor/distributions.py
```diff
@@ -104,6 +105,27 @@ def eager_log_prob(cls, *params):
...
+        assert all(isinstance(v, (Number, Tensor)) for v in params.values())
+        assert isinstance(value, Variable) and value.name in sampled_vars
+        inputs_, tensors = align_tensors(*params.values())
+        raw_dist = self.dist_class(**dict(zip(self._ast_fields[:-1], tensors)))
+        if getattr(raw_dist, "has_rsample", False):
+            raw_sample = raw_dist.rsample(tuple(v.dtype for v in sample_inputs.values()))
+        else:
+            raw_sample = raw_dist.sample(tuple(v.dtype for v in sample_inputs.values()))
+        raw_score_function = raw_dist.score_parts(raw_sample).score_function
+        inputs = OrderedDict((k, v) for k, v in sample_inputs.items())
```
fritzo: nit: simplify to `OrderedDict(sample_inputs.items())`?

funsor/distributions.py
```diff
@@ -104,6 +105,27 @@ def eager_log_prob(cls, *params):
...
+        params = OrderedDict(self.params)
+        value = params.pop("value")
+        assert all(isinstance(v, (Number, Tensor)) for v in params.values())
+        assert isinstance(value, Variable) and value.name in sampled_vars
+        inputs_, tensors = align_tensors(*params.values())
+        raw_dist = self.dist_class(**dict(zip(self._ast_fields[:-1], tensors)))
+        if getattr(raw_dist, "has_rsample", False):
+            raw_sample = raw_dist.rsample(tuple(v.dtype for v in sample_inputs.values()))
+        else:
+            raw_sample = raw_dist.sample(tuple(v.dtype for v in sample_inputs.values()))
```
fritzo: nit: Could you factor out the common subexpression as
```py
sample_shape = torch.Size(v.size for v in sample_inputs.values())
```

funsor/distributions.py
```diff
@@ -104,6 +105,27 @@ def eager_log_prob(cls, *params):
...
+        raw_score_function = raw_dist.score_parts(raw_sample).score_function
```
fritzo: This performs unnecessary work: in case `has_rsample == True` it unnecessarily computes `log_prob`. Could we replace with
```py
if not raw_dist.has_rsample:
    log_prob = raw_dist.log_prob(raw_sample)
    dice_factor = Tensor(log_prob - log_prob.detach(), inputs)
    result += dice_factor
```
If you want to support `Rejector` you could add extra logic like `if type(self).score_parts is Distribution.score_parts)`, but I doubt anyone uses `Rejector`.

funsor/distributions.py
```diff
@@ -104,6 +105,27 @@ def eager_log_prob(cls, *params):
...
+        inputs = OrderedDict((k, v) for k, v in sample_inputs.items())
+        inputs.update(inputs_)
+        result = funsor.delta.Delta(value.name, Tensor(raw_sample, inputs, value.output.dtype))
+        if not dist.util.is_identically_zero(raw_score_function):
+            # scaling of dice_factor by num samples should already be handled by Funsor.sample
+            dice_factor = Tensor(raw_score_function - raw_score_function.detach(), inputs)
+            result += dice_factor
```
fritzo: ditto: see above

test/test_distributions.py
```diff
@@ -654,3 +656,185 @@ def von_mises(loc, concentration, value):
...
+        check_funsor(actual_mean, expected_mean.inputs, expected_mean.output)
+        assert_close(actual_mean, expected_mean, atol=atol, rtol=rtol)
+
+        actual_variance = Integrate(
+            sample_value,
+            (Variable('value', funsor_dist.inputs['value']) - actual_mean) ** 2,
+            frozenset(['value'])
+        ).reduce(ops.add, frozenset(sample_inputs))
+        grad_targets = [v.data for v in list(funsor_dist.params.values())[:-1]]
+        actual_grads = torch.autograd.grad(actual_variance.reduce(ops.add).sum().data, grad_targets, allow_unused=True)
```
fritzo: Just curious, is there a reason you're computing gradients wrt variance rather than mean or other statistic?

eb8680: No particular reason, so I just switched to mean by default and added variance and entropy as options.

funsor/distributions.py
```diff
@@ -104,6 +105,30 @@ def eager_log_prob(cls, *params):
...
 
+    def unscaled_sample(self, sampled_vars, sample_inputs):
+        params = OrderedDict(self.params)
+        value = params.pop("value")
+        assert all(isinstance(v, (Number, Tensor)) for v in params.values())
+        assert isinstance(value, Variable) and value.name in sampled_vars
+        inputs_, tensors = align_tensors(*params.values())
+        inputs = OrderedDict(sample_inputs.items())
+        inputs.update(inputs_)
+        sample_shape = tuple(v.dtype for v in sample_inputs.values())
```
fritzo: nit: `v.dtype` -> `v.size`, for readability and type checking

PR #326, opened 2020-04-21
make funsor.distributions backend agnostic

fehiepsi:
Probably this is the last piece that is not yet backend-agnostic?

#### Utilities
+ add `ops.astype` to convert a numeric array to a specific dtype (hence `funsor.testing.astype` is removed)
+ move `numeric_array` and `dummy_numeric_array` to `funsor.tensor`

#### API updates
+ Add `rng_key` to `unscaled_sample` methods of Gaussian/Tensor/Contraction/Delta/.... Looking like only `Contraction` requires us split the PRNG key.

#### Reorganize
+ Files `torch.py`, `jax.py` are renamed into `ops.py` and moved to `torch`, `jax` folders. These folders are used to register backend-dependent implementations.
+ `funsor.distributions` are refactored (without changing API) to support numpyro backend. Common code is kept as-is.

#### Distribution details
+ `funsor.distributions.Normal` is a function which returns either `funsor.jax.distributions.Normal` or `funsor.torch.distributions.Normal`. Those backend distributions have the same functionality as the previous `funsor.distributions.Normal`.
+ I need to have distributions' signatures are stored in the dict `FUNSOR_DIST_NAMES` because when creating `funsor.distributions.foo` distribution functions, we don't have backend distributions to lookup the signature yet. In addition, some distributions available in Pyro are not available in NumPyro.

#### Other changes
+ I have ported `funsor.pyro` to `funsor.numpyro` to test if things work as expected. Most of the changes in this PR are files in `funsor.numpyro`. The implementations are quite similar, except that `funsor.numpyro` is mostly backend-agnostic.

##### TODO
- [x] make funsor.numpyro.convert backend-agnostic
- [x] port funsor.pyro.distributions to numpyro ...
- [x] port funsor.pyro.hmm to numpyro ...
- [x] move backend-dependent code from funsor.distributions to funsor.pyro.distributions.

Currently, I got two issues:
- [x] numpyro sample method requires `key` argument; I think this is a good time to think about a solution for this issue
- [x] name collisions for numpyro and pyro distributions; I think I can resolve this issue by making `dist_class` a property attribute instead of a static attribute.

- [x] make all tests pass


eb8680:
@fehiepsi I think it's worth trying this approach but there are enough minor differences in APIs and naming conventions that it might end up being easier to just have separate `funsor.{backend}.distributions` modules for each backend, especially if we add a TF backend.  Even then we could still centralize some things, like the base `distribution` class and some of the `eager` patterns and `to_funsor`/`to_data` patterns in `funsor.distributions`.

> numpyro sample method requires key argument

Yeah, this is tricky.  I guess for now to fully support the JAX backend we'll just have to propagate and split key arguments through every `.unscaled_sample` implementation in Funsor.  Hopefully future refactoring will reduce the number of `unscaled_sample` implementations so that this is less of a maintenance burden.

eb8680:
@fehiepsi awesome work!  Right now this PR is quite large, and it looks like there's still some deduplication to do, so I'd like to clean things up before landing the changes in `master`.  One thing we can do to simplify it a bit is break out `funsor.numpyro` into one or more separate PRs.

As a first step toward this goal, I would like to deprecate `funsor.pyro.convert` entirely, replacing all calls to that API in the tests with calls to `to_funsor`/`to_data` and making `AffineNormal` backend-agnostic and self-contained.  I'll put up a PR that attempts this and you or I can then make similar changes here in `funsor.numpyro`, which should subsequently be much easier to separate from the rest of the changes here.

fehiepsi:
@eb8680 Thanks for taking a look! Agreed, I should make a separate PR which includes changes in `funsor.torch`, `funsor.jax` and remove all `funsor.numpyro`-related changes.

PR #327, opened 2020-05-04
Make funsor.distributions backend agnostic

fehiepsi:
Blocked by https://github.com/pyro-ppl/pyro-api/pull/19

This is similar to #326 but doesn't include `funsor.numpyro` stuff. I have resolved `has_rsample` issue and make the whole `test_distributions` backend agnostic.


test/test_distributions.py
```diff
@@ -704,18 +717,45 @@ def _check_sample(funsor_dist, sample_inputs, inputs, atol=1e-2, rtol=None,
...
-            expected_stat.reduce(ops.add).sum().data, grad_targets, allow_unused=True)
+        diff = (actual_stat - expected_stat).reduce(ops.add).data
+        return diff.sum(), diff
 
-        assert_close(actual_stat, expected_stat, atol=atol, rtol=rtol)
 
-        for actual_grad, expected_grad in zip(actual_grads, expected_grads):
-            if expected_grad is not None:
-                assert_close(actual_grad, expected_grad, atol=atol, rtol=rtol)
+def _check_sample(funsor_dist_class, params, sample_inputs, inputs, atol=1e-2,
```
fehiepsi, line 724: This is rewritten to also test for JAX distributions. The functionality should be the same as before.

funsor/terms.py
```diff
@@ -443,7 +443,7 @@ def reduce(self, op, reduced_vars=None):
         assert reduced_vars.issubset(self.inputs)
         return Reduce(op, self, reduced_vars)
 
-    def sample(self, sampled_vars, sample_inputs=None):
+    def sample(self, sampled_vars, sample_inputs=None, rng_key=None):
```
eb8680, line 446: We should look at making `test_samplers.py` backend-agnostic as well to exercise this new code path, but that doesn't have to happen in this PR.

fehiepsi: Oh, thanks for pointing it out. I missed that test file.

funsor/distributions.py
```diff
@@ -168,134 +172,89 @@ def _infer_param_domain(cls, name, raw_shape):
...
-    def dist_init(self, **kwargs):
-        return Distribution.__init__(self, *tuple(kwargs[k] for k in self._ast_fields))
 
-    dist_class = DistributionMeta(pyro_dist_class.__name__.split("_PyroWrapper_")[-1], (Distribution,), {
-        'dist_class': pyro_dist_class,
-        '__init__': dist_init,
-    })
+def make_dist(dist_name, param_names):
+    @makefun.with_signature(f"{dist_name}({', '.join(param_names)}, value='value')",
+                            eager_log_prob=functools.partial(_eager_log_prob, dist_name))
```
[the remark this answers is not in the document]
fehiepsi: This is used to call `dist.Categorical.eager_log_prob(...)`. It is not necessary if we don't merge the implementation: i.e. let users import `funsor.backend.distributions` separately.

funsor/jax/distributions.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+def _get_numpyro_dist(dist_name):
+    if dist_name in ['Binomial', 'Categorical', 'Multinomial'] or dist_name.startswith('Nonreparameterized'):
+        return globals().get('_NumPyroWrapper_' + dist_name)
+    else:
+        return getattr(dist, dist_name, None)
+
+
+def make_dist(backend_dist_class, param_names=()):
```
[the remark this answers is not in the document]
fehiepsi: Sure, we can move it back.

funsor/distributions.py
```diff
@@ -150,14 +149,19 @@ def _infer_value_domain(cls, **kwargs):
...
-        elif isinstance(support, constraints._RealVector):
+        elif support_name == "_RealVector":
             output = reals(raw_shape[-1])
-        elif isinstance(support, (constraints._LowerCholesky, constraints._PositiveDefinite)):
+        elif support_name in ["_LowerCholesky", "_PositiveDefinite"]:
             output = reals(*raw_shape[-2:])
-        elif isinstance(support, constraints._Real) and name == "logits" and \
-                isinstance(cls.dist_class.arg_constraints["probs"], constraints._Simplex):
+        elif support_name == "_Real" and name == "logits" and \
+                "Categorical" in cls.dist_class.__name__:
```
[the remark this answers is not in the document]
fehiepsi: I think we can revert this change. In NumPyro, distributions like `CategoricalLogits` does not have `probs` constraints. I think we just need to add an additional check that `probs` in `arg_constraints` attribute.

funsor/distributions.py
```diff
@@ -168,134 +172,89 @@ def _infer_param_domain(cls, name, raw_shape):
...
+    def dist_fn(*args, **kwargs):
+        backend_dist_module = import_module(BACKEND_TO_DISTRIBUTIONS_BACKEND[get_backend()])
+        return getattr(backend_dist_module, dist_name)(*args, **kwargs)
 
-    eager.register(dist_class, *((Tensor,) * (len(param_names) + 1)))(dist_class.eager_log_prob)
+    return dist_fn
 
-    return dist_class
 
+FUNSOR_DIST_NAMES = {
```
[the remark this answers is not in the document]
fehiepsi: Sure, we should do it if we use `import funsor.backend.distributions as dist` pattern. Let me think more if it is a good way. Right now I don't have to change the import statement, just simply `import funsor.distributions as dist`. But it does not matter much I guess.

funsor/distributions.py
```diff
@@ -168,134 +172,89 @@ def _infer_param_domain(cls, name, raw_shape):
...
-def _multinomial_infer_value_domain(cls, **kwargs):
-    instance = cls.dist_class(**{k: _dummy_tensor(domain) for k, domain in kwargs.items()}, validate_args=False)
-    return reals(*instance.event_shape)
-
-
-Binomial._infer_value_domain = classmethod(_multinomial_infer_value_domain)
-Multinomial._infer_value_domain = classmethod(_multinomial_infer_value_domain)
-DirichletMultinomial._infer_value_domain = classmethod(_multinomial_infer_value_domain)
+for dist_name, param_names in FUNSOR_DIST_NAMES.items():
+    locals()[dist_name] = make_dist(dist_name, param_names)
```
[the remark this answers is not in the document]
fehiepsi: This `make_dist` is different from the old `make_dist`. This is used to create a distribution function, e.g. `def Categorical(...)`. This is not necessary anymore if we import distribution backend explicitly.

funsor/distributions.py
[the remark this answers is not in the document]
fehiepsi: Maybe, for now, we just do what you suggested, and rename `funsor.distributions` to `funsor.distribution` to avoid confusion. We can decide what is a good way to do later.

fehiepsi:
@eb8680 I have made the change and required users import `funsor.jax.distributions` or `funsor.torch.distributions` explicitly. But this will not work with pyroapi (because pyroapi requires `funsor.distributions`). Could you suggest a solution to that issue?

eb8680:
@fehiepsi do you mean the `"funsor"` backend in `pyroapi` defined [here](https://github.com/pyro-ppl/pyro-api/blob/master/pyroapi/dispatch.py#L148) requires `funsor.distributions`?  If so, I think it should work to just delete that backend from `pyroapi` and add a custom one in `funsor.minipyro`:
```py
# in funsor/minipyro.py
register_backend('funsor', {
    'distributions': 'funsor.{}.distributions'.format(get_backend()),
    'handlers': 'funsor.minipyro',
    'infer': 'funsor.minipyro',
    'ops': 'funsor.compat.ops',
    'optim': 'funsor.minipyro',
    'pyro': 'funsor.minipyro',
})
```

eb8680:
@fehiepsi I put up a PR in `pyroapi` that should unblock this along with the fix above.

fehiepsi:
Thanks, @eb8680 !

fritzo:
It seems pretty easy to keep `pyroapi` working with a few staged PRs:
1. create a tiny funsor PR that aliases `funsor.torch.distributions` to `funstor.distributions` (possibly bump the funsor version and release funsor)
2. create a tiny Pyroapi PR that switches the `funsor` backend to use `funsor.torch.distributions` and bumps the pyroapi version; then release pyroapi
3. update this PR to the new pyroapi version

WDYT? I am happy to qickly approve such a coordinated sequence of PRs and releases if they keep things working.

eb8680:
This PR is one of the final steps in supporting multiple backends in
Funsor. I am reluctant to continue blocking it for an unrelated fix for
Funsor minipyro.

setup.py
```diff
@@ -49,9 +49,9 @@
...
         ],
         'jax': [
-            'jax==0.1.59',
-            'jaxlib==0.1.38',
-            'numpyro @ git+https://github.com/pyro-ppl/numpyro.git@f2aefcf1b034418244c1e465c64bce469388e566#egg=numpyro'
+            'jax @ git+https://github.com/google/jax.git@2512ec6ebebf1b26d2aefbe618b4c147c251d194#egg=jax',
+            'jaxlib>=0.1.43',
+            'numpyro @ git+https://github.com/pyro-ppl/numpyro.git@152c096b5fd06dabd0d6decd0aafdaff431bcc9c#egg=numpyro'
         ],
         'test': [
```
fritzo, line 56: After https://github.com/pyro-ppl/funsor/pull/330 is merged and pyro-api 0.1.2 is released, you'll need to update `pyro-api>=0.1.2` below

fritzo: OK I have released pyroapi version 0.1.2, so you should now be able to bump the version.

fehiepsi: Thanks, Fritz!

test/pyro/test_pyro_distribution.py
```diff
@@ -0,0 +1,57 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+import pyro.distributions as dist
```
fehiepsi: This test file is renamed to avoid confliction with `test/test_distribution.py`

setup.py
```diff
@@ -38,20 +38,17 @@
         'multipledispatch',
         'numpy>=1.7',
         'opt_einsum>=2.3.2',
-        # TODO: remove pyro-ppl and torch when funsor.minipyro and funsor.pyro is refactored
-        'pyro-ppl>=0.5',
         'pytest>=4.1',
-        'torch>=1.3.0',
```
fehiepsi, line 44: @eb8680 Finally, we can remove those pyro/pytorch default packages. I have made some changes from your last review:
+ Add `.expand` method to Pyro Wrapper distribution because those subclasses modify the `__init__` method (otherwise, using `.expand` in those wrappers will throw errors). This does not happen in NumPyro because there, the logic is defered to ExpandedDistribution.
+ Remove `import minipyro` by default in `funsor.__init__`. Please let me know if it is a good way to do. Probably we can test if `pyro/pytorch` installed, we will import minipyro.

There might be some errors in Travis. I'll fix them when Travis reports. Then I think we are good to go, pending your further suggestions.

test/pyro/test_pyro_distribution.py
fritzo: I believe a cleaner solution is to add a file `tests/pyro/__init__.py` so PyTest can distinguish the test module names. I just had this same problem a couple days ago in Pyro.

fehiepsi: Thanks!

setup.py
fritzo: You'll need `'pyro-api>=0.1'` below

fehiepsi:
I think Travis will pass after https://github.com/pyro-ppl/numpyro/pull/596 merged.

test/test_distribution.py
```diff
@@ -234,23 +241,24 @@ def dirichlet(concentration, value):
 
 @pytest.mark.parametrize('batch_shape', [(), (5,), (2, 3)], ids=str)
 @pytest.mark.parametrize('event_shape', [(1,), (4,), (5,)], ids=str)
+@pytest.mark.xfail(raises=AttributeError, reason="DirichletMultinomial is not implemented yet in NumPyro")
```
eb8680, line 244: nit: unless you're planning to implement `DirichletMultinomial` in NumPyro, can you change this `xfail` to check if `get_backend() != 'torch'` instead of `raises=AttributeError`?

test/test_distribution.py
```diff
@@ -633,19 +654,20 @@ def gamma(concentration, rate, value):
 
 @pytest.mark.parametrize('batch_shape', [(), (5,), (2, 3)], ids=str)
 @pytest.mark.parametrize('syntax', ['eager', 'lazy'])
+@pytest.mark.xfail(raises=AttributeError, reason="VonMises is not implemented yet in NumPyro")
```
eb8680, line 657: nit: unless you're planning to implement `VonMises` in NumPyro, can you change this `xfail` to check if `get_backend() != 'torch'`?

test/test_distribution.py
```diff
@@ -499,27 +513,30 @@ def test_mvn_gaussian(batch_shape):
...
+    backend_module = import_module(BACKEND_TO_DISTRIBUTIONS_BACKEND[get_backend()])
+    assert isinstance(d1, backend_module.MultivariateNormal)
     d2 = reinterpret(d1)
     assert issubclass(type(d2), GaussianMixture)
     actual = d2(**data)
     expected = d1(**data)
     assert_close(actual, expected)
 
 
+@pytest.mark.xfail(get_backend() == 'jax', reason='dist_to_funsor for jax backend is not available yet')
```
eb8680, line 525: It should be straightforward to get these tests working with the JAX backend.  I'll take care of it in a followup PR that deprecates most of `funsor.pyro.convert`.

funsor/distribution.py
```diff
@@ -360,97 +324,96 @@ def Bernoulli(probs=None, logits=None, value='value'):
...
     if logits is not None:
-        return BernoulliLogits(logits, value)
+        return backend_dist.BernoulliLogits(logits, value)  # noqa: F821
     raise ValueError('Either probs or logits must be specified')
 
 
-@eager.register(Beta, Funsor, Funsor, Funsor)
+def LogNormal(loc, scale, value='value'):
+    """
+    Wraps :class:`pyro.distributions.LogNormal` .
```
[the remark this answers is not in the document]
fehiepsi: I think it won't work. I'll tweak it.

test/test_distribution.py
fehiepsi: Oh, good idea. Thanks, it is better.

test/test_distribution.py
fehiepsi: Thanks!

test/test_samplers.py
```diff
@@ -281,16 +319,18 @@ def test_gaussian_mixture_distribution(batch_inputs, event_inputs):
     batch_inputs = OrderedDict(batch_inputs)
     event_inputs = OrderedDict(event_inputs)
     sampled_vars = frozenset(['f'])
+    np.random.seed(1)  # the default random seed generate bad gaussian in jax backend
```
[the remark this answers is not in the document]
fehiepsi: It seems that this is an unnecessary change. I'll revert it. I remember that one of the test cases gives a bad precision matrix.

funsor/distribution.py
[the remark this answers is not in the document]
fritzo: That should work via intersphinx mappings 🙂 

https://github.com/pyro-ppl/funsor/blob/4f2c13931030b0b793e72d74337f87e52e2c519c/docs/source/conf.py#L192-L198

test/test_distribution.py
fritzo: good idea, even `@pytest.mark.skipif(get_backend() != 'torch')`

eb8680:
@fehiepsi looks like a few of the examples still import `funsor.distributions`, otherwise the tests are passing.

fehiepsi:
Thanks for throughout reviews, @eb8680 !

PR #328, opened 2020-05-11
Fix a userwarning about integer divison in PyTorch 1.5

eb8680:
`Tensor.unscaled_sample` has been triggering the following `UserWarning` with PyTorch 1.5:
```
UserWarning: Integer division of tensors using div or / is deprecated, and in a future release div will perform true division as in Python 3. Use true_divide or floor_divide (// in Python) instead.
```
This PR makes the suggested change (`/` to `//`).  It may be necessary to merge this to placate Travis before any of https://github.com/pyro-ppl/pyro/pull/2307 can be merged.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ docs: FORCE
 	$(MAKE) -C docs html
 
 lint: FORCE
-	flake8 --per-file-ignores='funsor/distributions.py:F821'
+	flake8
 
 license: FORCE
 	python scripts/update_headers.py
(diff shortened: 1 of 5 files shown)


fehiepsi:
Oops, I have spent several hours today to debug and fix this bug in JAX backend. @@

eb8680:
@fehiepsi do you mean that this change broke the JAX backend or fixed it?

fehiepsi:
This fixes it! I haven't merged master to my branch, so some tests in `test_samplers` are failing without warning.

Issue #243, opened 2019-09-18
Nondeterministic typing error on module load


fritzo:
I have seen a couple nondeterministic errors on loading the funsor library with error messages
```
KeyError: (typing.Union[funsor.ops.LogAddExpOp, funsor.ops.NullOp], <class 'funsor.ops.AddOp'>, <class 'frozenset'>, typing.Tuple[typing.Union[funsor.torch.Tensor, funsor.terms.Number], funsor.gaussian.Gaussian])
```
This is tough to reproduce.

<details>

```sh
$ pytest -vx test/examples/test_bart.py --pdb
============================================== test session starts ==============================================
platform darwin -- Python 3.7.2, pytest-5.0.1, py-1.8.0, pluggy-0.12.0 -- /Users/fritzobermeyer/miniconda2/envs/pyro3/bin/python
cachedir: .pytest_cache
benchmark: 3.2.2 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /Users/fritzobermeyer/github/pyro-ppl/funsor, inifile: setup.cfg
plugins: xdist-1.26.1, nbval-0.9.1, forked-1.0.2, benchmark-3.2.2
collecting ...
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> traceback >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
test/examples/test_bart.py:5: in <module>
    import funsor
funsor/__init__.py:8: in <module>
    from . import (
funsor/adjoint.py:7: in <module>
    from funsor.cnf import Contraction, nullop
funsor/cnf.py:170: in <module>
    Tuple[Union[Tensor, Number], Gaussian]]
funsor/terms.py:232: in __getitem__
    return cls._type_cache[arg_types]
../../../miniconda2/envs/pyro3/lib/python3.7/weakref.py:137: in __getitem__
    o = self.data[key]()
E   KeyError: (typing.Union[funsor.ops.LogAddExpOp, funsor.ops.NullOp], <class 'funsor.ops.AddOp'>, <class 'frozenset'>, typing.Tuple[typing.Union[funsor.torch.Tensor, funsor.terms.Number], funsor.gaussian.Gaussian])
[...]

PR #329, opened 2020-05-12
Make Funsor compatible with Python 3.5

eb8680:
Necessary for use with Pyro, which supports Python 3.5 and uses it in Travis.

As far as I can tell, this PR contains all of the changes necessary to support Python 3.5 - I couldn't find any lurking order dependence in the obvious places, especially the distribution wrappers.  The only substantive issue I found and fixed was caused by a mysterious nondeterministic edge case in `multipledispatch`, perhaps related to #243.  Otherwise, most of the changes are either removals of f-strings or replacements of dictionaries with `OrderedDict`s in test case lists to make pytest test collection deterministic.


fritzo:
bye bye f-strings 😢

fritzo:
This is going to be a significant review effort because we have started using the ordered dictionary assumption, but Python 3.5 doesn't follow ordered dictionary semantics. I think we'll need to replace `dict` with `OrderedDict` in many places.

eb8680:
> we have started using the ordered dictionary assumption

I think it's mostly confined to the tests and perhaps to `funsor.distributions`, hopefully we can fix the obvious cases here and deal with others as they come up.

setup.py
```diff
@@ -32,7 +32,7 @@
     },
     author='Uber AI Labs',
     author_email='fritzo@uber.com',
-    python_requires=">=3.6",
+    python_requires=">=3.5",
```
fritzo, line 35: Could you also update the `Programming Language :: Python :: 3.6` classifier below?

test/test_gaussian.py
```diff
@@ -207,17 +207,17 @@ def test_smoke(expr, expected_type):
...
-    {'x': reals(2, 3)},
-    {'x': reals(), 'y': reals()},
-    {'x': reals(2), 'y': reals(3)},
-    {'x': reals(4), 'y': reals(2, 3), 'z': reals()},
+    OrderedDict([('x', reals())]),
+    OrderedDict([('x', reals(4))]),
+    OrderedDict([('x', reals(2, 3))]),
+    OrderedDict([('x', reals()), ('y', reals())]),
+    OrderedDict([('x', reals(2)), ('y', reals(3))]),
+    OrderedDict([('x', reals(4)), ('y', reals(2, 3)), ('z', reals())]),
```
fritzo, line 220: nit: I think we can revert these changes. I had set this test up so that we could write concise `{}`-dicts in the param statements, then convert them to `OrderedDict` via `OrderedDict(list(sorted(xxx.items())))` below. I think this is still nice to do for readability.

Please let me know if I've misunderstand some issue with pytest collection and the `OrderedDict` change is required.

test/test_gaussian.py
```diff
@@ -231,17 +231,17 @@ def test_align(int_inputs, real_inputs):
 
 
 @pytest.mark.parametrize('int_inputs', [
-    {},
-    {'i': bint(2)},
-    {'i': bint(2), 'j': bint(3)},
+    OrderedDict(),
```
fritzo, line 234: ditto here and in all subsequent tests

test/test_terms.py
```diff
@@ -45,6 +45,11 @@
 np.seterr(all='ignore')
 
 
+# have to make this deterministic for pytest collection to work
```
fritzo, line 48: Could you explain the type of error you saw, so I can keep an eye out for similar errors in the future?

funsor/registry.py
```diff
@@ -50,8 +51,11 @@ def register(self, key, *types):
         register = self.registry[key].register
         if self.default:
             objects = (object,) * len(types)
-            if objects != types:
-                register(*objects)(self.default)
+            try:
+                if objects != types and supercedes(types, objects):
+                    register(*objects)(self.default)
+            except TypeError:
+                pass  # mysterious source of ambiguity in Python 3.5 breaks this
```
eb8680, line 58: This is the only significant change in this PR.  It places stricter constraints on method resolution order in multiple dispatch.  If this is reverted, attempting to run the test suite in Python 3.5 will lead to a bunch of strange, nondeterministic and difficult-to-diagnose test failures ultimately caused by patterns not firing when they should.

test/test_gaussian.py
eb8680: The `OrderedDict` changes to test parameters in this PR are required for using `pytest-xdist`.  Otherwise, test collection will fail with multiple workers.

test/test_terms.py
eb8680: `pytest-xdist` seems to require all workers to have identical test cases that appear in the same order.  Some test cases in this file are generated by looping over the keys of `REDUCE_OP_TO_NUMERIC`, which is a dictionary.  If you revert this change and attempt to run the tests in Python 3.5 with multiple workers (`pytest -n auto`), test collection will nondeterministically fail because workers may not agree.

funsor/sum_product.py
```diff
@@ -267,7 +266,7 @@ def shift_funsor(f, t):
     if not lags:
         return naive_sequential_sum_product(sum_op, prod_op, trans, time_var, {})
 
-    period = int(np.lcm.reduce(list(lags)))
+    period = int(reduce(lambda a, b: a * b // gcd(a, b), list(lags)))
```
eb8680, line 269: `np.lcm.reduce` is not available in the latest version of Numpy that supports Python 3.5.

funsor/registry.py
fritzo: Weird. Do you know whether the Type error occurs in `!=`, `supercedes()`, `and`, or `register()()`?

test/test_gaussian.py
fritzo: Thanks for explaining! I'll avoid this pattern in the future.

funsor/registry.py
eb8680: Adding `supercedes` here fixes the test failures, but introduces occasional `TypeError`s which can be disregarded.

PR #335, opened 2020-07-09
Fix value domain for Bernoulli dist

bogiebro:
In the `torch` backend, Bernoulli distributions have Boolean support rather than IntegerInterval support. This adds the necessary check when constructing a Bernoulli distribution.

### DIFF
--- a/funsor/distribution.py
+++ b/funsor/distribution.py
@@ -142,6 +142,8 @@ class Distribution(Funsor, metaclass=DistributionMeta):
         out_shape = instance.event_shape
         if type(instance.support).__name__ == "_IntegerInterval":
             out_dtype = int(instance.support.upper_bound + 1)
+        elif type(instance.support).__name__ == "_Boolean":
+            out_dtype = int(2)
         else:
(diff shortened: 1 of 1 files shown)


bogiebro:
Yeah, it just throws an error. Strange that torch's Bernoulli `log_prob` function expects a real when its Categorical's `log_prob` expects an int

eb8680:
> Yeah, it just throws an error. Strange that torch's Bernoulli log_prob function expects a real when its Categorical's log_prob expects an int

Yes, it's an annoying edge case in `torch.distributions`, presumably there because people want to abuse `Bernoulli.log_prob` by applying it to continuous data to compute cross-entropy.  I'd prefer to avoid having `funsor.distributions` do any unexpected tensor data casting or copying for now, so unfortunately it's probably best to just leave `Bernoulli` as-is and make fixes upstream in PyTorch.  Thanks for noticing this inconsistency, though!

bogiebro:
I see. I guess the relevant issue is this one: https://github.com/pytorch/pytorch/issues/31779

PR #346, opened 2020-08-01
Add a generic enumerate_support method

eb8680:
This PR adds a generic `enumerate_support` method to `funsor.distribution.Distribution` that calls the underlying backend distribution's `enumerate_support` method.


fritzo:
What is the purpose of this PR? Is it used in NumPyro's enumeration?

eb8680:
> What is the purpose of this PR? Is it used in NumPyro's enumeration?

It will be used in enumeration in NumPyro and `pyro.contrib.funsor`.

PR #350, opened 2020-08-15
Support copy, deepcopy, pickle, detach

fritzo:
This aims to add support for Pyro's `detach(-)` function by adding support for `copy.copy()` and `copy.deepcopy()`, which is in turn implemented via new methods `.__copy__()`, `.__deepcopy__()` , and `.__reduce__()`. As a side effect of `.__reduce__()` we get pickle support.

## Tested
- [x] tests that `copy`ing and `deepcopy`ing immutable objects is a no-op, as is the [behavior of copy.copy()](https://github.com/python/cpython/blob/fa5d7251987c70a9c5d58b59a0b36ac9287eaafa/Lib/copy.py#L107) on other immutable objects like tuples and slices
- [x] test that copying a `Tensor` copies the underlying leaf `torch.Tensor`
- [x] test pickling of misc `Funsor`s
- [x] test `torch.save()` of misc `Funsor`s

### DIFF
--- a/funsor/terms.py
+++ b/funsor/terms.py
@@ -1,6 +1,7 @@
 # Copyright Contributors to the Pyro project.
 # SPDX-License-Identifier: Apache-2.0
 
+import copyreg
 import functools
 import itertools
 import math
@@ -172,6 +173,16 @@ def moment_matching(cls, *args):
 
 interpreter.set_interpretation(eager)  # Use eager interpretation by default.
 
+FUNSORS = {}
+
+
+def pickle_funsor(x):
+    return unpickle_funsor, (type(x).__name__, x._ast_values)
+
+
+def unpickle_funsor(name, args):
+    return FUNSORS[name](*args)
+
 
 class FunsorMeta(type):
     """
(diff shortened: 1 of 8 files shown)


test/test_terms.py
```diff
@@ -113,15 +116,46 @@ def test_quote(interp):
...
+
+
+@pytest.mark.parametrize('expr', EXPR_STRINGS)
+def test_pickle(expr):
+    x = eval(expr)
+    f = io.BytesIO()
+    pickle.dump(x, f)
+    f.seek(0)
+    y = pickle.load(f)
+    y is x
```
eb8680, line 155: `assert y is x`?

funsor/terms.py
```diff
@@ -355,6 +355,12 @@ def dtype(self):
     def shape(self):
         return self.output.shape
 
+    def __copy__(self):
+        return self
+
+    def __reduce__(self):
+        return type(self).__origin__, self._ast_values
```
eb8680, line 362: Does this mean a pickled lazy expression may be unpickled and evaluated eagerly?

fritzo: That's right, unpickling will respect under the load-time interpretation, not the dump-time interpretation. While I think the dump-time interpretation is slightly more intuitive, I don't see an easy way to implement that (the naive way would invoke O(size)-many reflect handlers). Thus I think it's fine to require users to load under `reflect`. Maybe there is even a use case for eagerly loading an approximation (say a sketch).

Issue #351, opened 2020-08-16
Replace domains -> types, .inputs+.output -> type hints


fritzo:
This issue proposes to replace funsor `Domain` objects with new subscripted types a la Python 3's `typing` module.

The goal is to make Funsor more Pythonic and to make funsors act more like Python functions with support pointwise operations. In particular many of our design questions about multiple inputs or outputs could be resolved by "following Python". A simple example of more-Pythonic syntax is:
```py
# decorator syntax to read .inputs and .outputs from a type annotation
@to_funsor
def f(x: Reals[3], y: Reals[3,3], t: Reals[()]) -> Reals[3]:
    ...

# explicitly specify type as Callable[..., ...]
g = to_funsor(lambda x: x.sum(), Callable[[Reals[3]], Reals[()])
    ...

h = f + g  # materialized view of lambda x, y, t: f(x, y, t) + g(x)
```

## Tasks
- [ ] Refactor domains to be typing-compatible type objects: `Bint[n]`, `Reals[()]`, `Reals[m,n]`, ...
- [ ] Update interfaces to make use of typing-compatible types (e.g. replace `function` with `to_funsor` of a callable)
- [ ] Possibly promoting `Funsor` subytype themselves to subtypes of `Callable`, as suggested by @eb8680 (though this may warrant a separate issue of larger scope)

eb8680:
This definitely seems like the right way to go.  Some design questions:
- Could we replace `find_domain` with some standard Python type inference library?  Are there existing special array types in NumPy or elsewhere we should use rather than making a custom `Domain`?
- Would some version of this also get us shape polymorphism (as in #214) for free?  If not, does it at least not put up an intrinsic barrier to shape variables ([`typing.Literal` is currently very limited](https://www.python.org/dev/peps/pep-0586/#interactions-with-generics))? 
- How would this interact with surrounding typed Python code?  That is, what is the type signature of `to_funsor`?
- Should we distinguish between a black-box `funsor.function` that creates a custom new `funsor.Function` term and a transparent `funsor.to_funsor` that attempts to extract a funsor expression by evaluating the function on appropriately typed `funsor.Variable`s?

>  Possibly promote Funsor types themselves to subtypes of Callable

[...]

Issue #322, opened 2020-03-25
Support nontrivially shaped bint domains


fritzo:
Motivation: this is required for e.g. `Independent(Categorical)` distributions and for `Function` tensors that input or output batched categoricals.

I believe this is a big change requiring pervasive updates to all shaping logic.

One design question is whether we allow heterogeneous bound as in e.g. heterogeneous Multionmial distributions. Maybe an initial version could implement homogeneous bound, then later a heterogeneous bound.

fritzo:
We may want to separate `Bint` from `Reals`, then generalize `Reals` to `Array` with a `.dtype` property. Some reasons for separating `Bint` from `Array[int]` include:
1. Numpy [broadcasting semantics](https://numpy.org/neps/nep-0021-advanced-indexing.html) can behave differently for ints versus arrays
2. Tuple getitem naturally inputs `Bint` but not for arrays (indeed `Bint` can be seen as dual to `Tuple`)

eb8680:
I'm going to close this as resolved by #417 and #356; if we need to implement specific patterns e.g. `Reduce` over nonscalar `Bint` domains for individual terms we can track them in separate issues.

PR #352, opened 2020-08-16
Refactor domains to be types

fritzo:
Addresses #351 

This refactors so that
- `Domain = type`
- `Bint[n]` and `Reals[m,n]` are type hints (still instances of `Domain`)

This does not appear to cause any overhead, e.g. test_sum_product.py seems to reduce from 95sec to 89 sec.

## Tasks
- [x] refactor `Domain`, `Reals[]`, `Bint[]`
- [x] get tests to pass using shims `reals()` and `bint()`
- [x] mark shims deprecated

for follow-up PRs:
- remove usage of shims from library code
- remove shims (`bint()`, `reals()`, etc.)

### DIFF
--- a/funsor/domains.py
+++ b/funsor/domains.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import operator
+import warnings
 from functools import reduce
 from weakref import WeakValueDictionary
 
@@ -18,7 +19,7 @@ class Domain(type):
         return cls.__name__
 
 
-class RealMeta(Domain):
+class RealType(Domain):
     def __getitem__(cls, shape):
         if not isinstance(shape, tuple):
             shape = (shape,)
@@ -30,7 +31,7 @@ class RealMeta(Domain):
             assert cls is Real
             assert all(isinstance(size, int) and size >= 0 for size in shape)
             name = "Real[{}]".format(",".join(map(str, shape)))
-            result = RealMeta(name, (Real,), {"shape": shape})
+            result = RealType(name, (Real,), {"shape": shape})
             Real._type_cache[shape] = result
         return result
 
@@ -38,34 +39,37 @@ class RealMeta(Domain):
     def num_elements(cls):
         return reduce(operator.mul, cls.shape, 1)
 
-    # SHIM
+    # DEPRECATED
     @property
-    def size(self):
-        raise AssertionError("reals() has no .size")
+    def dtype(self):
+        warnings.warn("domain.dtype is deprecated, "
+                      "use isinstance(domain, RealType) instead",
+                      DeprecationWarning)
+        return "real"
 
 
-class Real(type, metaclass=RealMeta):
+class Real(type, metaclass=RealType):
     """
     Type of a real-valued array with known shape::
 
         Real[()] = Real  # scalar
         Real[8]          # vector of length 8
         Real[3,3]        # 3x3 matrix
+
+    To dispatch on domain type, we recommend either ``@singledispatch``,
+    ``@multipledispatch``, or ``isinstance(domain, RealType)``.
     """
     _type_cache = WeakValueDictionary()
     shape = ()
 
     def __reduce__(self):
-        return RealMeta, (self.shape,)
-
-    # SHIM
-    dtype = "real"
+        return RealType, (self.shape,)
 
 
 Real._type_cache[()] = Real  # Real[()] is Real.
 
 
-class BintMeta(Domain):
+class BintType(Domain):
     def __getitem__(cls, size):
         # in some JAX versions, shape can be np.int64 type
         if get_tracing_state() or funsor.get_backend() == "jax":
@@ -75,7 +79,7 @@ class BintMeta(Domain):
             assert cls is Bint
             assert isinstance(size, int) and size >= 0
             name = "Bint[{}]".format(size)
-            result = BintMeta(name, (Bint,), {"size": size})
+            result = BintType(name, (Bint,), {"size": size})
             Bint._type_cache[size] = result
         return result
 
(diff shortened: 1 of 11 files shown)


fritzo:
@fehiepsi any idea whether the [jax failure](https://travis-ci.com/github/pyro-ppl/funsor/jobs/372473452) is real?

fehiepsi:
@fritzo Did you rerun the job? I can't observe the failure in the link. I observed this error when
+ we trace the `sample_shape` argument of a sampler 
+ or it is a JAX bug

I didn't find something wrong with the current implementation of [binomial sampler](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/distributions/util.py#L146) in NumPyro.

fritzo:
@fehiepsi Thanks for checking. I did rerun the test, and it passed. Before running I captured the failure output (in `details` above). I guess we can just keep this in mind, that JAX samplers can spuriously fail tests.

fritzo:
@eb8680 I'd like to perform this refactoring in two PRs:
1. this PR that changes the underlying types but preserves the old interface in most of funsor
2. a follow-up PR that moderinzes usage in funsor code: `bint(3) -> Bint[3]`, `reals() -> Real`, etc.

The first PR is small, and the second PR will be big but easy to review.

funsor/domains.py
```diff
@@ -2,79 +2,149 @@
...
 
 
-class Domain(namedtuple('Domain', ['shape', 'dtype'])):
-    """
-    An object representing the type and shape of a :class:`Funsor` input or
-    output.
-    """
-    def __new__(cls, shape, dtype):
-        assert isinstance(shape, tuple)
+class Domain(type):
```
eb8680, line 14: Would it be more Pythonic to define `Domain` with `typing.Generic`?
```py
from typing import Generic, Tuple, TypeVar, Union
from typing_extensions import Literal  # backport of typing.Literal in 3.8

S = TypeVar("S", bound=Tuple)
D = TypeVar("D", bound=Union[int, str])

class Domain(Generic[S, D]):
    def __init__(self, shape: S, dtype: D):
        self.shape = shape
        self.dtype = dtype

# version 1, more concise
Bint = Domain[Tuple[()], TypeVar("DB", bound=int)]
Reals = Domain[S, Literal['real']]

# version 2, nicer printing/pickling
class Bint(Domain[Tuple[()], TypeVar("DB", bound=int)]):
    pass

class Reals(Domain[S, Literal['real']]):
    pass

def reals(*shape):
    return Reals[Tuple[tuple(Literal[size] for size in shape)]]

def bint(size):
    return Bint[Literal[size]]
```

fritzo: My longer term vision is for `Domain` to be the type of more general ground values in Funsor including `Tuple[]`, eventually including pytrees. Therefore I think it no longer makes sense to paramterize `Domain` by (`shape`, `dtype`).

Indeed I feel we are already shoe-horning `bint()` and `reals()` into an ill-fitting abstraction. For example to address #322 we will probably need heterogeneous `size` attribute which is currently aliased to `dtype`. Even that alias feels like a hack to me.

eb8680: > My longer term vision is for Domain to be the type of more general ground values in Funsor including Tuple[], eventually including pytrees

I'm definitely on board with this, but it seems like in that vision there's still a need for something like `Reals` and `Bint` (and hence my version of `Domain` above, which we could call `Shape` or `Array` after Numpy?) to represent tensor output types.  We could just have the more general ground value base type be `typing.Any`, or a very broad `TypeVar` like `TypeVar("A", bound=Union[int, str, Shape, Tuple])`?

funsor/domains.py
```diff
@@ -2,79 +2,149 @@
...
+        return "real"
 
 
-def reals(*shape):
+class Real(type, metaclass=RealType):
     """
-    Construct a real domain of given shape.
+    Type of a real-valued array with known shape::
+
+        Real[()] = Real  # scalar
```
eb8680, line 55: Should this equality hold, or should `Real = Real[Tuple[Any, ...]]`, so that it is necessary to write `Real[()]` explicitly to specify a scalar?  The latter is closer to the behavior of `typing.Tuple`.

funsor/domains.py
```diff
@@ -2,79 +2,149 @@
...
+
+class Bint(type, metaclass=BintType):
     """
-    Construct a bounded integer domain of scalar shape.
+    Factory for bounded integer types::
+
+        Bint[5]  # integers ranging in {0,1,2,3,4}
+
+    To dispatch on domain type, we recommend either ``@singledispatch``,
+    ``@multipledispatch``, or ``isinstance(domain, BintType)``.
```
eb8680, line 115: When would we dispatch on domain type with `isinstance` rather than `issubclass`?  More broadly, when will domain types be instantiated, if ever?

fritzo: Currently we dispatch via `if domain.dtype == "real"`. I was thinking that if we are to generalize to other domain types (like `Tuple[...]`) then it would be cleanest to use `@singledispatch`. For lighter-weight uses (in my immediate refactoring plans) I will probably replace `if x.dtype == "real"` with `isinstance()`. Do you have a cleaner suggestion?

funsor/domains.py
fritzo: Good question, I'm undecided. While `Real[()]` is more consistent, `Real` is definitely more readable. I'm leaning towards readability. I think we can change our minds later?

funsor/domains.py
fritzo: Hmm, I guess there's is a future where `Bint` and `Real` are more similar... I'll have to think about this more. I really don't like the way we currently use `domain.dtype` as an alias for `size`, and I'd prefer to pull these concepts apart.

funsor/domains.py
fritzo: To be clear, we would never instantiate domains like `Real`, but we do instantiate domain types like `RealType`.

funsor/domains.py
eb8680: I'm fine with separating `Bint` and `Real`, it's the `BintType`/`RealType` metaclasses in this implementation that I am wondering if we could replace with something more Pythonic based on `typing.Generic` and `typing.Literal`. To clarify, this version of `Reals` simplified from my previous comment inherits directly from `Generic`:
```py
S = TypeVar("S", bound=Tuple[int, ...])

class Reals(Generic[S]):  # subtypes are cons-hashed
    def __init__(self, shape: S):
        self.shape = shape
        # dtype, etc. here

def reals(*shape):  # for illustration, calling reals() in an annotation not compatible with static type checkers?
    return Reals[Tuple[tuple(Literal[size] for size in shape)]]
```
It's more verbose, but conforms more closely to the spirit of `typing`.  Also, I'm not sure pointwise functions annotated with the metaclass-based design would be compatible with static type checkers when checked without the `function` wrapper (no types parametrized directly by values are allowed, regardless of the details of their implementations) - to what extent should that serve as a design constraint?

fritzo: I do like the idea of being more pythonic and trying to use `Generic`. However it remains to be seen whether `Generic` can serve our purpose. Issues include: we want to memoize via `WeakValueDict` rather than `lru_cache` imho; `Generic`'s implementation changes across Python versions, so it may be difficult to override `__class_getitem__` to cast parameters to int while jitting). But you have more experience with typing than I, so let's pair code on it.

> should [compatibility with type checkers] serve as a design constraint?

Yeah good question. I guess the objective of this refactoring is to replace Funsor's bespoke type annotations with more standard annotations and thereby make it easier for new Funsor users to learn and contribute. While that suggests mypy support is a secondary goal, I think it would be appreciated by IDE-based users. Let's see how hard it is, maybe we can add a `mypy` test stage.

Also I'd like to see whether we can roughly follow [NumPy types](https://numpy.org/devdocs/reference/typing.html). Let's discuss tomorrow.

funsor/domains.py
```diff
@@ -1,89 +1,161 @@
...
+
+class RealType(Domain):
+    _type_cache = WeakValueDictionary()
+
+    def __getitem__(cls, shape):
+        if not isinstance(shape, tuple):
+            shape = (shape,)
         # in some JAX versions, shape can be np.int64 type
         if get_tracing_state() or funsor.get_backend() == "jax":
             shape = tuple(map(int, shape))
```
fritzo, line 32: note that if we switch to `typing.Generic` we'll need to override `__getitem__` or `__class_getitem__` to do some casting.

funsor/domains.py
fritzo: Some other challenges with more pythonic type annotations are:
- some of `typing`'s annotations simply disappear during runtime
- there is not yet a standard way to inspect type hints; we may need to roll our own version-agnostic module similar to [typing_inspect](https://github.com/ilevkivskyi/typing_inspect)

funsor/domains.py
fritzo: Resolved by making `Real` an alias of `Reals[()]`

test/test_domains.py
```diff
@@ -21,4 +21,37 @@ def test_pickle(expr):
...
+
+    assert issubclass(Reals, Reals)
+    assert issubclass(Real, Real)
+    assert issubclass(Real, Reals)
+    assert issubclass(Reals[2], Reals)
+    assert issubclass(Reals[2], Reals[2])
+    assert not issubclass(Reals, Real)
+    assert not issubclass(Reals, Real[2])
+    assert not issubclass(Real, Reals[2])
+    assert not issubclass(Real[2], Real)
```
eb8680, line 52: Should `Real[2]` error in favor of `Reals[2]`?

fritzo: Yes, indeed I've already included that fix in #354. Do you mind if we handle that in the follow-up PR?

fritzo:
Thanks for reviewing!

PR #354, opened 2020-08-16
Add minimal type annotation support to funsors

fritzo:
Addresses #351 

This makes Funsors act a little more like functions in three ways:
1. `@funsor.function` can now be applied to an type-annotated function, e.g.

    ```diff
    - @funsor.function(reals(4), (reals(), bint(4)))
    - def max_and_argmax(x):
    + @funsor.function
    + def max_and_argmax(x: Real[4]) -> Tuple[Real, Bint[4]]:
          return torch.max(x, -1)
    ```
2. `@of_type` is replaced by `@symbolic` with nearly the same syntax as `@function`.
3. Adds a property `Funsor.__annotations__` to support `typing.get_type_hints`. In this PR, `__annotations__` is computed from `.inputs`, `.output`; in the future we may want to reverse the dependency?

In particular this PR illuminates a path towards generalizing funsor types from `Union[Real[shape], Bint[size]]` to more general PyTree-like types including nested `Tuple`s.

Regarding @eb8680's [question 4](https://github.com/pyro-ppl/funsor/issues/351#issuecomment-674467968), `to_funsor` does not yet interpret a callable as either `@function` or `@symbolic`. It is an open design question which of these should be called.

## Tested
- [x] unit tests
- [x] updated distribution tests to use the new syntax


test/test_domains.py
```diff
@@ -47,9 +47,9 @@ def test_subclass():
     assert issubclass(Reals[2], Reals)
     assert issubclass(Reals[2], Reals[2])
     assert not issubclass(Reals, Real)
-    assert not issubclass(Reals, Real[2])
+    assert not issubclass(Reals, Reals[2])
```
fritzo, line 50: @eb8680 I don't know how tests passed in #352 before this change, but this change was needed to make tests pass in this PR 🤷

fritzo:
Thanks for reviewing!

PR #361, opened 2020-09-09
Add helper to coerce distributions to funsors

fritzo:
Addresses https://github.com/pyro-ppl/pyro/issues/2618
Blocking https://github.com/pyro-ppl/pyro/pull/2620
pair coded with @eb8680

This adds a helper to coerce raw backend `Distrbution` classes to funsors if any of the `RawDistribution.__init__()` arguments are funsors rather than backend-specific arrays. This is a possible intermediate step to allow partial use of funsors in inference algorithms without the need to use a special `dist` module.

### DIFF
--- a/test/test_distribution.py
+++ b/test/test_distribution.py
@@ -32,7 +32,7 @@ if get_backend() == "torch":
     from funsor.pyro.convert import dist_to_funsor
 
 
-@pyest.mark.xfail(get_backend() == "jax", "flaky test")
+@pytest.mark.xfail(get_backend() == "jax", "flaky test")
 @pytest.mark.parametrize('batch_shape', [(), (5,), (2, 3)], ids=str)
 @pytest.mark.parametrize('eager', [False, True])
 def test_beta_density(batch_shape, eager):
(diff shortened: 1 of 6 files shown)


eb8680 (approved):

test/test_distribution.py
```diff
@@ -32,6 +32,7 @@
     from funsor.pyro.convert import dist_to_funsor
 
 
+@pyest.mark.xfail(get_backend() == "jax", "flaky test")
```
fritzo, line 35: @fehiepsi I've been seeing nondeterministic test failures with the jax backend. They resolve when I rerun the tests

fritzo:
looks like there are lots of shape errors in the JAX backend...

eb8680:
> looks like there are lots of shape errors in the JAX backend...

Do we need to make corresponding changes to the JAX distribution wrappers?

fehiepsi:
Thanks for fixing JAX issues! :+1:

fritzo:
@eb8680 thanks for fixing the JAX issues!

test/test_distribution.py
fehiepsi: Thanks for letting me know! I will keep an eye on this.

Issue #362, opened 2020-09-14
Move math from pyro.ops into funsor


fritzo:
During past projects that have used Pyro, I've tried to publish any reusable math components in [pyro.ops](http://docs.pyro.ai/en/stable/ops.html). In the future I'd like to publish these somewhere in Funsor so they are available in both Pyro and NumPyro. For example I'd like to move Pyro's [HaarTransform](http://docs.pyro.ai/en/stable/distributions.html#pyro.distributions.transforms.HaarTransform) into Funsor, unblocking `numpyro.contrib.epidemiology`.

This issue proposes reorganizing `funsor.ops` to contain all backend-generic low-level ops as follows:
```
funsor/ops/__init__.py  # imports Op and basic ops
funsor/ops/op.py  # defines Op
funsor/ops/builtin.py  # defines basic ops like add, mul, etc
funsor/ops/array.py  # defines things like GetitemOp, cat, stack, reshape
funsor/ops/transforms.py  # defines Haar wavelets, DCT transform
funsor/ops/phylo.py  # defines phylogenetic likelihoods
...
```
This module would be the main interface for ops, and the backend-specific modules like `funsor.torch.ops` would simply register implementations of those interfaces. We can organize `funsor.mybackend.ops` as either a monolithic module or as multiple submodules; it doesn't matter because it isn't an interface.

cc @eb8680 @fehiepsi

eb8680:
This seems like a good idea to me. What other ops should we move from Pyro and NumPyro versus eventual deprecation in favor of existing Funsor code?  What about `pyro.ops.gaussian`/`pyro.ops.gamma_gaussian`?

fritzo:
>What other ops should we move?

Probably most of pyro.ops.tensor_utils, .arrowhead, .hessian, .linalg, .newton. Maybe .integrator, .dual_averaging. But I think we can move as-needed.

> What about pyro.ops.gaussian

I think pyro.ops.gaussian is basically already ported to funsor.gaussian, but it would be good to profile speed before deprecating the Pyro versions.

> pyro.ops.gamma_gaussian?

As needed.

fritzo:
Fixed by #363

PR #363, opened 2020-09-14
Reorganize ops

fritzo:
Addresses #362 

This splits the `funsor.ops` module into three submodules:
- `ops.op` for `Op` base classes and registries
- `ops.builtin` for python builtin operations
- `ops.array` for numpy-like array functions  (mostly authored by @fehiepsi)

These boundaries are best-effort and will likely change a bit in the future, e.g.
- some numeric ops could be moved between `builtin` and `array`
- I'd prefer to avoid using `numpy` in `ops.builtin`
- some op classes could move from `ops.builtin` to `ops.op`

This also slightly changes the `Op` interface to eliminate direct need to use `Dispatcher` for array ops without default behavior; hopefully this will allow increased use of `Op` for pattern matching, but let me know if I've misunderstood.

## Tested
- built docs locally

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -194,7 +194,8 @@ intersphinx_mapping = {
     'numpy': ('http://docs.scipy.org/doc/numpy/', None),
     'torch': ('http://pytorch.org/docs/master/', None),
     'pyro': ('http://docs.pyro.ai/en/stable/', None),
-    'opt_einsum': ('https://optimized-einsum.readthedocs.io/en/stable/', None)
+    'opt_einsum': ('https://optimized-einsum.readthedocs.io/en/stable/', None),
+    'multipledispatch': ('https://multiple-dispatch.readthedocs.io/en/latest/', None),
 }
 
 # @jpchen's hack to get rtd builder to install latest pytorch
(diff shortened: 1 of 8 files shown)


eb8680:
Looks good since most of the changes are just file movement.

> I'd prefer to avoid using numpy in ops.builtin

One thing we could do here is move all Numpy-specific code into an explicit numpy backend submodule (`funsor.numpy`) and replace any numpy-dependent default implementations in `funsor.ops` with errors.  I believe this wouldn't change anything from the user's perspective, since we always start with `funsor.set_backend("numpy")` by default.

fritzo:
> move all Numpy-specific code

I don't mind numpy code in `ops.array`, but I would like to change all of `ops.builtin` to use python's `math` backend for default implementations. I have left that refactoring out of this first PR because this is intended as a simple reorganization, and because I am not sufficiently familiar with JAX to understand the implications of using Python's `math` versus `numpy`. Let's discuss this point on the follow-up PR.

fehiepsi:
I think we can use [these builtin Python operators](https://github.com/google/jax/blob/7569e800146982bd4084599fc2f3cd95daa2a846/jax/core.py#L473) for jax device arrays. It seems that only index/slice assignment/delete are not supported.

eb8680 (approved):

fritzo:
Thanks for reviewing!

PR #364, opened 2020-09-15
Avoid numpy in ops.builtin

fritzo:
Addresses #362

@fehiepsi can you please confirm that the new `math` defaults will still work with jax? Let me know if I'm misunderstanding anything.

### DIFF
--- a/funsor/ops/builtin.py
+++ b/funsor/ops/builtin.py
@@ -1,10 +1,9 @@
 # Copyright Contributors to the Pyro project.
 # SPDX-License-Identifier: Apache-2.0
 
+import math
 import operator
 from numbers import Number
-
-import numpy as np  # TODO remove dependency on numpy in this file
 
 from .op import DISTRIBUTIVE_OPS, PRODUCT_INVERSES, UNITS, Op, TransformOp
 
@@ -121,7 +120,7 @@ def _unary_add(x):
 
 @Op
 def sqrt(x):
-    return np.sqrt(x)
+    return math.sqrt(x)
 
 
 class ExpOp(TransformOp):
@@ -130,7 +129,7 @@ class ExpOp(TransformOp):
 
 @ExpOp
 def exp(x):
-    return np.exp(x)
+    return math.exp(x)
 
 
 @exp.set_log_abs_det_jacobian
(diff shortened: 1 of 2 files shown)


funsor/ops/builtin.py
```diff
@@ -161,12 +157,12 @@ def log_abs_det_jacobian(x, y):
...
 @Op
 def log1p(x):
-    return np.log1p(x)
+    return math.log1p(x)
 
 
 @Op
 def sigmoid(x):
-    return 1 / (1 + np.exp(-x))
+    return 1 / (1 + exp(-x))
```
fritzo, line 165: Design question: should we allow `Op`s within `Op`s (which will increase overhead of multipledispatch) or should we cleanly separate the low-level op layer from higher-level functions built using ops (e.g. @fehiepsi's higher-level gaussian math built on top of ops)?

I am leaning towards the former because it is more flexible. For example we can implement `haar_transform` and `inverse_haar_transform` as ops that use lower-level ops, but then register them as inverses.

funsor/ops/builtin.py
```diff
@@ -144,10 +143,7 @@ class LogOp(TransformOp):
 
 @LogOp
 def log(x):
-    if isinstance(x, bool) or (isinstance(x, np.ndarray) and x.dtype == 'bool'):
```
fritzo, line 147: note I've refactored this conditional into two different patterns: one for default values, one for numpy values (`np.array` and `np.generic`).

fehiepsi:
> the new math defaults will still work with jax

I think so. I don't see why it shouldn't work. In JAX, when the program is compiled, things like `1 * 2 * x` will become `2 * x`, i.e. the results of any computation on non-DeviceArray will be constant.

fehiepsi (approved):
LGTM.

funsor/ops/builtin.py
fritzo: @eb8680 do you have an opinion?

eb8680: I think it's fine to allow `Op`s within `Op`s. It's easier to write backend-agnostic code that way, and the `multipledispatch` overhead should remain low compared to the rest of Funsor since most ops only have a couple of signatures.

Issue #309, opened 2020-01-30
Normalizing flows, transforms as Funsors?


fritzo:
How would we implement normalizing flows / Transforms in Funsor?

As a first guess, each transform would be a funsor with one free variable of the same domain as its output, and we would have a `.inv` attribute like `torch.distributions.Transform.inv` to get the inverse transform. Additionally there would be rules for
- eager substitution `t(x)` which simply calls the forward transform
- inverse followed by eager substitution `t.inv(x)` which calls the inverse transform
- eager substitution of `Delta`s
    ```py
    Delta("x", x, log_density)(x=t("y"))
        => Delta("y", t.inv(x), log_density - t.log_abs_det_jacobian(x))
    ```
- eager reduce of a substitution
    ```py
    f(x=t(y)).reduce(ops.logaddexp, "y")
        => (f(x) + t.inv.log_abs_det_jacobian(x)).reduce(ops.logaddexp, "x")
    ```
- other reduction rules?

I think we could implement this as a single class that took a PyTorch/NumPyro transform object
```py
class Transform(Funsor):  # or maybe Bijector?
    "Funsor backed by a PyTorch transform object."
    def __init__(self, transform):
        self.transform = transform
    @property
    def inv(self):
        return Transform(self.transform.inv)
    def eager_subs(self, subs):
        ...
        return Tensor(self.transform(x.data))
    ...
```

cc @stefanwebb

fritzo:
Note @eb8680 and I sketched one possible direction in https://github.com/pyro-ppl/funsor/tree/lebesgue

eb8680:
I'm going to close this as resolved by #365, #427 and #442. We can open a new issue if we decide to revisit the design.

PR #365, opened 2020-09-20
Support conversion of Transforms and TransformedDistributions to and from Funsors

eb8680:
Addresses #309, #386. ~~Blocked by #387, #388, #389.~~

This PR adds support for converting `Transform`s and `TransformedDistribution`s to and from Funsors, and for sampling and scoring with`funsor.Distribution`s with transforms in their `value` slot.  

~~The latter is achieved with a new term I've called `Lebesgue` for want of a better idea, which computes log-det-jacobians of lazy expressions substituted into its free variable using `funsor.delta.solve`.~~ (**Update**: I'm moving `Lebesgue` to another PR, since it no longer seems necessary for the basic task of converting `TransformedDistribution`s.)

There are several ways this functionality could have been instantiated.  The design choices in this version were made with a view toward locality of changes - the overall distribution API is left unchanged, as is the behavior of `normalize` and `eager`, and I've reused `funsor.delta.solve` even though it should probably be broken up into separate inversion and linearization transformations.

The number of transforms currently supported is very limited, partly because we don't have many transforms wrapped as `TransformOp`s in Funsor and partly because `solve` cannot handle inverting more complex expressions.  I've at least tried to get the two most important higher-order transforms `ComposeTransform` and `InverseTransform` working, so that adding more transforms later will be relatively straightforward.  For ease of review, I have also chosen to leave support for the JAX backend to a followup PR, though it should simply involve copying the additions to `funsor.torch.distributions` verbatim.

Tested:
- Added several `TransformedDistribution` test cases using the infrastructure from #389 

Tasks remaining:
- [x] Add unit tests for density
- [x] Add unit tests for cons-hashing
- [x] Add unit tests for sampling
- [x] Add unit tests for transform conversion
- [x] Get generic conversion tests to pass

Out of scope for this PR:
- Support more transforms
- Add similar capability to the JAX backend
- Add more distribution conversion unit tests, especially interactions with `Independent` - added one `Independent` case, but conversion failing
- Add better documentation for working with transformed distributions and `Lebesgue` - moving `Lebesgue` to another PR


funsor/cnf.py
```diff
@@ -127,6 +128,14 @@ def unscaled_sample(self, sampled_vars, sample_inputs, rng_key=None):
                     terms.append(-gaussian.log_normalizer)
                     terms.append(term.unscaled_sample(greedy_vars, sample_inputs, rng_keys[0]))
                     result = Contraction(self.red_op, self.bin_op, self.reduced_vars, *terms)
+                elif any(isinstance(term, funsor.distribution.Distribution)
+                         and greedy_vars.intersection(term.value.inputs) for term in greedy_terms):
```
fritzo, line 132: nit: I believe `.isdisjoint()` is cheaper than `.intersection`, here and below

funsor/distribution.py
```diff
@@ -109,17 +110,16 @@ def _get_raw_dist(self):
         """
         Internal method for working with underlying distribution attributes
         """
-        if isinstance(self.value, Variable):
-            value_name = self.value.name
-        else:
-            raise NotImplementedError("cannot get raw dist for {}".format(self))
+        value_name = [name for name, domain in self.value.inputs.items()  # TODO is this right?
+                      if domain == self.value.output][0]
```
fritzo, line 114: This looks weird. Can you explain what's going on? When is `value_name != "value"`? The assertion in `self.__init__()` suggests `value_name == "value"` IIUC.

test/test_distribution_generic.py
```diff
@@ -37,6 +40,15 @@ def __getattribute__(self, attr):
     FAKES = _fakes()
 
 
+def normalize_with_subs(cls, *args):
```
fritzo, line 43: add a comment or docstring explaining the purpose of this interpretation

test/test_distribution_generic.py
```diff
@@ -46,15 +58,15 @@ def __getattribute__(self, attr):
...
+    def __init__(self, raw_dist, raw_params, expected_value_domain, xfail=False):
         self.raw_dist = raw_dist
         self.raw_params = raw_params
         self.expected_value_domain = expected_value_domain
         for name, raw_param in self.raw_params:
             if get_backend() != "numpy":
                 # we need direct access to these tensors for gradient tests
                 setattr(self, name, eval(raw_param))
-        TEST_CASES.append(self)
+        TEST_CASES.append(self if not xfail else xfail_param(self, reason="test case not yet supported"))
```
fritzo, line 69: nice! you might even let `xfail` hold the string reason, so you could convert...

test/test_distribution_generic.py
```diff
@@ -329,7 +341,60 @@ def __hash__(self):
...
+        funsor.Real,
+        xfail=get_backend() != "torch",
+    )
+
+    # SigmoidTransform (inversion not working)
+    DistTestCase(
+        "backend_dist.TransformedDistribution(backend_dist.Uniform(low=case.low, high=case.high), [backend_dist.transforms.SigmoidTransform(),])",  # noqa: E501
+        (("low", f"rand({batch_shape})"), ("high", f"2. + rand({batch_shape})")),
+        funsor.Real,
+        xfail=True,  # XXX failure to re-invert ops.sigmoid.inv, which is not atomic
```
fritzo, line 397: ...so you could convert this to
```py
xfail="failure to re-invert ops.sigmoid.inv, which is not atomic"
```

funsor/distribution.py
eb8680: `value_name` can potentially be anything, since `self.value` is a lazy expression. This logic is meant to solve the problem of identifying the value name when `self.value` is not a `Variable` or even has more than one input (which happens when constructing the Funsor version of a `TransformedDistribution`).

The simplest nontrivial example of the latter case would be an affine or power transform where the parameters are `funsor.Tensor`s with nontrivial `.inputs`, although these are not handled in this PR since `funsor.delta.solve` does not yet support inverting such expressions.

The main use for `value_name` in this PR is in `Distribution.unscaled_sample`, which needs to know `value_name` to construct a sample `Delta` with the correct `.inputs`.

fritzo (approved):
Thanks for addressing nits!

PR #368, opened 2020-09-27
Move set_backend(os.environ(...)) from __init__.py to conftest.py

fritzo:
Addresses #362, #207, https://github.com/pyro-ppl/pyro/issues/2630

## Motivation

This PR aims to resolve cyclic dependency issues that are preventing Pyro from depending on `funsor`. The underlying issue is that `funsor.torch` depends on Pyro for distributions and some math helpers, but we would like Pyro to be able to depend on Funsor. This cyclic dependency is an issue because `funsor.torch` calls `import pyro` and then `pyro` calls `import funsor`. This PR resolves the issue by delaying backend-specific module loading until `set_backend()` is called, so that Pyro can `import funsor` without any Pyro dependencies. Then once `pyro` is fully loaded, Pyro code that uses Funsor can call `funsor.set_backend("torch")` as needed, triggering full loading of `funsor.torch`.

## Implementation

This PR moves the backend configuration
```py
set_backend(os.environ.get("FUNSOR_BACKEND", "numpy"))
```
from `funsor/__init__.py` to `test/conftest.py`. After this PR, the `FUNSOR_BACKEND` environment variable is no longer part of the public interface; instead it merely controls pytest testing. Henceforth all examples explicitly call `set_backend("torch")`; in the future we may also make the examples more generic and read backend from an environment variable.

### DIFF
--- a/examples/mixed_hmm/experiment.py
+++ b/examples/mixed_hmm/experiment.py
@@ -53,6 +53,7 @@ def parallel_loss_fn(model, guide, parallel=True):
 
 
 def run_expt(args):
+    funsor.set_backend("torch")
 
     optim = args["optim"]
     lr = args["learnrate"]
--- a/examples/mixed_hmm/experiment.py
+++ b/examples/mixed_hmm/experiment.py
@@ -11,6 +11,7 @@ import pyro
 import pyro.poutine as poutine
 import torch
 
+import funsor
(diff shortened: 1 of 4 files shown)


fehiepsi (approved):
LGTM! Thanks for fixing the circular import issue.

eb8680 (approved):

PR #369, opened 2020-09-27
Refactor MonteCarlo to be a StatefulInterpretation

fritzo:
Pair coded with @eb8680 

@eb8680 I have adapted `StatefuleInterpretation` from https://github.com/pyro-ppl/funsor/tree/adam-interpreter, making it a `namedtuple` as you had originally suggested, and turning on fallback logic by default as you suggested.

This PR:
- [x] Adds a `StatefulInterpretation` base class for interpretations that need to pass instance-specific data such as num_samples and rng_key.
- [x] Adds an `InterpreterStack` and makes the `MonteCarlo` interpretation fall back not to `eager` but the previous interpretation when it was installed.
- [x] Refactors `monte_carlo` to a stateful `MonteCarlo`.
- [x] Fixes the JAX implementation of `MonteCarlo` by tracking `rng_key`
  ~~@fehiepsi can you help me out with this?~~

### DIFF
--- a/funsor/interpreter.py
+++ b/funsor/interpreter.py
@@ -357,7 +357,7 @@ class StatefulInterpretation:
         cls.dispatch = cls.registry.dispatch
 
     def __call__(self, cls, *args):
-        return self.dispatch(cls, *args)(*args)
+        return self.dispatch(cls, *args)(self, *args)
 
     if _DEBUG:
         @classmethod
--- a/funsor/interpreter.py
+++ b/funsor/interpreter.py
@@ -6,7 +6,7 @@ import inspect
 import os
 import re
 import types
-from collections import OrderedDict
+from collections import OrderedDict, namedtuple
 from contextlib import contextmanager
 from functools import singledispatch
 
@@ -111,6 +111,7 @@ def interpretation(new):
     assert callable(new)
     global _INTERPRETATION
     old = _INTERPRETATION
+    new = InterpreterStack(new, old)
     try:
         _INTERPRETATION = new
         yield
@@ -311,6 +312,14 @@ def reinterpret(x):
         return recursion_reinterpret(x)
 
 
+class InterpreterStack(namedtuple("InterpreterStack", ["default", "fallback"])):
+    def __call__(self, cls, *args):
+        for interpreter in self:
+            result = interpreter(cls, *args)
+            if result is not None:
+                return result
+
+
 def dispatched_interpretation(fn):
     """
     Decorator to create a dispatched interpretation function.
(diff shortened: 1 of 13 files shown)


funsor/interpreter.py
```diff
@@ -324,13 +333,50 @@ def dispatched_interpretation(fn):
...
+
+        with interpretation(MyInterpretation(my_param=0.1)):
+            ...
+    """
+    def __init_subclass__(cls):
+        cls.registry = KeyedRegistry(default=lambda *args: None)
+        cls.dispatch = cls.registry.dispatch
+
+    def __call__(self, cls, *args):
+        return self.dispatch(cls, *args)(self, *args)
```
fritzo, line 360: This default implementation of `__call__()` relies on `InterpreterStack` to fall back to the previous interpretation, which is by assumption complete.

funsor/montecarlo.py
```diff
@@ -2,57 +2,34 @@
...
-        monte_carlo.sample_inputs = old
-
-
-@monte_carlo.register(Integrate, Funsor, Funsor, frozenset)
-def monte_carlo_integrate(log_measure, integrand, reduced_vars):
-    # FIXME: how to pass rng_key to here?
-    sample = log_measure.sample(reduced_vars, monte_carlo.sample_inputs)
+@MonteCarlo.register(Integrate, Funsor, Funsor, frozenset)
+def monte_carlo_integrate(state, log_measure, integrand, reduced_vars):
+    # FIXME use state.rng_key to here
```
fritzo, line 25: @fehiepsi can you help me out with the `rng_key` logic here? I have removed the `xfail` from your test case in hopes that we can fix the issue in this PR.

funsor/montecarlo.py
```diff
@@ -2,57 +2,34 @@
...
-    return result
-
+    expressions. This falls back to the previous interpreter in other cases.
 
-# This is a globally configurable parameter to draw multiple samples.
-monte_carlo.sample_inputs = OrderedDict()
+    :param rng_key:
+    """
+    def __init__(self, *, rng_key=None, **sample_inputs):
+        self.rng_key = rng_key
```
fritzo, line 19: @fehiepsi I'm not sure whether `rng_key` should be threaded through the constructor, or whether it should instead be read-and-written in pattern handlers like `monte_carlo_integrate()` below.

funsor/montecarlo.py
fehiepsi: Do we only use `monte_carlo_integrate` one? If so, I think you can simply add `rng_key=state.rng_key` keyword:
```
sample = log_measure.sample(reduced_vars, state.sample_inputs, rng_key=state.rng_key)
```

But if this code is run multiple time, then we need to split the key:
```python
from funsor.util import get_backend
...
    if state.rng_key is not None and get_backend() == "jax":
        import jax

        rng_key, state.rng_key = jax.random.split(state.rng_key)
    sample = log_measure.sample(reduced_vars, state.sample_inputs, rng_key=rng_key)
```

fehiepsi: @fritzo Could you help me replace `if rng_key is not None` at [this line](https://github.com/pyro-ppl/funsor/blob/master/funsor/cnf.py#L89) and 10 lines below by `if rng_key is not None and get_backend() == "jax"`? I feel doing so is safer.

fritzo: Thanks for the help, jax tests now pass :tada:

funsor/interpreter.py
```diff
@@ -324,13 +333,54 @@ def dispatched_interpretation(fn):
     return fn
 
 
+class StatefulInterpretationMeta(type):
```
eb8680, line 336: I guess this is required for Python 3.5?  Once we drop Python 3.5 support, we should remove this and other metaclasses where possible in favor of `__init_subclass__`.

eb8680 (approved):
LGTM, `MonteCarlo` is much cleaner.

funsor/interpreter.py
fritzo: Agreed, I was almost tempted to drop support for Python 3.5 in this PR, but that seemed like overkill.

fritzo:
@eb8680 Thanks for reviewing!
@fehiepsi Thanks for helping!

Issue #373, opened 2020-09-28
Refactor some existing interpreters into StatefulInterpretations


eb8680:
#369 added a `StatefulInterpretation` class for interpretations like `MonteCarlo` that make use of an extra piece of context to evaluate expressions.  We should rewrite other existing stateful interpretations in the same way.

Cases:
- [x] #369 `funsor.monte_carlo.monte_carlo`
- [ ] #375 `funsor.memoize.memoize`
- [ ] `funsor.adjoint.AdjointTape`
- [ ] `funsor.terms.substitute`
- [ ] `funsor.optimizer.optimize` and `apply_optimizer`

eb8680:
Addressed in #429

PR #375, opened 2020-10-02
Convert memoize to StatefulInterpretation

eb8680:
Addresses #373 

This PR converts `funsor.memoize.memoize` into a `StatefulInterpretation` as in #369.  I had to make a couple of changes to the interpreter logic to get this working:
1. Have `interpreter.interpretation` yield its input
2. Add an option to disable interpreter stacking in `interpretation`, which was enabled by default in #369 

The second change was necessary to avoid a memory blowup caused by an interpreter being stacked with itself. It suggests the need for more tinkering with interpreter design to address a number of cases including `memoize` and `AdjointTape` where an interpretation invokes the previously active interpretation and does something with the result.

### DIFF
--- a/funsor/interpreter.py
+++ b/funsor/interpreter.py
@@ -107,14 +107,14 @@ def set_interpretation(new):
 
 
 @contextmanager
-def interpretation(new):
+def interpretation(new, stack=True):
     assert callable(new)
     global _INTERPRETATION
     old = _INTERPRETATION
-    new = InterpreterStack(new, old)
+    new = InterpreterStack(new, old) if stack else new
     try:
         _INTERPRETATION = new
-        yield new.default
+        yield new.default if stack else new
     finally:
         _INTERPRETATION = old
 
--- a/funsor/interpreter.py
+++ b/funsor/interpreter.py
@@ -114,7 +114,7 @@ def interpretation(new):
     new = InterpreterStack(new, old)
     try:
         _INTERPRETATION = new
-        yield
+        yield new.default
     finally:
         _INTERPRETATION = old
 
--- a/funsor/memoize.py
+++ b/funsor/memoize.py
@@ -15,6 +15,6 @@ class Memoize(interpreter.StatefulInterpretation):
         key = (cls,) + tuple(id(arg) if (type(arg).__name__ == "DeviceArray") or not isinstance(arg, Hashable)
                              else arg for arg in args)
         if key not in self.cache:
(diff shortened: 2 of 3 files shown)


funsor/interpreter.py
```diff
@@ -107,14 +107,14 @@ def set_interpretation(new):
...
+def interpretation(new, stack=True):
     assert callable(new)
     global _INTERPRETATION
     old = _INTERPRETATION
-    new = InterpreterStack(new, old)
+    new = InterpreterStack(new, old) if stack else new
     try:
         _INTERPRETATION = new
-        yield
+        yield new.default if stack else new
```
fritzo, line 117: nit: might be slightly more readable as
```py
old = _INTERPRETATION
try:
    _INTERPRETATION = InterpreterStack(new, old) if stack else new
    yield new
finally:
    _INTERPRETATION = old
```

funsor/memoize.py
```diff
@@ -2,26 +2,19 @@
...
-        if key not in cache:
-            cache[key] = cls(*args)
-        return cache[key]
-
-    with interpreter.interpretation(memoize_interpretation):
-        yield cache
+        if key not in self.cache:
+            with interpreter.interpretation(interpreter._INTERPRETATION.fallback, stack=False):
+                self.cache[key] = cls(*args)
+        return self.cache[key]
```
fritzo, line 20: Hmm it kind of looks like the `Interpretation` objects should own their `.fallback` rather than being owned by a `Stack`. E.g. as this PR stands, `_INTERPRETATION` could be a
```py
Stack(inner_interpretation,
      Stack(self,
            outer_interpretation))
```
and it would take two runs through this `__call__()` to hit `fallback`. By contrast if each interpretation owned its `fallback` then we could immediately fallback. I don't fully understand, but let's discuss later today.

funsor/memoize.py
```diff
@@ -2,26 +2,19 @@
...
-    Exploit cons-hashing to do implicit common subexpression elimination
-    """
-    if cache is None:
-        cache = {}
+class Memoize(interpreter.StatefulInterpretation):
 
-    @interpreter.interpretation(interpreter._INTERPRETATION)  # use base
-    def memoize_interpretation(cls, *args):
+    def __init__(self, cache=None):
+        self.cache = cache if cache is not None else {}
```
fritzo, line 12: Nice, this looks like our old `opt_einsum` caching mechanism.

eb8680:
Closing in favor of #429

PR #377, opened 2020-10-03
Add DirichletMultinomial for jax backend

fehiepsi:
~This is blocked until DirichletMultinomial is available in NumPyro.~

~Blocked by https://github.com/pyro-ppl/numpyro/pull/772~

I also enhanced `ops.stack` to support int/float types in jax backend and enabled jax backend in various tests.

PR #378, opened 2020-10-04
Add beta bernoulli conjugacy

fehiepsi:
This PR supports the` collapse` handler by deferring the BernoulliProbs eager rule to Binomial with total_count=1.

Alternatively, I think we can define a BetaBernoulli contraction but I am not sure where we will put the implementation of `log_prob` computation. ~Should we implement a BetaBernoulli distribution in Pyro and do something like DirichletMultinomial?~ **Edit:** I think implementing Pyro's BetaBernoulli is unnecessary. We should find a way to do this directly in funsor.

Blocked by #377.

### DIFF
--- a/funsor/distribution.py
+++ b/funsor/distribution.py
@@ -232,6 +232,7 @@ FUNSOR_DIST_NAMES = [
     ('CategoricalLogits', ('logits',)),
     ('Delta', ('v', 'log_density')),
     ('Dirichlet', ('concentration',)),
+    ('DirichletMultinomial', ('concentration', 'total_count')),
     ('Gamma', ('concentration', 'rate')),
     ('Multinomial', ('total_count', 'probs')),
     ('MultivariateNormal', ('loc', 'scale_tril')),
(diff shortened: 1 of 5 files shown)


funsor/distribution.py
```diff
@@ -549,6 +549,12 @@ def eager_mvn(loc, scale_tril, value):
     return gaussian(**{var: value - loc})
 
 
+def eager_beta_bernoulli(red_op, bin_op, reduced_vars, x, y):
+    backend_dist = import_module(BACKEND_TO_DISTRIBUTIONS_BACKEND[get_backend()])
+    return eager_dirichlet_multinomial(red_op, bin_op, reduced_vars, x,
+                                       backend_dist.Binomial(total_count=1, probs=y.probs, value=y.value))
```
fritzo, line 555: Nice! It should be easy to later adapt this to handle plates by letting `total_count=plate.size`.

fritzo (approved):

fehiepsi:
Yeah, it is in my plan. I first need to implement a Dirichlet-Categorical pattern in funsor to test my understanding.

Thanks for reviewing, @fritzo!

PR #384, opened 2020-10-18
Add DirichletCategorical conjugacy

fehiepsi:
by converting Categorical to Multinomial. Still no plate is supported yet.

### DIFF
--- a/funsor/distribution.py
+++ b/funsor/distribution.py
@@ -556,6 +556,18 @@ def eager_beta_bernoulli(red_op, bin_op, reduced_vars, x, y):
                                        backend_dist.Binomial(total_count=1, probs=y.probs, value=y.value))
 
 
+def eager_dirichlet_categorical(red_op, bin_op, reduced_vars, x, y):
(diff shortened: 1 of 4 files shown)


fritzo:
any idea why tests are failing?

fehiepsi:
It is really strange. In my system, it even passes when I decreased the precision to 1e-7 or 1e-8. My only guess is due to some inconsistent behavior of JAX on travis...

fehiepsi:
@fritzo I added numerical tests for conjugacy. They will verify if backend implementations are correct. In addition, we can test some pairs that are not avaiable in backends such as GammaGamma.

fritzo (approved):
Thanks for adding stronger tests!

fehiepsi:
Thanks for reviewing! This is my first time to actual "use" Integrate and reduce stuffs. They are pretty convenient! Only one comment is arg names such as log_measure is a bit abstract. I know some usage cases of Integrate but can't figure what log measure means. :D

Issue #386, opened 2020-10-26
Full-featured distribution wrappers


eb8680:
The biggest single issue blocking wider use of Funsor in Pyro and NumPyro right now is the incomplete coverage of distributions.  

At a high level, the goal is to be able to perform all distribution operations that appear in any Pyro or NumPyro model (e.g. sampling and scoring) on Funsors directly, where distribution funsors are obtained from using `to_funsor` to automatically convert distributions to Funsors initially and using `funsor.to_data` to automatically convert the final results back to raw PyTorch/JAX objects.

We have gotten pretty far with the generic wrappers in `funsor.distribution` and `funsor.{jax,torch}.distributions`, but finishing the job and achieving full coverage of `pyro.distributions` remains a challenge because of the large distribution API and number of distributions, many small impedance mismatches and legacy design choices in PyTorch (e.g. the data type of `Bernoulli`), and difficulty of programmatic access and automation (e.g. no generic tool for constructing random valid instances of a distribution given a `batch_shape`).

I've tried to collect the remaining Funsor-specific tasks in this issue so that we can better measure progress toward this goal.  We may also need to do additional work upstream in Pyro, NumPyro or PyTorch distributions.

Transforms and `TransformedDistribution`s (some design discussion in #309):
- [ ] #365 Basic conversion of `TransformedDistribution`s to and from `funsor.Distributions`
- [ ] Near-complete coverage of basic `Transform`s in Pyro and NumPyro
- [ ] Wrappers for `ConditionalTransform`s and `TransformedModule`s
- [ ] Direct TFP `Bijector` wrappers

Other basic distribution modifiers:
- [ ] Near-full coverage of basic distributions (easy enough to implement; the difficulty is testing and fixing PyTorch bugs)
- [ ] Find a workaround for casting issues in `Bernoulli` (#348)
- [ ] Converting `funsor.Independent` to and from `Independent` distributions
- [ ] Some Funsor analogue of `Distribution.expand()`
- [ ] Generic support for sampling via `Funsor.sample()` in `funsor.pyro.FunsorDistribution` - done properly, this may eliminate the need for first-class conjugate distribution implementations in backends, e.g. `DirichletMultinomial`
- [ ] Direct wrapping of TFP distributions in the JAX backend - this would probably involve a new subclass `class TFPDistribution(funsor.distribution.Distribution)` with TFP-specific implementations of `_infer_value_domain` and `_infer_param_domain`

Atomic distribution computations implemented in the backend libraries:
[...]

PR #389, opened 2020-10-30
Automate distribution testing

eb8680:
Addresses #386. ~~Blocked by #388.~~

This PR adds a new file `test_distribution_generic.py` that refactors the distribution tests. Here there is one generic test for each distribution method, and all that is needed to test a new distribution is to add a recipe for constructing a random instance to a list of test cases.  This change is necessary if we want to approach full coverage of the dozens of distributions in PyTorch, Pyro/NumPyro and TFP without thousands of lines of manually duplicated testing logic.

I have also had to add a number of small fixes to get the new tests to pass, which I would argue is a sign of the value of these tests - even the first version here is exercising many small distribution API edge cases that would be hard to catch with the current approach.  There are also a couple of new testing utilities (`ops.isnan` and `testing.random_scale_tril`).

If the approach in this PR works, I will delete many of the one-off tests in `test_distribution.py` in a followup PR.

Remaining tasks:
- [x] Get generic gradient test running
- [x] Add enough test cases to reach coverage parity with the tests in `test_distribution.py`
- [x] Get all deterministic tests to pass and avoid nans

Triaged
- ~~Get all Monte Carlo tests to pass~~ - I am skeptical of the viability and usefulness of doing this and have simply disabled and removed these tests for now.


test/test_distribution_generic.py
```diff
@@ -0,0 +1,433 @@
...
+
+@pytest.mark.skipif(not _ENABLE_MC_DIST_TESTS, reason="slow and finicky Monte Carlo tests disabled by default")
+@pytest.mark.parametrize("case", TEST_CASES, ids=str)
+@pytest.mark.parametrize("sample_shape", [(200000,), (400, 400)], ids=str)
+@pytest.mark.parametrize("statistic", [
+    "mean",
+    "variance",
+    pytest.param("entropy", marks=[pytest.mark.skipif(get_backend() == "jax", reason="entropy not implemented")])
+])
+def test_generic_stats_sample(case, statistic, sample_shape):
```
eb8680, line 375: This test comparing Monte Carlo estimates of summary statistics with ground-truth values is slow (especially on the JAX backend) and finicky.  I've disabled it by default but could also remove it entirely - I'm not sure failures are very informative.

test/test_distribution_generic.py
```diff
@@ -0,0 +1,433 @@
...
+@pytest.mark.skipif(not _ENABLE_MC_DIST_TESTS, reason="slow and finicky Monte Carlo tests disabled by default")
+@pytest.mark.skipif(get_backend() != "torch", reason="not working yet on jax")
+@pytest.mark.parametrize("case", TEST_CASES, ids=str)
+@pytest.mark.parametrize("sample_shape", [(200000,), (400, 400)], ids=str)
+@pytest.mark.parametrize("statistic", [
+    "mean",
+    "variance",
+    pytest.param("entropy", marks=[pytest.mark.skipif(get_backend() == "jax", reason="entropy not implemented")])
+])
+def test_generic_grads_sample(case, statistic, sample_shape):
```
eb8680, line 399: Ditto for this test comparing Monte Carlo estimates of gradients of summary statistics wrt parameters versus gradients of ground-truth values - it's slow and finicky, and I'm not sure failures are very informative.  It's disabled by default, but I am open to removing it entirely.

test/test_distribution_generic.py
```diff
@@ -0,0 +1,433 @@
...
+TEST_CASES = []
+
+for batch_shape in [(), (5,), (2, 3)]:
+
+    # BernoulliLogits
+    TEST_CASES += [DistTestCase(
+        "backend_dist.Bernoulli(logits=case.logits)",
+        (("logits", f"rand({batch_shape})"),),
+        funsor.Real,
+    )]
```
eb8680, line 72: After this PR, adding a new distribution to `funsor.distributions` will be as simple as adding it to the list of distributions to be wrapped in `funsor/{backend}/distributions.py` and adding a new test case to this list.

test/test_distribution_generic.py
```diff
@@ -0,0 +1,433 @@
...
+        return hash((self.raw_dist, self.raw_params, self.expected_value_domain))
+
+
+TEST_CASES = []
+
+for batch_shape in [(), (5,), (2, 3)]:
+
+    # BernoulliLogits
+    TEST_CASES += [DistTestCase(
+        "backend_dist.Bernoulli(logits=case.logits)",
```
eb8680, line 69: Note I've chosen to encode test cases with strings of Python code. This should allow us to apply these generic tests even to quite complicated backend distribution expressions, e.g. `TransformedDistribution(Normal(loc, scale).to_event(1), [TanhTransform,]).mask(mask)`

test/test_distribution_generic.py
fritzo: > adding it to the list

Is there a `.register()` command allowing users to dynamically extend the list in their own code? Similar to the dynamic registration mechanisms in `kl_divergence.register()` or `biject_to.register`? Even better, could we automatically register distributions in Funsor the first time they are encountered?

funsor/distribution.py
```diff
@@ -134,23 +134,28 @@ def unscaled_sample(self, sampled_vars, sample_inputs, rng_key=None):
...
         sample_args = (sample_shape,) if get_backend() == "torch" else (rng_key, sample_shape)
         if self.has_rsample:
-            raw_sample = raw_dist.rsample(*sample_args)
+            raw_value = raw_dist.rsample(*sample_args)
         else:
-            raw_sample = ops.detach(raw_dist.sample(*sample_args))
+            raw_value = ops.detach(raw_dist.sample(*sample_args))
 
-        result = funsor.delta.Delta(value.name, Tensor(raw_sample, inputs, value.output.dtype))
+        funsor_value = to_funsor(raw_value, output=value_output, dim_to_name=dim_to_name)
```
eb8680, line 150: I had to update `funsor.distribution.Distribution.unscaled_sample` to use `to_funsor` and `to_data` throughout to get some tests to pass.

test/test_distribution_generic.py
eb8680: > Is there a .register() command allowing users to dynamically extend the list in their own code?

Yes, `funsor.distribution.make_dist` basically plays this role - it takes a backend distribution class and (optionally) some parameter names as input and generates a new `funsor.distribution.Distribution` with generic `eager`, `to_funsor` and `to_data` patterns that should work for most use cases, provided the user has correctly implemented `.arg_constraints` and `.support` in their custom distribution.

eb8680:
@fehiepsi Travis seems to be hanging somewhere in the JAX tests, but I haven't been able to reproduce it locally and I can't figure out where from the logs, although I assume it's related to the distribution tests.  Any idea what might be going on?

eb8680:
One thing I'm noticing is that `test_dirichlet_sample` and the other sampler tests in `test/test_distributions.py` are using huge amounts of memory under the JAX backend both in absolute terms and relative to the PyTorch backend, and may be leaking memory.  This was true before this PR but maybe it got worse?

funsor/torch/distributions.py
```diff
@@ -154,6 +155,19 @@ def _infer_param_domain(cls, name, raw_shape):
...
 
+###########################################################
+# Converting distribution funsors to PyTorch distributions
+###########################################################
+
+@to_data.register(Multinomial)  # noqa: F821
+def multinomial_to_data(funsor_dist, name_to_dim=None):
+    probs = to_data(funsor_dist.probs, name_to_dim)
+    total_count = to_data(funsor_dist.total_count, name_to_dim)
+    if isinstance(total_count, numbers.Number) or len(total_count.shape) == 0:
```
fritzo, line 166: Do we need to worry about `int(total_count)` thwarting PyTorch tracing? Should we preserve scalar `Tensor`s?

test/test_distribution_generic.py
```diff
@@ -0,0 +1,433 @@
...
+class DistTestCase:
+
+    def __init__(self, raw_dist, raw_params, expected_value_domain):
+        self.raw_dist = raw_dist
+        self.raw_params = raw_params
+        self.expected_value_domain = expected_value_domain
+        for name, raw_param in self.raw_params:
+            if get_backend() != "numpy":
+                # we need direct access to these tensors for gradient tests
+                setattr(self, name, eval(raw_param))
```
fritzo, line 54: Why not simply add `TEST_CASES.append(self)` here, so that creating a `DistTestCase` automatically registers it? IIUC you're simply iterating over all instances, right?

fritzo (commented):
...still reviewing test_distribution_generic.py. I think it will be easier to read after adding `TEST_CASES.append(self)` to `.__init__()`.

funsor/torch/distributions.py
eb8680: I made this change because `torch.distributions.Multinomial` raises an error given a tensor `total_count`, even if it is a scalar tensor, and the generic conversion tests were failing.  It's definitely an issue, but there are already lots of JIT issues with scalars and tuple shapes throughout the codebase (encountered while working on `pyro.contrib.funsor`), and my inclination with this PR and further work on distributions is to avoid special-casing to the greatest extent possible even if that means deferring to odd implementation details in the backends.  I think a better fix in this instance would be to allow scalar tensor `total_count` upstream so this change could be reverted.

This is also a good reminder to add some generic JIT tests for distribution wrappers in a followup PR.

fehiepsi:
@eb8680 I think the memory issue is due to "caching" mechanism of jax. It cached the compiled code so that the next time, a function can be executed fast given inputs with the same shape. I guess we can split
```python
pytest -v -n auto --ignore=test/examples --ignore=test/pyro --ignore=test/pyroapi
```
into smaller procedures, e.g.
```python
pytest -v -n auto --ignore=test/examples --ignore=test/pyro --ignore=test/pyroapi --ignore=test/distribution.py
pytest -v -n auto test/distribution.py
```

eb8680:
Would that also explain the run time differences? For comparison, the new sampler smoke tests in this PR were taking ~1s on a single core on my laptop for PyTorch vs ~100s for JAX

fehiepsi:
How many tests there are? If there are 100 tests, then it is normal to me.

eb8680:
> How many tests there are? If there are 100 tests, then it is normal to me.

Around that, yeah

test/test_distribution_generic.py
eb8680: I removed these tests from this PR.

test/test_distribution_generic.py
eb8680: Done.

fritzo (approved):

PR #392, opened 2020-11-07
Support conjugacy for beta plate binomial

fehiepsi:
Pair code with @fritzo and @eb8680.

Current issue:
- ~[ ] We need to address `expand` logic under `plate` handler somewhere: either funsor or pyro/numpyro. The latter is prefered, according to the discussion.~ it seems that no need to do so. We can just simply skip `plate` handler processes a Funsor `msg["fn"]`.

TODO
- [x] Lazily reduce `plate` dim for `conditional.value`.

Currently, there is several FIXME because we want to support the case `Multinomial.value` is a Funsor.

### DIFF
--- a/funsor/torch/distributions.py
+++ b/funsor/torch/distributions.py
@@ -41,7 +41,7 @@ from funsor.distribution import (  # noqa: F401
 )
 from funsor.domains import Real, Reals
 import funsor.ops as ops
-from funsor.tensor import Function, Tensor, dummy_numeric_array
+from funsor.tensor import Tensor, dummy_numeric_array
 from funsor.terms import Binary, Funsor, Reduce, Variable, eager, to_data, to_funsor
 from funsor.util import methodof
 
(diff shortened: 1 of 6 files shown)


funsor/torch/distributions.py
```diff
@@ -275,3 +276,5 @@ def deltadist_to_funsor(pyro_dist, output=None, dim_to_name=None):
     eager_gamma_poisson)
 eager.register(Binary, ops.SubOp, JointDirichletMultinomial, DirichletMultinomial)(  # noqa: F821
     eager_dirichlet_posterior)
+eager.register(Reduce, ops.AddOp, Multinomial[Tensor, Funsor, Funsor], frozenset)(  # noqa: F821
```
fehiepsi, line 279: For `pyro.collapse` purpose, it is enough to use `Multinomial[Tensor, Funsor, Tensor]` here. Then we don't have to worry about `x.value.reduce(...)` issue in `eager_plate_multinomial` above. Is there a simple solution for that and I should go ahead and replace `Funsor` by `Tensor` here in the mean time?

eb8680: >  should go ahead and replace Funsor by Tensor here in the mean time?

Sure, makes sense to me.  I don't think there's a simple solution to the plate membership indication problem, at least not one simple enough to block this PR.

eb8680 (approved):

fritzo (approved):

Issue #293, opened 2019-12-21
Add generalized parallel-scan algorithms for dynamic factor graphs


eb8680:
Tasks:
- [x] #292 Implement parallel-scan `sarkka_bilmes_product` operation
- [ ] Support global variables, either by passing a `global_vars` argument or following the subscripting design of #167 and assuming non-subscripted variables are global by default
- [ ] Handle partial windows
- [ ] Update interface to follow the subscripting design in #167 
- [ ] Change function name to something more descriptive
- [ ] Add Gaussian tests
- [ ] Add `moment_matching` tests
- [ ] Promote `sarkka_bilmes_product` to a first-class Funsor, like `MarkovProduct`
- [ ] Add docs defining the operation and explaining the parallelism-memory tradeoff for both `naive_sarkka_bilmes_product` and `sarkka_bilmes_product`
- [ ] Add some examples
- [ ] Generalize `partial_sum_product` to call `MarkovProduct`

eb8680:
@ordabayevy I copied the work plan from #398 to this issue.  I suggest going ahead with making `modified_partial_sum_product` use the first-class Funsor `funsor.sum_product.MarkovProduct` in place of `sequential_sum_product` and testing that the result is fully compatible with `funsor.optimizer.apply_optimizer`, since I think that may be necessary for your problem, but the other stuff is less important.

ordabayevy:
@eb8680 sounds good. Can you point to an example that tests compatibility with `funsor.optimizer.apply_optimizer`?

eb8680:
> Can you point to an example that tests compatibility with funsor.optimizer.apply_optimizer?

See my PR #400 - that should address this point.

PR #398, opened 2020-11-15
Generalize partial sum product to Markov models

ordabayevy:
This attempts to generalize `partial_sum_product` to support elimination of markov variables following discussion with @eb8680 and @fritzo at pyro-ppl/pyro#2687.

**TODO**
- [x] `modified_partial_sum_product` that eliminates first-order markov variables using parallel-scan algorithm.
- [x] Add more tests
- [ ] <s>Rename `modified_partial_sum_product` or replace `partial_sum_product` by it?</s>

**Tests**
- [x] `_expected_modified_partial_sum_product` unrolls one markov  dim at a time
- [x] `test_modified_partial_sum_product_0`
- [x] `test_modified_partial_sum_product_1`
- [x] `test_modified_partial_sum_product_2`
- [x] `test_modified_partial_sum_product_3`
- [x] `test_modified_partial_sum_product_4`
- [x] `test_modified_partial_sum_product_5`
- [x] `test_modified_partial_sum_product_6`
- [x] `test_modified_partial_sum_product_7`: xfail `x` and `y` markov vars with different ordinals
- [x] `test_modified_partial_sum_product_8`
- [x] `test_modified_partial_sum_product_9`
- [x] `test_modified_partial_sum_product_10`
- [x] `test_modified_partial_sum_product_11`
- [x] `test_modified_partial_sum_product_12`: xfail `w_curr` in time plate but lower ordinal than `y` and `x` markov vars
- [x] `test_modified_partial_sum_product_13`
- [x] `test_modified_partial_sum_product_14`
- [x] `test_modified_partial_sum_product_15`: xfail `y` depends on two time dims `time` and `tones`
- [x] `test_modified_partial_sum_product_16`: `x_prev->y_curr` and `y_prev->x_curr`

**Note**
- Markov sites cannot be within local plates inside (see tests 2 and 10)

**Update**
- Added unrolling of markov vars in local plates to handle tests 2 and 10

**Useful links**
- pyro-ppl/funsor#167
- pyro-ppl/funsor#204
- pyro-ppl/funsor#293 
- pyro-ppl/numpyro#686
- pyro-ppl/funsor#399

### DIFF
--- a/funsor/sum_product.py
+++ b/funsor/sum_product.py
@@ -120,10 +120,10 @@ def modified_partial_sum_product(
     assert isinstance(plates, frozenset)
     assert isinstance(time, str)
     assert isinstance(step, dict)
-    sum_vars = eliminate - plates
     time_plate = frozenset({time})
     plates |= time_plate
-    all_markov_vars = time_plate | frozenset(step.keys()) | frozenset(step.values())
+    sum_vars = eliminate - plates
+    all_markov_vars = frozenset(step.keys()) | frozenset(step.values())
 
     var_to_ordinal = {}
     ordinal_to_factors = defaultdict(list)
@@ -143,14 +143,18 @@ def modified_partial_sum_product(
         leaf_factors = ordinal_to_factors.pop(leaf)
         leaf_reduce_vars = ordinal_to_vars[leaf]
         for (group_factors, group_vars) in _partition(leaf_factors, leaf_reduce_vars):
+            #breakpoint()
             nonmarkov_vars = group_vars - all_markov_vars
             markov_vars = group_vars.intersection(all_markov_vars)
             # contract non markov vars
             f = reduce(prod_op, group_factors).reduce(sum_op, nonmarkov_vars)
-            if markov_vars and var_to_ordinal[time] == leaf:
-                # contract markov vars
+            if markov_vars:
+            # contract markov vars
+                if all_markov_vars.intersection(f.inputs) != markov_vars:
+                    raise ValueError("intractable!")
+                group_step = {k: v for (k, v) in step.items() if v in markov_vars}
                 time_var = Variable(time, f.inputs[time])
-                f = sequential_sum_product(sum_op, prod_op, f, time_var, step)
+                f = sequential_sum_product(sum_op, prod_op, f, time_var, group_step)
                 f = f.reduce(sum_op, markov_vars - time_plate)
 
             remaining_sum_vars = sum_vars.intersection(f.inputs)
@@ -160,45 +164,10 @@ def modified_partial_sum_product(
             else:
                 new_plates = frozenset().union(
                     *(var_to_ordinal[v] for v in remaining_sum_vars))
-                if new_plates != leaf:
-                    f = f.reduce(prod_op, leaf - new_plates - time_plate)
-                    ordinal_to_factors[new_plates].append(f)
-                else:
-                    if markov_vars.isdisjoint(remaining_sum_vars):
-                        raise ValueError("intractable!")
-
-                    # unroll local plates
-                    local_plates = frozenset().union(
-                        *(var_to_ordinal[v] for v in remaining_sum_vars - markov_vars))
-                    factors = [f]
-                    for plate in (leaf - local_plates - time_plate):
-                        unrolled_factors = []
-                        for factor in factors:
-                            slice_factors = [factor(
-                                **{plate: i},
-                                **{var: '{}_{}_{}'.format(var, plate, i) for var in markov_vars},
-                            ) for i in range(factor.inputs[plate].size)]
-                            unrolled_factors.extend(slice_factors)
-
-                        unrolled_step = {}
-                        for k, v in step.items():
-                            if k in markov_vars:
-                                unrolled_step.update(
-                                    {'{}_{}_{}'.format(k, plate, i): '{}_{}_{}'.format(v, plate, i)
-                                     for i in range(f.inputs[plate].size)}
-                                )
-                            else:
-                                unrolled_step[k] = v
-
-                        unrolled_markov_vars = frozenset(
-                            ('{}_{}_{}'.format(var, plate, i) for
-                             var in markov_vars for i in range(f.inputs[plate].size))
-                        )
-                        all_markov_vars = (all_markov_vars - markov_vars) | unrolled_markov_vars
-                        factors = unrolled_factors
-                        step = unrolled_step
-                    ordinal_to_factors[local_plates].extend(factors)
-                    ordinal_to_vars[local_plates] |= unrolled_markov_vars
+                if new_plates == leaf:
+                    raise ValueError("intractable!")
+                f = f.reduce(prod_op, leaf - new_plates - time_plate)
+                ordinal_to_factors[new_plates].append(f)
 
     return results
 
--- a/funsor/sum_product.py
+++ b/funsor/sum_product.py
@@ -134,7 +134,7 @@ def modified_partial_sum_product(
                 for k, v in step.items():
                     group_vars -= frozenset(k) | frozenset(v)
                 f = reduce(prod_op, group_factors).reduce(sum_op, group_vars)
-                f = modified_sequential_sum_product(sum_op, prod_op, f, time, step)
+                f = sequential_sum_product(sum_op, prod_op, f, Variable(time, f.inputs[time]), step)
                 f = f.reduce(sum_op, frozenset(step.values()))
                 f = f.reduce(sum_op, frozenset(step.keys()))
             else:
@@ -240,106 +240,6 @@ def sequential_sum_product(sum_op, prod_op, trans, time, step):
             contracted = Cat(time, (contracted, extra))
         trans = contracted
         duration = (duration + 1) // 2
-    return trans(**{time: 0})
-
-
-def modified_sequential_sum_product(sum_op, prod_op, trans, time, step=frozenset(), window=1):
-    """
-    Parallel-scan algorithm. For :math:`\bigotimes` consists of
-    positional renaming of variables and contraction operations.
-    
-    .. math::
-
-       a_{i} = f^{x_{wi-w+1} \dots x_{wi}}_{x_{wi-2w+1} \dots x_{wi-w}}
-
-       a_{i} \bigotimes a_{j} = f^{x_{wj-w+1} \dots x_{wj}}_{x_{wi-2w+1} \dots x_{wi-w}}
-
-       a_{t} \bigotimes a_{t+1} = f^{x_{wt+1} \dots x_{wt+w}}_{x_{wt-2w+1} \dots x_{wt-w}}
-
-    For `window == 1`:
-
-    .. math::
-
-       a_{i} = f^{x_{i}}_{x_{i-1}}
-
-       a_{i} \bigotimes a_{j} = f^{(x_{i},x_{j-1})}_{x_{i-1}} f^{x_{j}}_{(x_{i},x_{j-1})} = f^{x_{j}}_{x_{i-1}}
-
-       a_{t} \bigotimes a_{t+1} = f^{x_{t}}_{x_{t-1}} f^{x_{t+1}}_{x_{t}} = f^{x_{t+1}}_{x_{t-1}}
-    
-
-    :param ~funsor.ops.AssociativeOp sum_op: A semiring sum operation.
-    :param ~funsor.ops.AssociativeOp prod_op: A semiring product operation.
-    :param ~funsor.terms.Funsor trans: A transition funsor.
-    :param Variable time: The time input dimension.
-    :param dict step: A dict mapping previous variables to current variables.
-        This can contain multiple pairs of prev->curr variable names.
-    """
-    assert isinstance(sum_op, AssociativeOp)
-    assert isinstance(prod_op, AssociativeOp)
-    assert isinstance(trans, Funsor)
-    # for compatibility with sequential_sum_product interface
-    if isinstance(time, Variable):
-        time, duration = time.name, time.output.size
-        if time in trans.inputs:
-            assert duration == trans.inputs[time].size
-    else:
-        duration = trans.inputs[time].size
-    if isinstance(step, dict):
-        step = frozenset(step.items())
-    assert isinstance(time, str)
-    assert isinstance(step, frozenset)
-    assert isinstance(window, int)
-    assert all(len(s) == window + 1 for s in step)
-
-    if duration % window:
-        raise NotImplementedError("TODO handle partial windows")
-
-    # create a new funsor of the length duration // window
-    if window > 1:
-        new_duration = duration // window
-        new_trans = Number(0., 'real')
-        new_step = frozenset()
-        new_to_old = {}
-        for seq in step:
-            new_seq = tuple(''.join(seq[max(0, i+1-window):i+1]) for i in range(window*2))
-            new_step |= frozenset({new_seq})
-            new_to_old.update(zip(new_seq, seq))
-        for w in range(window):
-            old_to_new = {}
-            for values in step:
-                for i, v in enumerate(values):
-                    old_to_new[v] = ''.join(values[max(0, i+w+1-window):i+w+1])
-            new_trans = new_trans + trans(**{time: Slice(time, w, duration, window)}, **old_to_new)
-        duration = new_duration
-        trans = new_trans
-        step = new_step
-
-    # variable renaming
-    drop = tuple("_drop_{}_window_{}".format(i, w) for w in range(window) for i in range(len(step)))
-    prev = tuple(s[w] for w in range(window) for s in step)
-    curr = tuple(s[window+w] for w in range(window) for s in step)
-    prev_to_drop = dict(zip(prev, drop))
-    curr_to_drop = dict(zip(curr, drop))
-    drop = frozenset(prev_to_drop.values())
-
-    # parallel-scan contraction
-    while duration > 1:
-        even_duration = duration // 2 * 2
-        x = trans(**{time: Slice(time, 0, even_duration, 2, duration)}, **curr_to_drop)
-        y = trans(**{time: Slice(time, 1, even_duration, 2, duration)}, **prev_to_drop)
-        contracted = Contraction(sum_op, prod_op, drop, x, y)
-
-        if duration > even_duration:
-            extra = trans(**{time: Slice(time, duration - 1, duration)})
-            contracted = Cat(time, (contracted, extra))
-        trans = contracted
-        duration = (duration + 1) // 2
-    curr_to_drop = frozenset(
-            key for (key, value) in curr_to_drop.items()
-            if not value.endswith('_window_{}'.format(window-1)))
-    trans = trans.reduce(sum_op, curr_to_drop)
-    if window > 1:
-        trans = trans(**new_to_old)
     return trans(**{time: 0})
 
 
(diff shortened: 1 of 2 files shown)


ordabayevy:
Not sure why `test/test_sum_product.py::test_markov_partial_sum_product_multi` is failing. The test passes locally.

eb8680:
@ordabayevy awesome progress!  I'm impressed, especially considering how poorly documented the code in `funsor.sum_product` is 😆 

I think in this first PR we should focus on getting a basic version of the algorithm correct, ignoring the optimizer and window size > 1 as well as any interface changes to `sarkka_bilmes_product`.  Once you have that and we have collected a large number of test cases, we should be able to work our way up to the full version fairly quickly via a sequence of smaller PRs with simpler conceptual transitions between steps. 

To that end, can I suggest something like the following sequence of PRs?

1. A version of `modified_partial_sum_product` that works for time lag 1, and tests that compare it to `partial_sum_product` applied to the unrolled factors.  I think you've mostly got that here, but let's remove `modified_sequential_sum_product` from this first PR in favor of `sequential_sum_product` and focus on adding lots of test cases.
2. Making `modified_partial_sum_product` use `MarkovProduct` and additional tests that verify that it works with `funsor.optimize.apply_optimizer`
3. A version of `modified_partial_sum_product` that uses `sarkka_bilmes_product` directly, and tests that compare it to `partial_sum_product` applied to unrolled factors
4. Tests verifying that the `sarkka_bilmes_product`-based `modified_partial_sum_product` works with `funsor.optimize.apply_optimizer`, and any fixes necessary to make them pass (e.g. a first-class `funsor.Funsor` wrapper for `sarkka_bilmes_product` analogous to `MarkovProduct`)
5. A nice Funsor example or two to show off the new algorithm
6. Optional: Any interface changes to `partial_sum_product` or `sarkka_bilmes_product` (e.g. your `modified_sequential_sum_product` interface)
7. Optional: Additional enhancements/features (e.g. support for partial windows)
8. Optional: removal of less general algorithms implemented in previous steps
9. Integration with Pyro for use with enumeration

eb8680:
> Not sure why `test/test_sum_product.py::test_markov_partial_sum_product_multi` is failing. The test passes locally.

It looks like a floating point precision error caused by unrolling, which can happen occasionally since you're generating test data (i.e. factors) randomly via `funsor.testing.random_tensor`.  Reducing the size of the test data time axis might help, or you can tweak the test tolerance by changing the `atol` and `rtol` arguments to `assert_close` in your tests.

ordabayevy:
@eb8680 thanks for your feedback! Keeping it simple for this PR is a good idea. I have removed `modified_sequential_sum_product` to a separate PR #399 in case it might be useful in the future.

ordabayevy:
Hi @eb8680 . I have added tests `test_partial_sum_product_hmm_example_[1-4]` based on models given in HMM examples and `test_partial_sum_product_hmm_example_[8-10]` are their variations.  Expected values are calculated by `_expected_hmm_example` via unrolling time slices as you mentioned before. Current algorithm doesn't work for tests `_2` and `_10` because markov sites are inside local plates. Seems like these local plates have to be unrolled  before using `sequential_sum_product`(?). I'll try to add unrolling step to see if it works.

test/test_sum_product.py
```diff
@@ -86,21 +87,616 @@ def test_partition(inputs, dims, expected_num_components):
...
+        "time": Bint[duration],
+        "x_prev": Bint[x_dim],
+        "x_curr": Bint[x_dim],
+    }))
+
+    y = random_tensor(OrderedDict({
+        "sequences": Bint[sequences],
+        "time": Bint[duration],
+        "tones": Bint[tones],
+        "y_prev": Bint[y_dim],
```
eb8680, line 174: Note this `y_curr` and the `sequences` plate means this test does not correspond to model 0, although I think it's still a reasonable test case since factors `x` and `y` are completely disconnected.

test/test_sum_product.py
```diff
@@ -86,21 +87,616 @@ def test_partition(inputs, dims, expected_num_components):
...
+        "sequences": Bint[sequences],
+        "time": Bint[duration],
+        "x_prev": Bint[x_dim],
+        "x_curr": Bint[x_dim],
+    }))
+
+    y = random_tensor(OrderedDict({
+        "sequences": Bint[sequences],
+        "time": Bint[duration],
+        "tones": Bint[tones],
```
eb8680, line 223: When are the plates `tones` and `sequences` being eliminated by `modified_partial_sum_product`? It should be before and after `time` respectively, rather than both before or both after.

test/test_sum_product.py
```diff
@@ -86,21 +87,616 @@ def test_partition(inputs, dims, expected_num_components):
...
+        "x_curr": Bint[x_dim],
+        "y_prev": Bint[y_dim],
+        "y_curr": Bint[y_dim],
```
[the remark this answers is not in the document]
ordabayevy: Yeah, I kept `y_prev` and `y_curr` and unobserved in order to consider more general cases. As I mentioned above we might be able to generalize to these cases by using `sequential_sum`.

test/test_sum_product.py
ordabayevy: For different combinations of `vars1` and `vars2` it follows that order to eliminate plates. Second case is when `tones` removed before `time` and third case is `sequences` eliminated after `time`.

funsor/sum_product.py
```diff
@@ -93,6 +93,91 @@ def partial_sum_product(sum_op, prod_op, factors, eliminate=frozenset(), plates=
...
+                    raise ValueError("intractable!")
+                for v in sum_vars.intersection(f.inputs):
+                    if time.issubset(var_to_ordinal[v]) and var_to_ordinal[v] < leaf:
+                        raise ValueError("intractable!")
+                # TODO: allow sequential_sum_product accept time as frozenset
+                time = next(iter(time))
+                time_var = Variable(time, f.inputs[time])
+                group_step = {k: v for (k, v) in plate_to_step[time].items() if v in markov_vars}
+                f = sequential_sum_product(sum_op, prod_op, f, time_var, group_step)
+                f = f.reduce(sum_op, frozenset(group_step.keys()) | frozenset(group_step.values()))
```
ordabayevy, line 164: @eb8680 @fritzo thanks for helping with the code! Here I made some changes to its structure. `sequential_sum_product` is in a separate logic block (and that's the only extra code compared to `partial_sum_product`):
- since `sequential_sum_product` is called at most only once I feel like there is no need for yet another function `_markov_product`
- now it is easier to test for intractable cases (line 157)
- additionally we can easily check that there is at most one markov plate per branch (line 154)

WDYT?

eb8680: > Here I made some changes to its structure. sequential_sum_product is in a separate logic block (and that's the only extra code compared to partial_sum_product)

Makes sense to me.  I do wonder if there's a simpler way to formulate the intractability conditions that unifies them into a single condition, but I guess we don't need to figure that out now.

funsor/sum_product.py
```diff
@@ -93,6 +93,91 @@ def partial_sum_product(sum_op, prod_op, factors, eliminate=frozenset(), plates=
...
+            # this is the only addinional part to partial_sum_product
+            markov_vars = group_vars.intersection(markov_sum_vars)
+            if markov_vars:
+                time = markov_prod_vars.intersection(f.inputs)
+                if not len(time) == 1:
+                    raise ValueError("intractable!")
+                for v in sum_vars.intersection(f.inputs):
+                    if time.issubset(var_to_ordinal[v]) and var_to_ordinal[v] < leaf:
+                        raise ValueError("intractable!")
+                # TODO: allow sequential_sum_product accept time as frozenset
```
eb8680, line 159: Can you explain this TODO?

test/test_sum_product.py
```diff
@@ -86,21 +87,774 @@ def test_partition(inputs, dims, expected_num_components):
...
+     frozenset()),
+])
+@pytest.mark.parametrize('x_dim,time', [
+    (3, 1), (1, 5), (3, 5),
+])
+@pytest.mark.parametrize('sum_op,prod_op', [(ops.logaddexp, ops.add), (ops.add, ops.mul)])
+def test_modified_partial_sum_product_0(sum_op, prod_op, vars1, vars2,
+                                        x_dim, time):
+
+    f1 = random_tensor(OrderedDict({}))
```
eb8680, line 174: Thanks for renaming these

funsor/sum_product.py
```diff
@@ -93,6 +93,91 @@ def partial_sum_product(sum_op, prod_op, factors, eliminate=frozenset(), plates=
     return results
 
 
+def modified_partial_sum_product(
+        sum_op, prod_op, factors, eliminate=frozenset(),
+        plate_to_step=dict()):
+    """
+    Modified partial sum-product that supports contraction of markov sites.
```
eb8680, line 100: Make the docstring a bit more informative, e.g. "Generalization of the tensor variable elimination algorithm of `funsor.sum_product.partial_sum_product` to handle markov dimensions in addition to plate dimensions.  Markov dimensions are eliminated efficiently using the parallel-scan algorithm in `funsor.sum_product.sequential_sum_product`."

test/test_sum_product.py
```diff
@@ -86,21 +87,774 @@ def test_partition(inputs, dims, expected_num_components):
...
+    factors1 = impl(sum_op, prod_op, factors, vars1, plates)
+    factors2 = impl(sum_op, prod_op, factors1, vars2, plates)
     actual = reduce(prod_op, factors2)
 
-    expected = sum_product(sum_op, prod_op, factors, vars1 | vars2, plates)
+    expected = sum_product(sum_op, prod_op, factors, vars1 | vars2, frozenset(plates))
     assert_close(actual, expected)
 
 
+def _expected_hmm_example(sum_op, prod_op, factors, plate_to_step,
```
eb8680, line 113: Have you verified manually that the unrolling logic in this function produces the right factors for a few of the test cases? It looks correct to me but it's still somewhat complicated and all of the new tests depend on it.

eb8680 (commented):
The algorithm and tests mostly look good, except for a few nits. Can you also add a couple additional test cases?

1. Like `test_modified_partial_sum_product_5`, but with a global sum variable coupling the two time dimensions
2. Something exercising the `len(time) == 1` intractability case

funsor/sum_product.py
ordabayevy: This is irrelevant to this PR. I removed it.
(I think it would be simpler to pass just the name of the time dim to the `sequential_sum_product` instead of `time_var`)

test/test_sum_product.py
ordabayevy: I reworked the unrolling logic (it unrolls one markov dim at a time for all factors now). [Here](https://docs.google.com/presentation/d/1K2aG9reNF0KWehGzxI8b7SsMqW-FaOPY1tfdb2r2vnA/edit?usp=sharing) I have manually unrolled markov dims for tests 11 and 14 and compared to the unrolled factors by this function. It is complicated and would be nice if someone else also verified it.

ordabayevy (commented):
Sorry for the late reply. New added tests:

1. `test_modified_partial_sum_product_13` with a global sum variable coupling the two time dimensions
2. This one is tricky. I came up with `test_modified_partial_sum_product_14` but then realized that it actually might be tractable. With some changes in the code of the algorithm it seems to pass this test. I also have checked its unrolling manually [here]( actually).

The added docstring is as you typed it.

test/test_sum_product.py
```diff
@@ -86,21 +87,1006 @@ def test_partition(inputs, dims, expected_num_components):
...
+        "x_curr": Bint[x_dim],
+        "y_prev": Bint[y_dim],
+        "y_curr": Bint[y_dim],
+    }))
+
+    factors = [f, g, h]
+    plate_to_step = dict({
+        "sequences": {},
+        "time": {"x_prev": "x_curr"},
+        "tones": {"y_prev": "y_curr"}
```
eb8680, line 1062: I guess the reason this doesn't fail as intractable is that you haven't included `y` in the `time` plate, so it's treating `tones` as nested entirely within a single step of `time`.

test/test_sum_product.py
```diff
@@ -86,21 +87,1006 @@ def test_partition(inputs, dims, expected_num_components):
...
+     frozenset({"sequences"})),
+])
+@pytest.mark.parametrize('w_dim,x_dim, y_dim, sequences, days, weeks, tones', [
+    (3, 2, 3, 2, 5, 4, 3),
+    (3, 1, 3, 2, 5, 4, 3),
+    (3, 2, 1, 2, 5, 4, 3),
+    (3, 2, 3, 2, 1, 4, 3),
+])
+@pytest.mark.parametrize('sum_op,prod_op', [(ops.logaddexp, ops.add), (ops.add, ops.mul)])
+def test_modified_partial_sum_product_13(sum_op, prod_op, vars1, vars2,
```
eb8680, line 965: `test_modified_partial_sum_product_13` looks good, thanks.

test/test_sum_product.py
```diff
@@ -110,58 +110,87 @@ def test_partial_sum_product(impl, sum_op, prod_op, inputs, plates, vars1, vars2
     assert_close(actual, expected)
 
 
-def _expected_hmm_example(sum_op, prod_op, factors, plate_to_step,
-                          global_vars, local_var_dict):
+def _expected_modified_partial_sum_product(
+        sum_op, prod_op, factors, plate_to_step, global_vars,
+        local_var_dict, local_markov_var_dict, markov_to_step):
```
eb8680, line 115: Can you explain the function of this new `local_markov_var_dict` argument? It doesn't seem to be used in any of the tests except `test_modified_partial_sum_product_14`.

test/test_sum_product.py
eb8680: For this to fail as intractable, you might have to include two markov variables `y_prev_tones` and `y_prev_time` for `y` in `h`, one for each time variable, and put them in `plate_to_step` appropriately:
```py
plate_to_step = dict({
    "sequences": {},
    "time": {"x_prev": "x_curr", "y_time_prev": "y_curr"},
    "tones": {"y_prev": "y_curr", "y_tones_prev": "y_curr"}
})
```

test/test_sum_product.py
```diff
@@ -110,58 +110,87 @@ def test_partial_sum_product(impl, sum_op, prod_op, inputs, plates, vars1, vars2
...
+        for factor in factors:
+            if markov_plate in factor.inputs:
+                slice_factors = [factor(
+                    **{markov_plate: i},
+                    **{var: '{}_{}_{}'.format(var, markov_plate, i+1)
+                       for var in local_vars},
+                    **{'{}_curr'.format(var): '{}_{}_{}_curr'.format(var, markov_plate, i+1)
+                       for var in local_markov_vars},
+                    **{'{}_prev'.format(var): '{}_{}_{}_prev'.format(var, markov_plate, i+1)
+                       for var in local_markov_vars},
```
eb8680, line 176: To clarify, is the purpose of these two new substitutions to handle cases like the current, tractable version of `test_modified_partial_sum_product_14` where one time variable is completely nested within a single step of another?

eb8680:
OK, I think I understand the unrolling logic now modulo a couple of questions, and it looks correct to me along with your hand-unrolled examples, at least for the test cases in this PR.  

Perhaps in a future PR It would be a good idea to provide a general `funsor.sum_product.unroll` implementation that converts a `modified_partial_sum_product` problem instance (i.e. the arguments to `modified_partial_sum_product`) into a sum-product problem instance with no time variables or plates.

I'm willing to merge this PR once we have an xfailing test covering `len(time) > 1`.

test/test_sum_product.py
ordabayevy: Yep, that's right.

funsor/sum_product.py
```diff
@@ -93,6 +93,98 @@ def partial_sum_product(sum_op, prod_op, factors, eliminate=frozenset(), plates=
...
+    plates = frozenset(plate_to_step.keys())
+    sum_vars = eliminate - plates
+    prod_vars = eliminate.intersection(plates)
+    markov_sum_vars = frozenset()
+    for step in plate_to_step.values():
+        markov_sum_vars |= frozenset(step.keys()) | frozenset(step.values())
+    markov_sum_vars &= sum_vars
+    markov_prod_vars = frozenset(k for k, v in plate_to_step.items() if v and k in eliminate)
+    markov_sum_to_prod = defaultdict(set) 
+    for markov_prod in markov_prod_vars:
```
ordabayevy, line 131: Collects all time dims that a markov var depends on in a set. In line 162 raise intractable error if the set size is not equal 1.

funsor/sum_product.py
```diff
@@ -93,6 +93,98 @@ def partial_sum_product(sum_op, prod_op, factors, eliminate=frozenset(), plates=
...
+    ordinal_to_vars = defaultdict(set)
+    for var, ordinal in var_to_ordinal.items():
+        ordinal_to_vars[ordinal].add(var)
+
+    results = []
+    while ordinal_to_factors:
+        leaf = max(ordinal_to_factors, key=len)
+        leaf_factors = ordinal_to_factors.pop(leaf)
+        leaf_reduce_vars = ordinal_to_vars[leaf]
+        for (group_factors, group_vars) in _partition(leaf_factors, leaf_reduce_vars | markov_prod_vars):
```
ordabayevy, line 153: `| markov_prod_vars` is added to `_partition` for `test_modified_partial_sum_product_16` to pass. Otherwise leaf factors (f2 and f3) are partitioned into two different groups.

test/test_sum_product.py
ordabayevy: See `test_modified_partial_sum_product_15`

ordabayevy (commented):
Two more tests:

1. test 15 intractable because `y` depends on two time dims `time` and `tones`
2. test 16 where `y_prev->x_curr` and `x_prev->y_curr`

eb8680:
Last 2 tests look good. One final task before merging: can you run `test_sum_product.py` locally with all of your new tests under the JAX backend (`FUNSOR_BACKEND=jax pytest test_sum_product.py ...`)? 

We don't include those on Travis because they're so slow and memory-intensive, so you'll need to run them locally to catch any potential bugs.

ordabayevy:
@eb8680 passed `FUNSOR_BACKEND=jax pytest test_sum_product.py -n auto`:

<details>
```
............................................................................................. [  3%]
............................................................................................. [  7%]
............................................................................................. [ 11%]
............................................................................................. [ 14%]
............................................................................................. [ 18%]
............................................................................................. [ 22%]
............................................................................................. [ 25%]
............................................................................................. [ 29%]
............................................................................................. [ 33%]
.............................................................................................. [ 37%]
............................................................................................. [ 40%]
............................................................................................. [ 44%]
............................................................................................. [ 48%]
............................................................................................. [ 51%]
............................................................................................. [ 55%]
............................................................................................. [ 59%]
............................................................................................. [ 62%]
............................................................................................. [ 66%]
............................x...x....x.x..x....x.............x.x..x.......................... [ 70%]
...........................................................x...................x....x..x..x.. [ 74%]
.x.xxxx.......xx.x.....xxxxx.xxxxxx.................x...xx..x......x...x...x..x...........xx. [ 77%]
.......x..............x...x.....x.x.....x.......x.....x...x.xxxx.....x.x.x.x...x.x.......xxxx [ 81%]
xxx..x.xxx...x.xxx...xx.x...x..x.x.x.x...x.x.xx.........x..x......................x......x... [ 85%]
......x.....x...xx.x.........x..x...x.x.x..x...xx.xxxx.x.x..xx.xx..xx.xx..................x.. [ 88%]
.x..........x..........x...x..x..x...xx...xx.......x...x.......xx..x...xxx..x..x.xxxx.xxxxxxx [ 92%]
xxx..xx.xxx.......x.xxx..xx..x..x..xxxxxx..x......x..x.....x......x....x..........x..x....... [ 96%]
............x.....x............x..........................x...................x.........x.... [ 99%]
.                                                                                             [100%]
=========================== 2325 passed, 188 xfailed in 258.88s (0:04:18) ===========================
```
</details>

eb8680 (approved):
Looks good to me.  Nice work!

ordabayevy:
Yay! Thanks for all the help!

PR #400, opened 2020-11-29
Update modified_partial_sum_product to use MarkovProduct and add test

eb8680:
Addresses #293. ~~Blocked by #401.~~

This PR replaces `sequential_sum_product` in @ordabayevy's `funsor.sum_product.modified_partial_sum_product` from #398 with `MarkovProduct` and adds a test verifying that the result is compatible with `funsor.optimizer.apply_optimizer`.

### DIFF
--- a/test/test_sum_product.py
+++ b/test/test_sum_product.py
@@ -25,7 +25,7 @@ from funsor.sum_product import (
     sum_product
 )
 from funsor.tensor import Tensor, get_default_prototype
-from funsor.terms import Variable, eager_or_die, moment_matching, reflect
+from funsor.terms import Variable, eager, eager_or_die, lazy, moment_matching, reflect
 from funsor.testing import assert_close, random_gaussian, random_tensor
 from funsor.util import get_backend
 
@@ -182,9 +182,10 @@ def _expected_modified_partial_sum_product(
                 unrolled_factors.append(factor)
         factors = unrolled_factors
 
-    expected = sum_product(sum_op, prod_op, factors, reduce_vars, plates)
+    with interpretation(lazy):
+        expected = sum_product(sum_op, prod_op, factors, reduce_vars, plates)
 
-    return expected
+    return apply_optimizer(expected)
 
 
 @pytest.mark.parametrize('vars1,vars2', [
@@ -657,6 +658,7 @@ def test_modified_partial_sum_product_8(sum_op, prod_op, vars1, vars2,
     assert_close(actual, expected, atol=5e-4, rtol=5e-4)
 
 
+@pytest.mark.parametrize("use_lazy", [False, True])
 @pytest.mark.parametrize('vars1,vars2', [
     (frozenset(),
      frozenset({"sequences", "time", "w_prev", "w_curr", "x_prev", "x_curr", "tones", "y_curr"})),
@@ -671,7 +673,7 @@ def test_modified_partial_sum_product_8(sum_op, prod_op, vars1, vars2,
     (3, 2, 3, 2, 5, 4), (3, 1, 3, 2, 5, 4), (3, 2, 1, 2, 5, 4), (3, 2, 3, 2, 1, 4),
 ])
 @pytest.mark.parametrize('sum_op,prod_op', [(ops.logaddexp, ops.add), (ops.add, ops.mul)])
-def test_modified_partial_sum_product_9(sum_op, prod_op, vars1, vars2,
+def test_modified_partial_sum_product_9(use_lazy, sum_op, prod_op, vars1, vars2,
                                         w_dim, x_dim, y_dim, sequences, time, tones):
 
     f1 = random_tensor(OrderedDict({}))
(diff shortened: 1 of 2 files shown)


eb8680:
@ordabayevy can you review this PR?

ordabayevy (commented):
Looks good to me. What are you trying to test with `test_modified_partial_sum_product_17`?  Is `apply_optimizer` supposed to be tested only on some of the tests and not all of them?

eb8680:
> What are you trying to test with test_modified_partial_sum_product_17? Is apply_optimizer supposed to be tested only on some of the tests and not all of them?

`funsor.optimizer.optimize` cheaply evaluates the finitary sum-product operations `reduce(prod_op, group_factors).reduce(sum_op, group_vars)` in each iteration of `modified_partial_sum_product`. I believe it's trivial for most of the other tests, since there's only one or two factors per ordinal, so I added a new test case `test_modified_partial_sum_product_17` with more factors in a single ordinal.

eb8680:
The CI failure is unrelated to the changes in this PR and appears to be caused instead by the change enabling validation by default in Pyro: https://github.com/pyro-ppl/pyro/pull/2701

I'll put up a separate PR with a fix.

test/test_sum_product.py
```diff
@@ -657,6 +658,7 @@ def test_modified_partial_sum_product_8(sum_op, prod_op, vars1, vars2,
     assert_close(actual, expected, atol=5e-4, rtol=5e-4)
 
 
+@pytest.mark.parametrize("use_lazy", [False, True])
```
fritzo, line 661: nit: `, ids=["eager", "lazy"]` would clarify test output

test/test_sum_product.py
```diff
@@ -803,6 +807,7 @@ def test_modified_partial_sum_product_10(sum_op, prod_op, vars1, vars2,
     assert_close(actual, expected, atol=5e-4, rtol=5e-4)
 
 
+@pytest.mark.parametrize('use_lazy', [False, True])
```
fritzo, line 810: ditto: `, ids=["eager", "lazy"]`

test/test_sum_product.py
```diff
@@ -941,6 +948,7 @@ def test_modified_partial_sum_product_12(sum_op, prod_op, vars1, vars2,
         reduce(prod_op, factors2)
 
 
+@pytest.mark.parametrize('use_lazy', [False, True])
```
fritzo, line 951: ditto

test/test_sum_product.py
```diff
@@ -1182,6 +1192,74 @@ def test_modified_partial_sum_product_16(sum_op, prod_op, vars1, vars2,
     assert_close(actual, expected, atol=5e-4, rtol=5e-4)
 
 
+@pytest.mark.parametrize('use_lazy', [True, False])
```
fritzo, line 1195: ditto

test/test_sum_product.py
```diff
@@ -1182,6 +1192,74 @@ def test_modified_partial_sum_product_16(sum_op, prod_op, vars1, vars2,
...
+    with interpretation(lazy if use_lazy else eager):
+        factors1 = modified_partial_sum_product(sum_op, prod_op, factors, vars1, plate_to_step)
+        factors2 = modified_partial_sum_product(sum_op, prod_op, factors1, vars2, plate_to_step)
+        actual = reduce(prod_op, factors2)
+    actual = apply_optimizer(actual)
+
+    local_var_dict = {"time": frozenset({"y_curr", "z0", "z1", "z2"})}
+    local_markov_var_dict = {"time": frozenset()}
+    global_vars = frozenset()
+    markov_to_step = dict({"time": {"x"}})
```
fritzo, line 1254: Should this be `{"time": {"x_prev": "x_curr"}}` or `{"time": {}}`? I wouldn't expect `Dict[str, Set[str]]` to be a supported datatype for this arg.

test/test_sum_product.py
```diff
@@ -1182,6 +1192,74 @@ def test_modified_partial_sum_product_16(sum_op, prod_op, vars1, vars2,
...
+
+    f3_3 = random_tensor(OrderedDict({
+        "time": Bint[time],
+        "x_curr": Bint[x_dim],
+        "y_curr": Bint[y_dim],
+        "z2": Bint[z_dim],
+    }))
+
+    factors = [f1, f2, f3_1, f3_2, f3_3]
+    plate_to_step = dict({"time": {"x_prev": "x_curr"}})
```
fritzo, line 1243: nit: why all the superfluous `dict({...})` wrappers?

fritzo (commented):
LGTM modulo nits.

Also I sure have benefitted from the ascii art summaries in pyro's test_enumeration.py. It seems like drudgery until 4 months later when you pleasantly find yourself not needing to parse inscrutable tensor dimensions.

fritzo (approved):

test/test_sum_product.py
eb8680: I'm not sure, although note that this argument is only passed to the test helper `_expected_modified_partial_sum_product`.

Probably the right thing to do here is to implement a general `funsor.sum_product.unroll` function that takes the same arguments as `funsor.sum_product.{modified_}partial_sum_product` and returns a new set of arguments specifying an equivalent sum-product problem with no plates or Markov dimensions and use that to replace the custom unrolling logic in this test helper.

ordabayevy: `_expected_modified_partial_sum_product` assumes that markov vars have a `_prev` and `_curr` name convention and converts `x` from `{"time": {"x"}}` to a pair of `x_prev` and `x_curr`. It was simpler to quickly write unrolling logic for nested markov dims this way. I agree that there should be a general and  better documented `funsor.sum_product.unroll` function.

fritzo: Thanks for explaining, that makes sense.

fritzo (approved):
LGTM, thanks for addressing nits!

PR #402, opened 2020-12-01
Update distribution shape inference to handle independent dims

eb8680:
Addresses #386, https://github.com/pyro-ppl/pyro/pull/2702

Consider the following snippet of Pyro code simplified from a test in https://github.com/pyro-ppl/pyro/pull/2702 :
```py
with pyro.plate("data", T, dim=-1):
    y = pyro.sample("y", dist.Normal(torch.ones(3), 1.).to_event(1))
    pyro.sample("z", dist.Normal(y, 1.).to_event(1), obs=data)
```
Attempting to wrap this in a `poutine.collapse()` context fails when `dist.Normal(y, 1)` is incorrectly converted to a Funsor. In particular, there is a mismatch between `y.output`, which is `Reals[3]` and the `.output` value `Real` expected for the `loc` param of `funsor.torch.distributions.Normal` given by `funsor.distribution.Distribution._infer_value_domain`.

In cases like the above, however, it is always possible to infer the correct parameter and value `.output` shapes generically using the broadcasting logic in the underlying backend distribution.  This PR implements this behavior in `funsor.distribution.DistributionMeta.__call__` and `distribution_to_data`.  

As a result, it is possible after this PR to represent `Independent` distributions without an intermediary `funsor.Independent` by passing a `Variable` with extra output dimensions as the `value`:
```py
# diagonal multivariate normal in pyro:
dist.Normal(zeros(3), 1.).to_event(1)

# funsor equivalent, after this PR - note Reals[3] for x
Normal(loc=zeros(3), scale=1., value=Variable("x", Reals[3]))
```
This should considerably simplify pattern-matching over `Independent` distributions.

To make this work, converting a `funsor.distribution.Distribution` to data via `funsor.to_data` will have to include a step handling nontrivial event shapes by calling `.to_event` and unsqueezing parameters.

More ambitiously, we could also attempt to handle `to_funsor` conversion of backend `Independent` distributions by substituting an appropriately shaped `Variable` for their value rather than resorting to lazy application of `funsor.terms.Independent`, which would make pattern-matching for `collapse`ing multivariate distributions much easier, but it's not clear yet whether this can be done generically, especially in the presence of transforms.  I have not attempted to do this in this PR.

Tasks:
- [x] Update `funsor.distribution.DistributionMeta`
- [x] Update `distribution_to_data`
- [x] Add tests
- [x] Fix `DirichletMultinomial` broadcasting errors

### DIFF
--- a/funsor/distribution.py
+++ b/funsor/distribution.py
@@ -453,9 +453,11 @@ def distribution_to_data(funsor_dist, name_to_dim=None):
     for param_name, funsor_param in zip(funsor_dist._ast_fields, funsor_dist._ast_values[:-1]):
         param = to_data(funsor_param, name_to_dim=name_to_dim)
 
+        # infer the independent dimensions of each parameter separately, since we chose to keep them unbroadcasted
         param_event_shape = getattr(funsor_dist._infer_param_domain(param_name, funsor_param.output.shape), "shape", ())
         param_indep_shape = funsor_param.output.shape[:len(funsor_param.output.shape) - len(param_event_shape)]
         for i in range(max(0, len(indep_shape) - len(param_indep_shape))):
+            # add singleton event dimensions, leave broadcasting/expanding to backend
             param = ops.unsqueeze(param, -1 - len(funsor_param.output.shape))
 
         params.append(param)
--- a/funsor/distribution.py
+++ b/funsor/distribution.py
@@ -16,7 +16,7 @@ import funsor.delta
 import funsor.ops as ops
 from funsor.affine import is_affine
 from funsor.cnf import Contraction, GaussianMixture
-from funsor.domains import Array, Real, Reals, RealsType
+from funsor.domains import Array, BintType, Real, Reals, RealsType
 from funsor.gaussian import Gaussian
 from funsor.interpreter import gensym
 from funsor.tensor import (Tensor, align_tensors, dummy_numeric_array, get_default_prototype,
@@ -147,14 +147,18 @@ class Distribution(Funsor, metaclass=DistributionMeta):
     def eager_log_prob(cls, *params):
         params, value = params[:-1], params[-1]
         params = params + (Variable("value", value.output),)
-        raw_dist, value_name, value_output, dim_to_name = reflect(cls, *params)._get_raw_dist()
+        instance = reflect(cls, *params)
+        raw_dist, value_name, value_output, dim_to_name = instance._get_raw_dist()
         assert value.output == value_output
         name_to_dim = {v: k for k, v in dim_to_name.items()}
-        dim_to_name.update({-1 - d -len(raw_dist.batch_shape): name
+        dim_to_name.update({-1 - d - len(raw_dist.batch_shape): name
                             for d, name in enumerate(value.inputs) if name not in name_to_dim})
         name_to_dim.update({v: k for k, v in dim_to_name.items() if v not in name_to_dim})
         raw_log_prob = raw_dist.log_prob(to_data(value, name_to_dim=name_to_dim))
-        return to_funsor(raw_log_prob, Real, dim_to_name=dim_to_name)
+        log_prob = to_funsor(raw_log_prob, Real, dim_to_name=dim_to_name)
+        inputs = value.inputs.copy()
+        inputs.update(instance.inputs)
+        return log_prob.align(tuple(k for k, v in inputs.items() if k in log_prob.inputs and isinstance(v, BintType)))
 
     def unscaled_sample(self, sampled_vars, sample_inputs, rng_key=None):
 
--- a/funsor/distribution.py
+++ b/funsor/distribution.py
@@ -16,7 +16,7 @@ import funsor.delta
 import funsor.ops as ops
 from funsor.affine import is_affine
 from funsor.cnf import Contraction, GaussianMixture
-from funsor.domains import Array, BintType, Real, Reals, RealsType
+from funsor.domains import Array, Real, Reals, RealsType
 from funsor.gaussian import Gaussian
 from funsor.interpreter import gensym
 from funsor.tensor import (Tensor, align_tensors, dummy_numeric_array, get_default_prototype,
--- a/funsor/distribution.py
+++ b/funsor/distribution.py
@@ -22,7 +22,7 @@ from funsor.interpreter import gensym
 from funsor.tensor import (Tensor, align_tensors, dummy_numeric_array, get_default_prototype,
                            ignore_jit_warnings, numeric_array, stack)
 from funsor.terms import Funsor, FunsorMeta, Independent, Number, Variable, \
-    eager, to_data, to_funsor
+    eager, reflect, to_data, to_funsor
 from funsor.util import broadcast_shape, get_backend, getargspec, lazy_property
 
 
@@ -120,14 +120,6 @@ class Distribution(Funsor, metaclass=DistributionMeta):
             return Number(0.)  # distributions are normalized
         return super(Distribution, self).eager_reduce(op, reduced_vars)
 
-    @classmethod
-    def eager_log_prob(cls, *params):
-        inputs, tensors = align_tensors(*params)
-        params = dict(zip(cls._ast_fields, tensors))
-        value = params.pop('value')
-        data = cls.dist_class(**params).log_prob(value)
-        return Tensor(data, inputs)
-
     def _get_raw_dist(self):
         """
         Internal method for working with underlying distribution attributes
@@ -150,6 +142,19 @@ class Distribution(Funsor, metaclass=DistributionMeta):
     @property
     def has_enumerate_support(self):
         return getattr(self.dist_class, "has_enumerate_support", False)
+
+    @classmethod
+    def eager_log_prob(cls, *params):
+        params, value = params[:-1], params[-1]
+        params = params + (Variable("value", value.output),)
+        raw_dist, value_name, value_output, dim_to_name = reflect(cls, *params)._get_raw_dist()
+        assert value.output == value_output
+        name_to_dim = {v: k for k, v in dim_to_name.items()}
+        dim_to_name.update({-1 - d -len(raw_dist.batch_shape): name
+                            for d, name in enumerate(value.inputs) if name not in name_to_dim})
+        name_to_dim.update({v: k for k, v in dim_to_name.items() if v not in name_to_dim})
+        raw_log_prob = raw_dist.log_prob(to_data(value, name_to_dim=name_to_dim))
+        return to_funsor(raw_log_prob, Real, dim_to_name=dim_to_name)
 
     def unscaled_sample(self, sampled_vars, sample_inputs, rng_key=None):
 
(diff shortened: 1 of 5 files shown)


funsor/distribution.py
```diff
@@ -191,7 +213,13 @@ def _infer_value_domain(cls, **kwargs):
         # rely on the underlying distribution's logic to infer the event_shape given param domains
         instance = cls.dist_class(**{k: dummy_numeric_array(domain) for k, domain in kwargs.items()},
                                   validate_args=False)
-        out_shape = instance.event_shape
+
+        # Note inclusion of batch_shape here to handle independent event dimensions.
+        # The arguments to _infer_value_domain are the .output shapes of parameters,
+        # so any extra batch dimensions that aren't part of the instance event_shape
+        # must be broadcasted output dimensions by construction.
+        out_shape = instance.batch_shape + instance.event_shape
```
eb8680, line 221: This change to `_infer_value_domain` is the meat of the PR; the rest of it is computing the correct arguments for `_infer_value_domain`.

funsor/distribution.py
```diff
@@ -129,6 +143,23 @@ def has_rsample(self):
     def has_enumerate_support(self):
         return getattr(self.dist_class, "has_enumerate_support", False)
 
+    @classmethod
+    def eager_log_prob(cls, *params):
+        params, value = params[:-1], params[-1]
+        params = params + (Variable("value", value.output),)
+        instance = reflect(cls, *params)
+        raw_dist, value_name, value_output, dim_to_name = instance._get_raw_dist()
```
eb8680, line 151: I had to refactor `eager_log_prob` to use `Distribution._get_raw_dist()` to get the new tests to pass.

funsor/distribution.py
```diff
@@ -142,7 +173,8 @@ def unscaled_sample(self, sampled_vars, sample_inputs, rng_key=None):
         sample_shape = tuple(v.size for v in sample_inputs.values())
         sample_args = (sample_shape,) if get_backend() == "torch" else (rng_key, sample_shape)
         if self.has_rsample:
-            raw_value = raw_dist.rsample(*sample_args)
+            # TODO fix this hack by adding rsample and has_rsample to Independent upstream in NumPyro
+            raw_value = getattr(raw_dist, "rsample", raw_dist.sample)(*sample_args)
```
eb8680, line 177: @fehiepsi it looks like `Independent` and other distribution modifiers like `MaskedDistribution` don't implement `rsample`/`has_rsample`. Is this intentional?

fehiepsi: Yes, currently `rsample` or `has_rsample` is not available in NumPyro (they are not used by any inference algorithms in NumPyro). In [jax backend](https://github.com/pyro-ppl/funsor/blob/master/funsor/jax/distributions.py#L114), I also added a similar workaround. Could you add this fix [here](https://github.com/pyro-ppl/funsor/blob/master/funsor/jax/distributions.py#L218)?

funsor/distribution.py
```diff
@@ -57,12 +57,34 @@ class DistributionMeta(FunsorMeta):
...
+            if k == "value":
+                continue
+
+            # compute unbroadcasted param domains
+            domain = cls._infer_param_domain(k, getattr(kwargs[k], "shape", ()))
+            # use to_funsor to infer output dimensions of e.g. tensors
+            domains[k] = domain if domain is not None else to_funsor(v).output
+
+            # broadcast individual param domains with Funsor inputs
+            # this avoids .expand-ing underlying parameter tensors
```
fehiepsi, line 74: What is the expected domain of `scale` for `Normal(Reals[2], 1.)` and `Normal(Reals[2], torch.ones(2))`. Currently, `domains["scale"]` will be `Real` in both case. The second case will trigger an error at `to_funsor(v, output=domains[k])` below.

In either case, I guess we need to rewrite `eager_normal` or `eager_mvn` to address `Reals[2]` loc. cc @fritzo

eb8680: > What is the expected domain of scale for `Normal(Reals[2], 1.)` and `Normal(Reals[2], torch.ones(2))`?

In the first case, it's `Real`, and in the second, it's `Reals[2]`.  I guess I should add a second broadcasting condition below to handle the case where the parameter is a raw tensor:
```py
if ops.is_numeric_array(v):  # at this point we know all of v's dims are output dims
    domains[k] = Reals[broadcast_shape(v.shape, domains[k].shape)]
```

> In either case, I guess we need to rewrite eager_normal or eager_mvn to address Reals[2] loc

Can you explain what change you think may be necessary?

funsor/distribution.py
eb8680: Ok, thanks, I pushed this patch

fritzo:
It looks like you just need to relax precision of `test_dirichlet_density()` for the jax backend.

eb8680:
@fehiepsi there are a ton of unrelated new failures on the last JAX build, any idea what's going on? Seems like a mix of weird new numerical errors and `ValueError: cannot compare objects of type <class 'jax.interpreters.xla._DeviceArray'>`.

fehiepsi (approved):
The logic to get expected outputs looks reasonable to me, pending further comments by @fritzo.

I'll try to address `eager_normal`, `eager_mvn` issues in a follow-up PR. Currently, those eager rules assume `Real` args but we have `Reals` or a mix of Real, Reals now.

eb8680:
> I'll try to address eager_normal, eager_mvn issues in a follow-up PR. Currently, those eager rules assume Real args but we have Reals or a mix of Real, Reals now.

Hmm, I wonder if we'll have to do this for all the other eager distribution patterns in `funsor/distribution.py` as well...

funsor/distribution.py
```diff
@@ -400,10 +439,32 @@ def __call__(self, cls, args, kwargs):
 
 @to_data.register(Distribution)
 def distribution_to_data(funsor_dist, name_to_dim=None):
-    params = [to_data(getattr(funsor_dist, param_name), name_to_dim=name_to_dim)
-              for param_name in funsor_dist._ast_fields if param_name != 'value']
-    pyro_dist = funsor_dist.dist_class(**dict(zip(funsor_dist._ast_fields[:-1], params)))
     funsor_event_shape = funsor_dist.value.output.shape
+
+    # attempt to generically infer the independent output dimensions
+    instance = funsor_dist.dist_class(**{
```
fritzo, line 445: Beyond the scope of this PR, I'm concerned with the increasing overhead of shape computations that need to do tensor ops. I like @fehiepsi's recent suggestion of implementing `.forward_event_shape()` for transforms. I think it would be worthwhile to discuss and think about extensions to the Distribution interface that could replace all this need to create an throw away dummy distributions.

(Indeed in theory an optimizing compiler could remove all this overhead, but in practice our tensor backends either incur super-linear compile time cost, or fail to cover the wide range of probabilistic models we would like to handle. And while these dummy tensor ops are cheap, they add noise to debugging efforts.)

funsor/distribution.py
```diff
@@ -400,10 +439,32 @@ def __call__(self, cls, args, kwargs):
...
+        k: dummy_numeric_array(domain)
+        for k, domain in zip(funsor_dist._ast_fields, (v.output for v in funsor_dist._ast_values))
+        if k != "value"
+    }, validate_args=False)
+    event_shape = broadcast_shape(instance.event_shape, funsor_dist.value.output.shape)
+    # XXX is this final broadcast_shape necessary? should just be event_shape[...]?
+    indep_shape = broadcast_shape(instance.batch_shape, event_shape[:len(event_shape) - len(instance.event_shape)])
```
fritzo, line 452: nit: could you make this more explicit and readable by declaring whether the result is negative or nonnegative? I ask because numpy slicing semantics treats the two sets very differently.
```py
reinterpreted_batch_ndims = len(event_shape) - len(instance.event_shape)
assert reinterpreted_batch_ndims > 0  # XXX is this ever nonzero?
indep_shape = broadcast_shape(instance.batch_shape, event_shape[:reinterpreted_batch_ndims])
```

funsor/jax/distributions.py
```diff
@@ -215,6 +215,13 @@ def deltadist_to_data(funsor_dist, name_to_dim=None):
 # Converting PyTorch Distributions to funsors
 ###############################################
 
+dist.Independent.has_rsample = property(lambda self: self.base_dist.has_rsample)
+dist.Independent.rsample = dist.Independent.sample
+dist.MaskedDistribution.has_rsample = property(lambda self: self.base_dist.has_rsample)
+dist.MaskedDistribution.rsample = dist.MaskedDistribution.sample
+dist.TransformedDistribution.has_rsample = property(lambda self: self.base_dist.has_rsample)
+dist.TransformedDistribution.rsample = dist.TransformedDistribution.sample
```
fritzo, line 223: Can you add a TODO pointing to a NumPyro issue to fix this bug, so we can delete this workaround once the bug is fixed? cc @fehiepsi

funsor/torch/distributions.py
```diff
@@ -137,7 +137,7 @@ def _infer_value_domain(**kwargs):
 @functools.lru_cache(maxsize=5000)
 def _infer_value_domain(cls, **kwargs):
     instance = cls.dist_class(**{k: dummy_numeric_array(domain) for k, domain in kwargs.items()}, validate_args=False)
-    return Reals[instance.event_shape]
+    return Reals[instance.batch_shape + instance.event_shape]
```
fritzo, line 140: Hmm can you explain this? As I understood only `backend.Independent` distributions would trigger batch dims being converted to event dims, but this line suggests we no longer support default interpretation of batch_shape as funsor inputs 😕

fritzo (commented):
(thanks for your patience @eb8680)

funsor/jax/distributions.py
fehiepsi: @neerajprad Should we add those new attributes to NumPyro distributions? We can make default behaviors for them so that there would be only a few changes in the code.

neerajprad: Sounds good. In numpyro, for distributions that have reparametrized samplers available both will be the same so we can just add a default `Distribution.rsample` method which delegates to `sample` and throws a `NotImplemented` error when not available.

funsor/distribution.py
eb8680: Yes, I agree the repeated creation of distribution instances here is not ideal. Perhaps we could add counterparts of some of the shape inference methods from TFP (e.g. [`event_shape_tensor`](https://www.tensorflow.org/probability/api_docs/python/tfp/distributions/Distribution#event_shape_tensor), [`param_shapes`](https://www.tensorflow.org/probability/api_docs/python/tfp/distributions/Distribution#param_shapes)) upstream in `torch.distributions`.

funsor/torch/distributions.py
eb8680: `_infer_value_domain` takes the `.output` domains of the distribution's parameters as arguments, so `instance.batch_shape` contains only *output* dimensions of the parameters/value, not dimensions that would have otherwise been converted to funsor inputs.

Before this PR, passing a distribution parameters with extra output dimensions would simply error; now these dimensions broadcast:
```py
with interpretation(lazy):
    d = funsor.torch.distributions.Normal(loc=funsor.Tensor(torch.zeros(3), OrderedDict()), scale=1.)
assert d.inputs["value"] == d.loc.output == Reals[3]
```

fritzo: Thanks for explaining!

eb8680:
@fritzo any other comments? I'd like to bump the Funsor dependency version in Pyro to the latest `master` to unblock @ordabayevy's PR https://github.com/pyro-ppl/pyro/pull/2716 and we might as well include these changes too.

fritzo:
Thanks for the reminder and thanks for addressing nits! LGTM.

PR #403, opened 2020-12-03
Partial unrolling of plated factor graphs to standard factor graphs

ordabayevy:
`partial_unroll` function performs partial unrolling operation described in [the TVE paper](https://arxiv.org/abs/1902.03210). This will be helpful after it gets updated to unroll markov dims (in future PR) to facilitate testing of the `modified_partial_sum_product` algorithm.

### DIFF
--- a/funsor/sum_product.py
+++ b/funsor/sum_product.py
@@ -45,6 +45,61 @@ def _partition(terms, sum_vars):
     return components
 
 
+def _unroll_plate(factors, plate, size, var_to_ordinal):
+    # replicated variables
+    plate_vars = set()
+    for var, ordinal in var_to_ordinal.items():
+        if plate in ordinal:
+            plate_vars.add(var)
+
+    # unroll factors
+    unrolled_factors = []
+    for factor in factors:
+        if plate in factor.inputs:
+            unrolled_factors.extend([factor(
+                    **{plate: i},
+                    **{var: "{}_{}{}".format(var, plate, i) for var in plate_vars}
+                ) for i in range(size)])
+        else:
+            unrolled_factors.append(factor)
+
+    # unroll variables
+    for var in plate_vars:
+        ordinal = var_to_ordinal.pop(var)
+        var_to_ordinal.update({"{}_{}{}".format(var, plate, i): ordinal.difference(plate)
+                               for i in range(size)})
+
+    return unrolled_factors, var_to_ordinal
+
+
+def partial_unroll(factors, plates=frozenset()):
+    """
+    Performs partial unrolling of plated factor graphs to standard factor graphs.
+
+    :return: a list of partially unrolled Funsors
+        and a frozenset of partially unrolled variables.
+    """
+    assert isinstance(factors, (tuple, list))
+    assert all(isinstance(f, Funsor) for f in factors)
+    assert isinstance(plates, frozenset)
+
+    var_to_ordinal = {}
+    plate_to_size = {}
+    for f in factors:
+        ordinal = plates.intersection(f.inputs)
+        for var in frozenset(f.inputs) - plates:
+            var_to_ordinal[var] = var_to_ordinal.get(var, ordinal) & ordinal
+        for plate in ordinal:
+            plate_to_size[plate] = f.inputs[plate].size
+
+    # unroll one plate at a time
+    for plate in plates:
+        size = plate_to_size[plate]
+        factors, var_to_ordinal = _unroll_plate(factors, plate, size, var_to_ordinal)
+
+    return factors, frozenset(var_to_ordinal.keys())
+
+
 def partial_sum_product(sum_op, prod_op, factors, eliminate=frozenset(), plates=frozenset()):
     """
     Performs partial sum-product contraction of a collection of factors.
--- a/funsor/sum_product.py
+++ b/funsor/sum_product.py
@@ -45,18 +45,33 @@ def _partition(terms, sum_vars):
     return components
 
 
-def _unroll_plate(factors, plate, var_to_ordinal):
+def _unroll_plate(factors, var_to_ordinal, sum_vars, plate):
+    # size of the plate
+    size = next(iter(f.inputs[plate].size for f in factors if plate in f.inputs))
+
     # replicated variables
     plate_vars = set()
     for var, ordinal in var_to_ordinal.items():
         if plate in ordinal:
             plate_vars.add(var)
 
+    # make sure that all vars in the plate are being unrolled
+    assert plate_vars.issubset(sum_vars)
+
+    # unroll variables
+    for var in plate_vars:
+        sum_vars -= frozenset({var})
+        sum_vars |= frozenset({"{}_{}{}".format(var, plate, i)
+                               for i in range(size)})
+        ordinal = var_to_ordinal.pop(var)
+        new_ordinal = ordinal.difference(plate)
+        var_to_ordinal.update({"{}_{}{}".format(var, plate, i): new_ordinal
+                               for i in range(size)})
+
     # unroll factors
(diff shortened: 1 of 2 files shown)


funsor/sum_product.py
```diff
@@ -45,6 +45,61 @@ def _partition(terms, sum_vars):
...
+    # unroll variables
+    for var in plate_vars:
+        ordinal = var_to_ordinal.pop(var)
+        var_to_ordinal.update({"{}_{}{}".format(var, plate, i): ordinal.difference(plate)
+                               for i in range(size)})
+
+    return unrolled_factors, var_to_ordinal
+
+
+def partial_unroll(factors, plates=frozenset()):
```
eb8680, line 75: I think it would make sense for `partial_unroll` to take an `eliminate` argument and return an updated version of `eliminate` with any plates removed and variables replaced with unrolled versions.

funsor/sum_product.py
```diff
@@ -45,6 +45,61 @@ def _partition(terms, sum_vars):
...
+            var_to_ordinal[var] = var_to_ordinal.get(var, ordinal) & ordinal
+        for plate in ordinal:
+            plate_to_size[plate] = f.inputs[plate].size
+
+    # unroll one plate at a time
+    for plate in plates:
+        size = plate_to_size[plate]
+        factors, var_to_ordinal = _unroll_plate(factors, plate, size, var_to_ordinal)
+
+    return factors, frozenset(var_to_ordinal.keys())
```
eb8680, line 100: Per my suggestion above, I think this should return an updated version of `eliminate` rather than a set of all the variables, eliminated or otherwise.

funsor/sum_product.py
```diff
@@ -45,6 +45,61 @@ def _partition(terms, sum_vars):
...
+
+    :return: a list of partially unrolled Funsors
+        and a frozenset of partially unrolled variables.
+    """
+    assert isinstance(factors, (tuple, list))
+    assert all(isinstance(f, Funsor) for f in factors)
+    assert isinstance(plates, frozenset)
+
+    var_to_ordinal = {}
+    plate_to_size = {}
```
eb8680, line 87: nit: `plate_to_size` seems unnecessary, just inline reading the size from `.inputs`

funsor/sum_product.py
```diff
@@ -45,6 +45,61 @@ def _partition(terms, sum_vars):
     return components
 
 
+def _unroll_plate(factors, plate, size, var_to_ordinal):
```
eb8680, line 48: nit: remove and inline `size`

eb8680 (commented):
Nice, this will be helpful for testing.

funsor/sum_product.py
```diff
@@ -45,6 +45,67 @@ def _partition(terms, sum_vars):
...
+                ) for i in range(size)])
+        else:
+            unrolled_factors.append(factor)
+
+    # unroll variables
+    for var in plate_vars:
+        ordinal = var_to_ordinal.pop(var)
+        new_ordinal = ordinal.difference(plate)
+        var_to_ordinal.update({"{}_{}{}".format(var, plate, i): new_ordinal
+                               for i in range(size)})
```
ordabayevy, line 73: `size` on this line is defined in the for loop above. There is probably a better way to do this ...

ordabayevy (commented):
@eb8680 thanks for reviewing! It  returns unrolled `eliminate` and remaining `plates` now.

test/test_sum_product.py
```diff
@@ -80,9 +81,9 @@ def test_partition(inputs, dims, expected_num_components):
 @pytest.mark.parametrize('sum_op,prod_op', [(ops.add, ops.mul), (ops.logaddexp, ops.add)])
 @pytest.mark.parametrize('inputs,plates', [('a,abi,bcij', 'ij')])
 @pytest.mark.parametrize('vars1,vars2', [
+    ('cj', 'abi'),
     ('', 'abcij'),
     ('c', 'abij'),
-    ('cj', 'abi'),
```
ordabayevy, line 85: I'll fix this in the next commit.

funsor/sum_product.py
```diff
@@ -45,6 +45,67 @@ def _partition(terms, sum_vars):
...
+    assert all(isinstance(f, Funsor) for f in factors)
+    assert isinstance(eliminate, frozenset)
+    assert isinstance(plates, frozenset)
+    sum_vars = eliminate - plates
+    unrolled_plates = eliminate & plates
+
+    var_to_ordinal = {}
+    for f in factors:
+        ordinal = plates.intersection(f.inputs)
+        for var in sum_vars.intersection(f.inputs):
```
eb8680, line 96: Sorry - to clarify, I think you should still be unrolling all plates and variables, but you should only *return* an updated version of `eliminate`, so that the output of `partial_unroll` can be passed directly to `partial_sum_product` with an empty `plates` set.  As of this comment, you're no longer unrolling all variables, although you are returning `eliminate`.

funsor/sum_product.py
```diff
@@ -45,6 +45,67 @@ def _partition(terms, sum_vars):
...
+    :return: a list of partially unrolled Funsors,
+        a frozenset of partially unrolled variable names,
+        and a frozenset of remaining plates.
+    """
+    assert isinstance(factors, (tuple, list))
+    assert all(isinstance(f, Funsor) for f in factors)
+    assert isinstance(eliminate, frozenset)
+    assert isinstance(plates, frozenset)
+    sum_vars = eliminate - plates
+    unrolled_plates = eliminate & plates
```
eb8680, line 91: I think you should always unroll all plates and variables - I can't think of a use case for `partial_unroll` where you would only want to unroll some.

funsor/sum_product.py
eb8680: How about `size = next([f.inputs[plate].size for f in factors if plate in f.inputs])`?

eb8680 (commented):
I see the main use case for `partial_unroll` as a testing helper (and eventually as part of a definition of generalized plated factor graphs, as in the TVE paper). To that end, I think it should implement complete unrolling of all plates and variables, whether eliminated or not, and return an updated `factors` and `eliminate` (which might only be a subset of the unrolled variables) that can be thrown back into `partial_sum_product`.
```py
partial_sum_product(sum_op, prod_op, factors, eliminate, plates)
# equivalent to
factors, eliminate = partial_unroll(factors, plates, eliminate)
reduce(prod_op, factors).reduce(sum_op, eliminate)
```

ordabayevy:
> To that end, I think it should implement complete unrolling of all plates and variables, whether eliminated or not

You mean equivalent of this, right?

```py
partial_unroll(factors, eliminate|plates, plates)
```

> return an updated `factors` and `eliminate` (which might only be a subset of the unrolled variables) that can be thrown back into `partial_sum_product`.

These two `partial_sum_product` with original and unrolled factors and variables return the same list of factors:

```py
partial_sum_product(sum_op, prod_op, factors, eliminate, plates)
# equivalent to
unrolled_factors, unrolled_vars, remaining_plates = partial_unroll(factors, eliminate, plates)
partial_sum_product(sum_op, prod_op, unrolled_factors, unrolled_vars, remaining_plates)
```

> you should still be unrolling all plates and variables, but you should only return an updated version of `eliminate`, so that the output of `partial_unroll` can be passed directly to `partial_sum_product` with an empty plates

The potential problem here is that `partial_sum_product` will return list of unrolled factors for `vars` that were not eliminated. However, then the names of those unrolled `vars` will be missing if you want to eliminate them later.

ordabayevy:
Failure unrelated to this PR (`rtol` in `test_distribution_generic`)

eb8680 (approved):
LGTM after Zoom discussion

PR #405, opened 2020-12-10
Fix trivial lint to test JAX backend

fehiepsi:


eb8680 (approved):
Thanks for looking into this! Feel free to merge yourself if tests pass.

fehiepsi:
Luckily that the issue is easy to fix. Otherwise, it will be tricky to address a bunch of numerical issues. :))

PR #406, opened 2020-12-13
Allow empty tensor conversion

fehiepsi:
I'm not sure if funsor supports Bint[0] at all but this is needed when using `numpyro.factor` with `enum`. Under the hood, `factor` leverages a Unit site with empty size: `event_shape=(0,)`, but `to_funsor` couldn't convert the empty array to funsor.

I also added a smoke test, which fails without this change.


fritzo (approved):
Seems reasonable to support.

PR #407, opened 2020-12-15
Update modified_partial_sum_product to sum over init factors

ordabayevy:
In this PR:
- Update `modified_partial_sum_product` to eliminate `init` factors as well
- Change `step` from a dict to a set of tuples in `modified_partial_sum_product` and `partial_unroll`
- Add `init` factors in the tests
- Remove `test_modified_partial_sum_product_15` because the new `step` only works for markov vars that depend only one markov dimension.

### DIFF
--- a/funsor/sum_product.py
+++ b/funsor/sum_product.py
@@ -102,10 +102,22 @@ def partial_unroll(factors, eliminate=frozenset(), plate_to_step=dict()):
     """
     Performs partial unrolling of plated factor graphs to standard factor graphs.
 
-    Currently only plates with history={0, 1} are supported.
+    For plates (history=0) unrolling operation appends ``_{i}`` suffix
+    to variable names for index ``i`` in the plate. For markov dimensions (history>0)
+    unrolling operation removes the suffix for the j-th Markov variable name counted
+    from the end for the tuple of names in the ``step`` (e.g., j=0 for "x_curr"
+    and j=1 for "x_prev") and then appends ``_{i+history-j}`` to the name for index ``i``.
     Markov vars are assumed to have names that follow ``var_suffix`` formatting
-    (e.g., ``("x_0", "x_prev", "x_curr")``).
+    (e.g., ``("x_0", "x_prev", "x_curr")`` for history=1).
 
+    :param factors: A collection of funsors.
+    :type factors: tuple or list
+    :param frozenset eliminate: A set of free variables to unroll,
+        including both sum variables and product variable.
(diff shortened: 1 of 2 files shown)


eb8680 (commented):
Can you update the docstrings of `modified_partial_sum_product` and `partial_unroll` to reflect the new type of `plate_to_step` and the `init` factor requirement?

Also, what's the intended behavior when no factor contains an init var?

ordabayevy:
Updated docstrings for `modified_partial_sum_product` and `partial_unroll`.

> Also, what's the intended behavior when no factor contains an init var?

Hmm ... I don't have any intended behavior in mind for that case. Currently the `KeyError` for `var_to_ordinal[v]` is thrown in: 

```
307                     new_plates = frozenset().union(
308  ->                     *(var_to_ordinal[v] for v in remaining_sum_vars))
```

eb8680:
One thing I'm still not clear on in the new `partial_unroll` logic: what guarantees that init var names will match the first timestep of unrolled Markov vars? Is it the assumption that init vars in the input factors are named `{var}_0`?

I think it would simplify this PR a bit (especially since we have no tests for `history > 1`) if you made your changes to `partial_unroll` more like your changes to `modified_partial_sum_product`: rather than modifying the guts of `_unroll_plate`, just transform the initial `plate_to_step` in `partial_unroll` into the form expected before this PR, so that the actual unrolling algorithm works exactly the same as before. You can even factor out the logic from `modified_partial_sum_product` into a small helper function and use that in both places.

ordabayevy:
>  Is it the assumption that init vars in the input factors are named {var}_0?

Yes. Added that in the docstring of the `partial_unroll`

> I think it would simplify this PR a bit (especially since we have no tests for history > 1)

Agree. That's how I actually did it in the earlier commits.

ordabayevy:
I had this issue multiple times when `test_distribution_generic.py` fails randomly. It seems like `rtol` is small enough that leads to occasional `AssertionError`s.
```
FAILED test/test_distribution_generic.py::test_generic_log_prob
...
                if rtol is not None:
>                   assert (diff / (atol + expected.detach().abs())).max() < rtol, msg
E                   AssertionError: Expected:
E                   tensor([ 8.1429e+00,  8.7689e-01,  1.4912e+00, -2.1520e-04,  3.2218e+00])
E                   Actual:
E                   tensor([ 8.1429e+00,  8.7689e-01,  1.4912e+00, -2.1514e-04,  3.2218e+00])
```

eb8680 (approved):

PR #408, opened 2020-12-18
Fix rsample not available in NumPyro

fehiepsi:
This unblocks https://github.com/pyro-ppl/numpyro/pull/844.

Currently, we checked if a numpyro dist has the attribute `has_rsample`. If it does, `has_rsample` will be True. That logic is incorrect because `has_rsample` can be a property or get the value False. This PR fixes that bug.


eb8680 (approved):
LGTM, feel free to merge when tests pass

fehiepsi:
Thanks for reviewing!

PR #409, opened 2020-12-20
Support partial windows in sarkka_bilmes_product

eb8680:
Addresses #293, https://github.com/pyro-ppl/numpyro/pull/848

This PR adds support for partial windows in `funsor.sum_product.sarkka_bilmes_product`, i.e. for factors where `duration % period != 0`.  The algorithm now breaks the input funsor into two contiguous pieces: one containing the rightmost `duration - (duration % period)` slices and one containing the remaining slices on the left.  It then recursively calls `sarkka_bilmes_product` on the right piece and sequentially combines the result with slices from the left piece.

Tested:
- Added cases to existing `sarkka_bilmes_product` tests in `test_sum_product.py`

PR #410, opened 2020-12-21
Defer has_rsample to raw_dist

fehiepsi:
Fixes failing tests in #409. Currently, funsor Distribution assumes that `has_rsample` is a class attribute but in NumPyro (and derived classes such as TransformedDistribution, Independent, MaskedDistribution, ExpandedDistribution), it is a property. So `getattr(self.dist_class, "has_rsample", False)` will always return False.

I think we can also cache `has_rsample` or `has_enumerate_support` in `self._has_rsample` or `self._has_enumerate_support` when calling `_infer_value_domain`.

### DIFF
--- a/funsor/distribution.py
+++ b/funsor/distribution.py
@@ -138,10 +138,6 @@ class Distribution(Funsor, metaclass=DistributionMeta):
         return raw_dist, value_name, value_output, dim_to_name
 
     @property
-    def has_rsample(self):
-        return getattr(self.dist_class, "has_rsample", False)
-
-    @property
     def has_enumerate_support(self):
         return getattr(self.dist_class, "has_enumerate_support", False)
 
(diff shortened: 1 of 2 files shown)


eb8680 (approved):

fritzo (approved):
Thanks for fixing this. LGTM if tests pass.

fehiepsi:
@fritzo This seems to be a flaky test: `actual = tensor([-6.3121e-05]), expected = tensor([-6.3241e-05]), atol = 1e-06, rtol = 0.0001` of
```
test_generic_log_prob[True-backend_dist.TransformedDistribution(backend_dist.Uniform(low=case.low, high=case.high), [backend_dist.transforms.TanhTransform(), backend_dist.transforms.ExpTransform()]) (('low', 'rand((5,))'), ('high', '2. + rand((5,))'))] 
```
under torch backend. I can't reproduce it locally for many runs. We can keep an eye on it to adjust the parameters if it happens again.

PR #413, opened 2020-12-28
[WIP] compute_expectation wrt plate and markov variables

ordabayevy:
Addresses https://github.com/pyro-ppl/pyro/issues/2724.

This function computes the expected value of cost terms wrt to log measures.

### DIFF
--- a/funsor/sum_product.py
+++ b/funsor/sum_product.py
@@ -340,42 +340,19 @@ def modified_partial_sum_product(sum_op, prod_op, factors,
     return results
 
 
-def expected_terms(sum_op, prod_op, factors, integrand,
-                   eliminate=frozenset(), plate_to_step=dict()):
+def compute_expectation(factors, integrand, eliminate=frozenset(), plate_to_step=dict()):
     """
     Compute the expected cost term E_q[logp] or E_q[-logq] w.r.t. q
     """
-    assert callable(sum_op)
-    assert callable(prod_op)
     assert isinstance(factors, (tuple, list))
     assert all(isinstance(f, Funsor) for f in factors)
     assert isinstance(integrand, Funsor)
     assert isinstance(eliminate, frozenset)
     assert isinstance(plate_to_step, dict)
-    # process plate_to_step
-    plate_to_step = plate_to_step.copy()
-    prev_to_init = {}
-    for key, step in plate_to_step.items():
-        # map prev to init; works for any history > 0
-        for chain in step:
-            init, prev = chain[:len(chain)//2], chain[len(chain)//2:-1]
-            prev_to_init.update(zip(prev, init))
-        # convert step to dict type required for MarkovProduct
-        plate_to_step[key] = {chain[1]: chain[2] for chain in step}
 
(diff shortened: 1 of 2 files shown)


ordabayevy (commented):
@eb8680 can you review this?
Some notes:
- `TraceMarkovEnum_ELBO` that uses `compute_expectation` passes all of the tests locally for updated `test_vectorized_markov::test_guide_enumerated_elbo` from https://github.com/pyro-ppl/pyro/pull/2726
-  This PR can be simplified if `funsor.adjoint` can be used instead of `forward_backward_terms` function written here.

eb8680:
@ordabayevy I'm impressed you got this working, and hopefully your model should be unblocked at least locally via this PR and https://github.com/pyro-ppl/pyro/pull/2725 .  Please let me know if that's not the case.

However, I am reluctant to merge this in its current form, especially without any tests - the algorithm implemented here is quite complex, and there is a substantial amount of code duplication from `modified_partial_sum_product` that I would prefer to reduce or avoid if possible given the changes we expect to make to the internals of `modified_partial_sum_product` in the future (#293).  It's also not yet clear to me how we will need to change the algorithm or interface to facilitate sharing of computation across multiple `integrand`s, as is done (somewhat imperfectly) in Pyro's `pyro.infer.util.Dice.compute_expectation`.

Can we talk sometime this week about how it might be simplified? I have some intuitions but not a fully fleshed out plan. It would probably be useful for both of us to prepare by reading [this paper](https://www.aclweb.org/anthology/D09-1005.pdf), which discusses general dynamic programming algorithms for computing expectations.

ordabayevy:
> Please let me know if that's not the case.

Okay, I'll let you know if there are any problems switching to Funsor backend and trying out this version of `TraceMarkovEnum_ELBO`.

> Can we talk sometime this week about how it might be simplified? 

Thursday and Friday will work for me.

ordabayevy:
@eb8680 I've created an HMM version of my model where both `m` and `theta` are guide enumerated. I've tested it on simulated and real data and it seems to work! Although a bit slow which might be because `compute_expectation` works only in eager interpretation. As we discussed last week lazy implementation is blocked by the `_scatter` function which doesn't have a lazy equivalent right now.

eb8680:
@ordabayevy that's great!  I'll look into making `_scatter` a first-class funsor. One other thing you can try to speed it up is using the PyTorch JIT - it won't make the underlying tensor operations any faster but it should at least compile away the Funsor pattern-matching overhead.

That should be as simple as creating a `JitTraceMarkovEnum_ELBO` class [similar to `JitTraceEnum_ELBO`](https://github.com/pyro-ppl/pyro/blob/dev/pyro/contrib/funsor/infer/traceenum_elbo.py#L171) and dropping it into your Pyro code:
```py
from pyro.contrib.funsor.infer.elbo import Jit_ELBO
...
class JitTraceMarkovEnum_ELBO(Jit_ELBO, TraceMarkovEnum_ELBO):
    pass
```

ordabayevy:
Closing as obsolete.

PR #414, opened 2020-12-28
Ensure alpha conversion always runs when a new variable is bound

eb8680:
This PR fixes a longstanding alpha-conversion edge case by guaranteeing that newly bound variables are always alpha-renamed, which is currently not the case due to the way alpha-conversion interacts with cons-hashing.  For context, this issue turned out to be the root cause of some strange, difficult-to-debug errors that came up when I tested `funsor.adjoint` with `funsor.sum_product.modified_partial_sum_product`.  I imagine @ordabayevy will run into this at some point, so think of this PR as preemptive unblocking.

Recall that alpha conversion in Funsor is implemented with name-mangling of subexpressions in `reflect`: whenever a variable is newly bound in a Funsor expression, we append to its name a unique suffix of the form `f"__BOUND_{identifier}"`, where `identifier` is some short piece of information ensuring that the mangled name will not collide with bound variables elsewhere in the expression.  Right now, `identifier` is just a global counter incremented after each `funsor.interpreter.gensym()` call.  

However, to ensure that expressions that bind variables are cons-hashed, we have to avoid mangling names twice.  This is currently achieved by only mangling bound variables whose names do not already contain a `__BOUND_` suffix. Unfortunately, this heuristic can break down in cases where subexpressions with bound variables are (temporarily) exposed for further processing. One place where this can occur is in `funsor.adjoint`, especially in the nested `funsor.terms.Subs` expressions in `funsor.sum_product.sequential_sum_product`.

This PR replaces the global counter `identifier` with the (`hash()` of) the cons-hash key of the binding expression.  At the cost of slightly longer (but still constant-size) bound variable names, this change should guarantee that alpha-conversion always behaves as expected.

Tested:
- Added a simple regression test that fails on `master`
- Exercised explicitly by existing tests in `test_alpha_conversion.py` and implicitly in any test that compares identities of expressions with bound variables (`x1 is x2`)

### DIFF
--- a/funsor/terms.py
+++ b/funsor/terms.py
@@ -41,19 +41,19 @@ def substitute(expr, subs):
         return interpreter.reinterpret(expr)
 
 
-def _alpha_mangle(expr):
+def _alpha_mangle(expr, identifier):
     """
     Rename bound variables in expr to avoid conflict with any free variables.
-
-    FIXME this does not avoid conflict with other bound variables.
+    Returns expr._ast_values with mangled names.
     """
-    alpha_subs = {name: interpreter.gensym(name + "__BOUND")
-                  for name in expr.bound if "__BOUND" not in name}
+    # how we know which variables to mangle: we assume variable names include
+    # constant-size information about the original binding context,
+    # in the form of a cons-hash key of the binding context.
+    alpha_subs = {name: name.split("__BOUND")[0] + "__BOUND_" + identifier
+                  for name in expr.bound if identifier not in name}
     if not alpha_subs:
-        return expr
-
-    ast_values = expr._alpha_convert(alpha_subs)
-    return reflect(type(expr), *ast_values)
+        return expr._ast_values
+    return expr._alpha_convert(alpha_subs)  # return mangled _ast_values
 
 
 def reflect(cls, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


funsor/terms.py
```diff
@@ -81,8 +74,24 @@ def reflect(cls, *args, **kwargs):
...
+        alpha_mangled_args = result._alpha_convert(alpha_subs)
+
+        # TODO eliminate code duplication below
+        # this is currently necessary because .bound is computed in __init__().
+        result = super(FunsorMeta, cls_specific).__call__(*alpha_mangled_args)
+        result._ast_values = alpha_mangled_args
+
+        # we also make the old cons cache_key point to the new mangled value.
+        # this guarantees that alpha-conversion only runs once for this expression.
+        cls._cons_cache[cache_key] = result
```
eb8680, line 91: This line ensures that `_alpha_mangle` does not break cons-hashing.

eb8680:
The CI failures are `RecursionError`s in `examples/hmm.py`. After some investigation it seems like there's an odd interaction happening between `substitute` and `funsor.optimize.apply_optimizer` that's causing the stack of interpreters to blow up.  I don't think it's related to the changes in this PR; they just happened to make it more obvious.  I'll have to fix that first before this can be merged.

eb8680:
> there's an odd interaction happening between substitute and funsor.optimize.apply_optimizer that's causing the stack of interpreters to blow up

Update: I believe I have fixed this by forcing alpha-renaming substitution to run under `reflect`, which I believe is always correct.  The same example is still giving a `RecursionError`, but I think the root cause has changed. Still investigating...

eb8680:
[This recent paper](https://arxiv.org/abs/2105.02856) seems quite relevant, although it solves a slightly different problem.

PR #415, opened 2020-12-31
Change P to _PREV_ in sarkka_bilmes_generic

fehiepsi:
to avoid name confliction in Pyro/NumPyro models.


eb8680 (approved):
LGTM, thanks for fixing this!

PR #418, opened 2021-01-09
Add method for converting Pyro ExpandedDistribution to funsor

eb8680:
Group coded with @fritzo, @ordabayevy, @fehiepsi. Addresses #386. Blocking @ordabayevy's work.

PR #419, opened 2021-01-09
support ExpandedDistribution for jax backend

fehiepsi:
Follow up #418.

Though we can eagerly expand distributions, as in https://github.com/pyro-ppl/numpyro/pull/809/files, it might be better to not have to think about doing so each time we want to use funsor. :D

### DIFF
--- a/test/test_distribution_generic.py
+++ b/test/test_distribution_generic.py
@@ -40,6 +40,12 @@ if get_backend() != "numpy":
     FAKES = _fakes()
 
 
+if get_backend() == "jax":
+    _expanded_dist_path = "backend_dist.ExpandedDistribution"
+elif get_backend() == "torch":
+    _expanded_dist_path = "backend_dist.torch_distribution.ExpandedDistribution"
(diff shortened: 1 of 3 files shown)


test/test_distribution_generic.py
```diff
@@ -40,6 +40,12 @@ def __getattribute__(self, attr):
     FAKES = _fakes()
 
 
+if get_backend() == "jax":
+    _expanded_dist_path = "backend_dist.ExpandedDistribution"
+elif get_backend() == "torch":
+    _expanded_dist_path = "backend_dist.torch_distribution.ExpandedDistribution"
```
eb8680, line 46: Looks like you need to define a dummy value of `_expanded_dist_path` when `get_backend() == "numpy"` to get test collection to work, even though the tests themselves are all skipped.

fehiepsi: Thanks Eli!

eb8680 (approved):

Issue #412, opened 2020-12-22
Implement distribution shape helpers to avoid need for dummy tensors


fritzo:
As mentioned in https://github.com/pyro-ppl/funsor/pull/402#discussion_r542716992 we should add distribution methods upstream to perform all needed shape operations.

Currently Funsor's distribution shaping logic uses the pattern of creating fake distributions with dummy tensors to determine distribution event shape and batch shape. As a goal for Funsor use in Pyro 2.0, we should eliminate the need to ever instantiate a dummy tensor, and should eliminate the tensor shape caches.

## Tasks
- [x] #420 decide on shape helper interface
- [x] #420 refactor Funsor to use the new methods instead of fake distributions
- [x] https://github.com/pyro-ppl/pyro/pull/2739 add to Pyro / PyTorch distributions
- [x] https://github.com/pyro-ppl/numpyro/pull/901 add to NumPyro / JAX / TFP distributions

eb8680:
Now that https://github.com/pyro-ppl/pyro/pull/2739 has landed, what's left to do on the Funsor side?

fritzo:
I think it's safe to close this. Possible future tasks include:
- being a little stricter as in #426 (but this is not necessary)
- supporting dependent types via `Value`, which would be required for e.g. `LKJCholesky`

PR #420, opened 2021-01-11
Avoid creating distributions when inferring value domain

fritzo:
Addresses #412
pair coded with @eb8680 

This refactors logic in `Distribution._infer_value_domain()` to avoid creating a temporary dummy distribution *if* backend distributions implement a `.infer_shapes()` method.

### DIFF
--- a/funsor/memoize.py
+++ b/funsor/memoize.py
@@ -17,7 +17,7 @@ def memoize(cache=None):
 
     @interpreter.interpretation(interpreter._INTERPRETATION)  # use base
     def memoize_interpretation(cls, *args):
-        key = (cls,) + tuple(id(arg) if (type(arg).__name__ == "DeviceArray") or not isinstance(arg, Hashable)
+        key = (cls,) + tuple(id(arg) if ("DeviceArray" in type(arg).__name__) or not isinstance(arg, Hashable)
                              else arg for arg in args)
         if key not in cache:
             cache[key] = cls(*args)
--- a/funsor/memoize.py
+++ b/funsor/memoize.py
@@ -5,6 +5,7 @@ from collections.abc import Hashable
 from contextlib import contextmanager
 
 import funsor.interpreter as interpreter
+from funsor.util import get_backend
 
 
 @contextmanager
@@ -17,8 +18,16 @@ def memoize(cache=None):
 
     @interpreter.interpretation(interpreter._INTERPRETATION)  # use base
     def memoize_interpretation(cls, *args):
-        key = (cls,) + tuple(id(arg) if ("DeviceArray" in type(arg).__name__) or not isinstance(arg, Hashable)
-                             else arg for arg in args)
+        # JAX DeviceArray has .__hash__ method but raise the unhashable error there.
+        if get_backend() == "jax":
+            import jax
+
+            key = tuple(id(arg)
+                        if isinstance(arg, jax.interpreters.xla.DeviceArray)
+                        or not isinstance(arg, Hashable)
+                        else arg for arg in args)
+        else:
+            key = tuple(id(arg) if not isinstance(arg, Hashable) else arg for arg in args)
         if key not in cache:
             cache[key] = cls(*args)
         return cache[key]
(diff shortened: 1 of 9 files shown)


funsor/torch/distributions.py
```diff
@@ -115,16 +115,24 @@ def _infer_value_domain(**kwargs):
     return kwargs['v']
 
 
+@methodof(Categorical)  # noqa: F821
```
eb8680, line 118: Add `methodof(CategoricalLogits)`

fritzo:
Test failures appear unrelated.
```
...
from jaxlib import cusolver[0m
ImportError: cannot import name 'cusolver'[0m
Makefile:46: recipe for target 'test' failed
```
@fehiepsi any suggestions?

eb8680:
Looks like a new JAX bug? https://github.com/google/jax/issues/5374

eb8680:
@fehiepsi it looks like a different new JAX error related to hashing of `DeviceArray`s is causing the build to fail, any idea what's going on?

fehiepsi:
@fritzo Could you allow me rerun the build? Sometimes I would like to do it, for example to return this PR with the new jax release.

fehiepsi:
@eb8680 Opps, I missed your comment. Will talk a look soon...

fritzo:
> @fritzo Could you allow me to rerun the build?

@fehiepsi sure, feel free to rerun the builds. Do I need to authorize you on travis?

funsor/memoize.py
```diff
@@ -17,7 +17,7 @@ def memoize(cache=None):
 
     @interpreter.interpretation(interpreter._INTERPRETATION)  # use base
     def memoize_interpretation(cls, *args):
-        key = (cls,) + tuple(id(arg) if (type(arg).__name__ == "DeviceArray") or not isinstance(arg, Hashable)
+        key = (cls,) + tuple(id(arg) if ("DeviceArray" in type(arg).__name__) or not isinstance(arg, Hashable)
```
fehiepsi, line 20: At some point ago, `type(jax.numpy.ones(3)).__name__` is `"_DeviceArray"`

funsor/tensor.py
```diff
@@ -88,6 +88,16 @@ def __call__(cls, data, inputs=None, dtype="real"):
...
             data = data.__array__()
+        # XXX: in jaxlib 0.1.58 + jax 0.2.8, sometime opt_einsum returns
+        # a jaxlib.xla_client.Buffer, which is a type not supported in Tensor.
+        # It is unclear whether this is an issue. The good thing is: under jit,
+        # data will be a ShapedArray, so we won't go to this branch to do
+        # the extra job `jax.numpy.array(data)`.
+        elif "Buffer" in type(data).__name__ and get_backend() == "jax":
+            import jax
+
+            data = jax.numpy.array(data)
```
eb8680, line 99: Do we need to add something like this to `funsor.jax.ops.einsum` as well?

fehiepsi: @hawkinsp In https://github.com/google/jax/pull/2839, you mentioned that `xla_client.Buffer` is going to be deleted. But in the latest jax+jaxlib release, we are seeing that `Buffer` starts appearing (it didn't appear in previous releases). I just wonder if it is will be supported in the long run?

hawkinsp: Ah, this is a different `Buffer`. The one you are talking about was a Python class. The new one is a C++ class.

To improve dispatch times, we are moving a few parts of the JAX dispatch logic into C++, and in some cases (notably, uncomplicated `jit` calls), you will get a pure C++ object out (`Buffer`) without the additional Python wrapper class. It should duck type the same (e.g., have the same methods and act the same in most respects).

I believe both the Python and C++ variants report themselves as being instances of `DeviceArray`. Eventually we will only have the C++ version.

fritzo:
@fehiepsi thanks for the help!

funsor/tensor.py
fehiepsi: Thanks for the very clear explanation, @hawkinsp! I understand what's going on now.

PR #421, opened 2021-01-12
Unify op-caching metaclasses

eb8680:
Addresses #351. ~~Blocked by #420.~~

Some ops, such as `ops.reshape` and `ops.getitem`, take non-Funsor arguments in their constructors, and make use of custom metaclasses to memoize class creation.  This PR replaces those metaclasses with a single metaclass `funsor.ops.op.OpCacheMeta`.

The motivation for this PR is that the plan in #351 includes adding a `Finitary` funsor term representing lazy op application and refactoring `funsor.function` to generate new op classes and type inference rules on the fly, and we will want to generate these classes with something like `OpCacheMeta` to ensure that they are compatible with Funsor cons-hashing.

`OpCacheMeta` will also be useful for extending support for lazy evaluation on funsors to more ops, and for refactoring some funsor terms that are conceptually closer to an op than a first-class funsor.  For example, as part of #351 we will want to replace the Funsor term `funsor.tensor.Einsum` with an op class `EinsumOp` parametrized by `equation`, similar to how `ops.reshape` is parametrized by the target shape.

Tested:
- Exercised by existing tests of `reshape` and `getitem`

### DIFF
--- a/funsor/ops/array.py
+++ b/funsor/ops/array.py
@@ -6,7 +6,7 @@ import math
 import numpy as np
 
 from .builtin import AssociativeOp, add, atanh, exp, log, log1p, max, min, reciprocal, safediv, safesub, sqrt, tanh
-from .op import DISTRIBUTIVE_OPS, Op
+from .op import DISTRIBUTIVE_OPS, Op, OpCacheMeta
 
 _builtin_all = all
 _builtin_any = any
@@ -66,20 +66,7 @@ logaddexp = LogAddExpOp(_logaddexp, name="logaddexp")
 sample = SampleOp(_logaddexp, name="sample")
 
 
-class ReshapeMeta(type):
-    _cache = {}
-
-    def __call__(cls, shape):
-        shape = tuple(shape)
-        try:
-            return ReshapeMeta._cache[shape]
-        except KeyError:
-            instance = super().__call__(shape)
-            ReshapeMeta._cache[shape] = instance
-            return instance
-
-
-class ReshapeOp(Op, metaclass=ReshapeMeta):
+class ReshapeOp(Op, metaclass=OpCacheMeta):
     def __init__(self, shape):
         self.shape = shape
         super().__init__(self._default)
--- a/funsor/ops/array.py
+++ b/funsor/ops/array.py
@@ -6,7 +6,7 @@ import math
 import numpy as np
 
 from .builtin import AssociativeOp, add, atanh, exp, log, log1p, max, min, reciprocal, safediv, safesub, sqrt, tanh
-from .op import DISTRIBUTIVE_OPS, Op, OpCacheMeta
+from .op import DISTRIBUTIVE_OPS, Op, CachedOpMeta
 
 _builtin_all = all
 _builtin_any = any
@@ -66,7 +66,13 @@ logaddexp = LogAddExpOp(_logaddexp, name="logaddexp")
 sample = SampleOp(_logaddexp, name="sample")
 
 
-class ReshapeOp(Op, metaclass=OpCacheMeta):
+class ReshapeMeta(CachedOpMeta):
+    def __call__(cls, shape):
+        shape = tuple(shape)  # necessary to convert torch.Size to tuple
+        return super().__call__(shape)
+
+
+class ReshapeOp(Op, metaclass=ReshapeMeta):
     def __init__(self, shape):
         self.shape = shape
         super().__init__(self._default)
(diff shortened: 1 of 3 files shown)


funsor/ops/op.py
```diff
@@ -1,9 +1,30 @@
...
+class OpCacheMeta(type):
+    """
+    Metaclass for caching op instance construction.
+    """
+    # use a WeakKeyDictionary to allow garbage collection of out-of-scope op classes.
+    # this is especially important when some ops are dynamically generated and
+    # their instances are large, e.g. an op corresponding to a neural network.
+    _cls_caches = weakref.WeakKeyDictionary()  # class -> {args: instance}
+
+    def __call__(cls, *args, **kwargs):
```
fritzo, line 18: Unless there is a very good reason to allow `**kwargs`, I would recommend removing it and forcing users to pass positional arguments.

funsor/ops/op.py
```diff
@@ -1,9 +1,30 @@
 # Copyright Contributors to the Pyro project.
 # SPDX-License-Identifier: Apache-2.0
 
+import weakref
+
 from multipledispatch import Dispatcher
 
 
+class OpCacheMeta(type):
```
fritzo, line 9: nit: rename to `CachedOpMeta` or `MemoizedOpMeta`

funsor/ops/op.py
```diff
@@ -1,9 +1,30 @@
...
+    """
+    Metaclass for caching op instance construction.
+    """
+    # use a WeakKeyDictionary to allow garbage collection of out-of-scope op classes.
+    # this is especially important when some ops are dynamically generated and
+    # their instances are large, e.g. an op corresponding to a neural network.
+    _cls_caches = weakref.WeakKeyDictionary()  # class -> {args: instance}
```
fritzo, line 16: Replace this with a per-class dictionary that is created during `.__init_subclass__()`, thereby more closely mimicking existing behavior.
```py
def __init_subclass__(cls, default_name, **kwargs):
    super().__init_subclass__(**kwargs)
    cls._cache = {}
```

funsor/ops/op.py
```diff
@@ -1,9 +1,30 @@
...
+
+    def __call__(cls, *args, **kwargs):
+        if cls not in OpCacheMeta._cls_caches:
+            OpCacheMeta._cls_caches[cls] = {}
+        cls_cache = OpCacheMeta._cls_caches[cls]
+        key = tuple(args) + tuple(kwargs.items())
+        if key not in cls_cache:
+            cls_cache[key] = super(OpCacheMeta, cls).__call__(*args, **kwargs)
+        return cls_cache[key]
```
fritzo, line 25: Why not preserve the existing logic, which was highly optimized?
```py
        try:	
            return cls._cache[args]	
        except KeyError:
            instance = super(OpCacheMeta, cls).__call__(*args)	
            cls._cache[args] = instance
            return instance
```

funsor/ops/array.py
```diff
@@ -66,20 +66,7 @@ def _logaddexp(x, y):
 sample = SampleOp(_logaddexp, name="sample")
 
 
-class ReshapeMeta(type):
-    _cache = {}
-
-    def __call__(cls, shape):
-        shape = tuple(shape)
```
fritzo, line 73: I think we'll still need this line to be able to jit PyTorch functions.

eb8680: Ok, I put this back into a simpler `ReshapeMeta`. When we refactor `function`, we may want to move this logic inside of `CachedOpMeta` and always convert any args of `torch.Size` to tuples.

eb8680:
I think we'll need to merge #420 with @fehiepsi's JAX fixes before this can be merged.

funsor/ops/op.py
```diff
@@ -4,7 +4,25 @@
...
+            cls._cache[args] = instance
+            return instance
+
+
 class Op(Dispatcher):
+
+    def __init_subclass__(cls, **kwargs):
+        super().__init_subclass__(**kwargs)
+        cls._cache = {}
+
```
fritzo, line 25: nit: I think you'll want to split this up into
```py
class CachedOp(Op, metaclass=CachedOpMeta):
    def __init_subclass__(cls, **kwargs):
        ...

class CachedOp(metaclass=CachedOpMeta):
    ...

class GetitemOp(CachedOp):  # Yay, no need for metaclass!
    ...
```

eb8680: The reason I didn't do this was that I wanted to keep caching out of the `Op` class hierarchy initially.  I also suspect we will want most ops to be cached in the future as an extra layer of robustness in cons-hashing and serialization, even though most common op constructors should only ever be called once anyway.

fritzo (approved):
LGTM

PR #422, opened 2021-01-12
Break up type inference rules in find_domain into op-level patterns

eb8680:
Addresses #351. ~~Blocked by #421.~~

This PR breaks up the monolithic `funsor.domains.find_domain` into a `singledispatch` registry of patterns for individual ops, adds a new helper function `funsor.ops.op.make_op_and_type` for generating new op classes programmatically, and applies this helper liberally throughout the base op definition files.

The motivation for this change to `find_domain` is that it is now possible to register additional rules in `find_domain` for ops created elsewhere and to represent lazy application of any op with a single `Finitary` funsor. In particular, this is necessary for the plan in #351 to have `funsor.function` represent functions as newly created ops whose `find_domain` implementation can be read off from their type signatures.  `make_op_and_type` will also be useful for pattern-matching more generally, since most basic ops will now have their own classes.

Tested:
- Exercised by existing tests

### DIFF
--- a/funsor/adjoint.py
+++ b/funsor/adjoint.py
@@ -256,7 +256,7 @@ def _scatter(src, res, subs):
     return Tensor(data, inputs, res.dtype)
 
 
-@adjoint_ops.register(Subs, ops.LogAddExpOp, ops.AddOp, GaussianMixture, GaussianMixture, tuple)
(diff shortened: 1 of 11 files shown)


fritzo (approved):
LGTM as long as tests pass.

Issue #355, opened 2020-08-17
Make interpreters compatible with Python type-checking


eb8680:
Currently, while Funsor makes heavy use of `multipledispatch` on parametric types, interpretations and Funsor terms are not compatible with static type-checkers like MyPy or even runtime type-checkers like [pytypes](https://github.com/Stewori/pytypes).  It would improve the development experience, increase performance and beneficially constrain Funsor's design to make Funsor interpreters and lazy expressions more compatible with Python's standard `typing` module and even with MyPy.

At a high level, this seems straightforward: just make all subclasses of `funsor.Funsor` inherit from `typing.Generic` and move type annotations from `register` calls to function signatures, e.g.:
```py
Op = TypeVar("Op", bound=ops.AssociativeOp)
Arg = TypeVar("Arg", bound=Funsor)
ReducedVars = TypeVar("ReducedVars", bound=FrozenSet[str])

class Reduce(Funsor, Generic[Op, Arg, ReducedVars]):
    def __init__(self, op: Op, arg: Arg, reduced_vars: ReducedVars):
        ...  # note we should be able to remove all hand-written type assertions here

Lhs = TypeVar("Lhs", bound=Funsor)
Rhs = TypeVar("Rhs", bound=Funsor)

class Binary(Funsor, Generic[Op, Lhs, Rhs]):
    def __init__(self, op, lhs, rhs):
        ...

# a random example pattern (a+b).sum() -> (a.sum() + b.sum())
@eager.register(Reduce)
def eager_reduce_example_pattern(
        op: ops.AddOp, 
        arg: Binary[ops.AddOp, Funsor, Funsor],
        reduced_vars: FrozenSet[str]
    ):
    return Binary(arg.op, Reduce(op, arg.lhs, reduced_vars), Reduce(op, arg.rhs, reduced_vars))
```
We'd also have to modify `KeyedRegistry` along the lines of https://github.com/mrocklin/multipledispatch/pull/69 using `pytypes.deep_type` for dispatch to continue to work correctly, and define `FunsorMeta.__subclasscheck__` with `pytypes.is_subtype` rather than the custom logic it currently contains.

[...]

PR #423, opened 2021-01-12
Add Finitary funsor for representing lazy op application

eb8680:
Addresses #351. ~~Blocked by #421, #422.~~

This PR adds a new term `funsor.terms.Finitary` that represents lazy op application, and refactors `funsor.terms.Einsum` into a new op `EinsumOp` that makes use of the `find_domain` extensibility added in #422.

Tested:
- Exercised a bit by existing `Einsum` tests


fritzo (approved):
LGTM

funsor/tensor.py
```diff
@@ -695,6 +696,18 @@ def eager_getitem_tensor_tensor(op, lhs, rhs):
     return Tensor(data, inputs, lhs.dtype)
 
 
+# TODO handle variable length Tuple
+@eager.register(Finitary, Op, typing.Tuple[Tensor])
```
eb8680, line 700: Note the use of `typing.Tuple` in head patterns like this is not currently supported, and will have to be worked around like we did with `Contraction`.  This is what got me thinking about #355, #433 and #434 again.

eb8680: Added support for this in #451.

test/test_distribution.py
```diff
@@ -639,7 +639,7 @@ def test_mvn_affine_einsum():
     data = dict(x=Tensor(randn(2, 2)), y=Tensor(randn(())))
     with lazy:
         d = to_funsor(random_mvn((), 3), Real)
-        d = d(value=Einsum("abc,bc->a", c, x) + y)
+        d = d(value=Einsum("abc,bc->a", (c, x)) + y)
```
fritzo, line 642: nit: if `Einsum` is just a wrapper, how about letting it match the standard `np.einsum` syntax via `*args`?

funsor/tensor.py
```diff
@@ -1019,7 +1027,30 @@ def max_and_argmax(x: Reals[8]) -> Tuple[Real, Bint[8]]:
...
+        for name, size in zip(ein_input, x.shape):
+            other_size = size_dict.setdefault(name, size)
+            if other_size != size:
+                raise ValueError(
+                    "Size mismatch at {}: {} vs {}".format(name, size, other_size)
+                )
+    return Reals[tuple(size_dict[d] for d in ein_output)]
+
+
+def Einsum(equation, operands):
```
fritzo, line 1053: nit: How about standardizing as `def Einsum(equaiton, *operands):`?

fritzo (approved):
This is a great simplification!

Only one nit on keeping the standard `np.einsum` interface

PR #427, opened 2021-01-17
Add factory for dynamic TransformOp

fritzo:
pair coded with @eb8680 @fehiepsi @ordabayevy 

This introduces classes `WrappedTransformOp` and `LogAbsDetJacobianOp` for dynamically created ops from backend `Transform` objects and `TransformedDistribution`s. Unlike statically created ops, instances of these ops will never have custom rules; hence we do not create unique subclasses for each new op.

I have also refactored `Op` and `CachedOpMeta` and `WrappedOpMeta` to more finely control whether and how we cache op instance creation.

## Forwards compatibility

This PR xfails until the following PRs are implemented upstream. I have tested locally with these changes. It is safe to merge this PR before the upstream changes have merged, because the upstream interfaces are pretty stable:
- https://github.com/pytorch/pytorch/pull/50547 PyTorch `Constraint.event_dim`
- https://github.com/pytorch/pytorch/pull/50581 PyTorch `Transform.forward_shape()`
- https://github.com/pyro-ppl/pyro/pull/2739 Pyro `Constraint.event_dim` and `Transform.forward_shape()`
- https://github.com/pyro-ppl/numpyro/pull/876 NumPyro `Constraint.event_dim`
- https://github.com/pyro-ppl/numpyro/pull/887 NumPyro `Transform.forward_shape()`

## Tested
- [x] cache test for `WrappedTransformOp` and `LogAbsDetJacobianOp`
- [x] gc test for `WrappedTransformOp`
- [x] distribution test with `PowerTransform` (passes locally)
- [x] distribution test with `HaarTransorm` (passes locally)

### DIFF
--- a/funsor/ops/op.py
+++ b/funsor/ops/op.py
@@ -156,8 +156,9 @@ class TransformOp(UnaryOp):
 
 class ValidatedTransformOp(TransformOp):
     """
-    Like :class:`TransformOp` but additionally checks that the backing tranforn
-    is not batched. This check is performed only on the first :meth:`__call__`.
+    Like :class:`TransformOp` but additionally checks that the backing
+    transform is not batched. This check is performed only on the first
+    :meth:`__call__`.
     """
     def __init__(self, fn, *, name=None):
         super().__init__(fn, name=name)
@@ -171,7 +172,8 @@ class ValidatedTransformOp(TransformOp):
             event_shape = x.shape[len(x.shape) - fn.domain.event_dim:]
             shape = fn.forward_shape(event_shape)
             if len(shape) > fn.codomain.event_dim:
-                raise ValueError(f"Cannot treat batched transform {self.name} as an Op")
+                raise ValueError(f"Cannot treat transform {self.name} as an Op "
+                                 "because it is batched")
         return super().__call__(x)
 
 
--- a/funsor/ops/op.py
+++ b/funsor/ops/op.py
@@ -8,6 +8,20 @@ import weakref
 from multipledispatch import Dispatcher
 
 
+class WeakPartial:
+    """
+    Like ``functools.partial(fn, arg)`` but weakly referencing ``arg``.
+    """
+    def __init__(self, fn, arg):
+        self.fn = fn
+        self.weak_arg = weakref.ref(arg)
+        functools.update_wrapper(self, fn)
+
+    def __call__(self, *args):
+        arg = self.weak_arg()
+        return self.fn(arg, *args)
+
+
 class CachedOpMeta(type):
     """
     Metaclass for caching op instance construction.
@@ -71,7 +85,7 @@ class Op(Dispatcher):
         # Register all existing patterns.
         for supercls in reversed(inspect.getmro(type(self))):
             for pattern, fn in getattr(supercls, "_subclass_registry", ()):
-                self.register(*pattern)(functools.partial(fn, self))
+                self.add(pattern, WeakPartial(fn, self))
         # Save self for registering future patterns.
         Op._all_instances.add(self)
 
@@ -96,7 +110,7 @@ class Op(Dispatcher):
             # Register with all existing instances.
             for op in Op._all_instances:
                 if isinstance(op, cls):
-                    op.register(*pattern)(functools.partial(fn, op))
+                    op.add(pattern, WeakPartial(fn, op))
             # Ensure registration with all future instances.
             cls._subclass_registry.append((pattern, fn))
             return fn
--- a/funsor/ops/op.py
+++ b/funsor/ops/op.py
@@ -154,6 +154,27 @@ class TransformOp(UnaryOp):
         raise NotImplementedError
 
 
+class ValidatedTransformOp(TransformOp):
+    """
+    Like :class:`TransformOp` but additionally checks that the backing tranforn
+    is not batched. This check is performed only on the first :meth:`__call__`.
+    """
+    def __init__(self, fn, *, name=None):
+        super().__init__(fn, name=name)
+        self._pending_validation = [fn]
+
+    def __call__(self, x):
+        while self._pending_validation:
+            fn = self._pending_validation.pop()
+            if len(x.shape) < fn.domain.event_dim:
+                raise ValueError(f"Too few dimensions for input, in {self.name}")
+            event_shape = x.shape[len(x.shape) - fn.domain.event_dim:]
+            shape = fn.forward_shape(event_shape)
+            if len(shape) > fn.codomain.event_dim:
+                raise ValueError(f"Cannot treat batched transform {self.name} as an Op")
+        return super().__call__(x)
+
+
 class LogAbsDetJacobianOp(Op):
     pass
 
@@ -172,18 +193,12 @@ def make_transform_op(backend_transform):
     """
     name = type(backend_transform).__name__
 
-    # TODO Check that the op is not batched. Something like:
-    # if backend_transform.batch_shape:
-    #     raise ValueError("Cannot create an op from a transform "
-    #                      f"{name} with nontrivial batch shape "
-    #                      f"{backend_transform.batch_shape}.")
-
     # Create four ops.
-    op = make_op(backend_transform, TransformOp, name=name)
+    op = make_op(backend_transform, ValidatedTransformOp, name=name)
     op_ldaj = make_op(backend_transform.log_abs_det_jacobian,
                       LogAbsDetJacobianOp,
                       name=name + "_log_abs_det_jacobian")
-    inv = make_op(backend_transform.inv, TransformOp,
+    inv = make_op(backend_transform.inv, ValidatedTransformOp,
                   name=name + "_inv")
     inv_ldaj = make_op(backend_transform.log_abs_det_jacobian,
                        LogAbsDetJacobianOp,
--- a/funsor/ops/op.py
+++ b/funsor/ops/op.py
@@ -205,6 +205,7 @@ class WrappedTransformOp(TransformOp, metaclass=WrappedOpMeta):
     def __call__(self, x):
         if self._is_validated:
             return super().__call__(x)
+
         try:
             # Check for shape metadata available after
             # https://github.com/pytorch/pytorch/pull/50547
@@ -215,9 +216,10 @@ class WrappedTransformOp(TransformOp, metaclass=WrappedOpMeta):
             self.fn.codomain.event_dim
             self.fn.forward_shape
         except AttributeError:
-            backend = self.fn.__module__.split(">")[0]
+            backend = self.fn.__module__.split(".")[0]
             raise NotImplementedError(f"{self.fn} is missing shape metadata; "
                                       f"try upgrading backend {backend}")
+
         if len(x.shape) < self.fn.domain.event_dim:
             raise ValueError(f"Too few dimensions for input, in {self.name}")
         event_shape = x.shape[len(x.shape) - self.fn.domain.event_dim:]
--- a/funsor/ops/op.py
+++ b/funsor/ops/op.py
@@ -210,13 +210,6 @@ def make_transform_op(backend_transform):
     inv.set_inv(op)
     inv.set_log_abs_det_jacobian(inv_ldaj)
 
-    # Register funsor conversions.
-    from funsor.terms import Binary, Funsor, Unary
-    op.register(Funsor)(functools.partial(Unary, op))
-    inv.register(Funsor)(functools.partial(Unary, inv))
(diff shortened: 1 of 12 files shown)


fritzo:
Recording discussion from slack:

> Du Phan: Do you expect that we will change anything in those to_funsor(ExpTransform) https://github.com/pyro-ppl/funsor/pull/427/files#diff-2aa6bbdd9f351b5925977d25eacad85b7fb658e38f07a4e1b33efa76d47d4b50R277 ? It seems to me that the default transform_to_funsor should take care of the job .

> Fritz Obermeyer: If we wanted to replace the existing ExpOp with the new transform_to_funsor() logic, we would need to address caching and .__hash__(). That is a complex issue because PyTorch already uses the .__eq__() method for expensive tensor comparisons, and that usage is incompatible with lightweight caching. However I see no need to address this issue right away. @fehiepsi what is your motivation?

> Du Phan: Ah, I just thought that if to_funsor(ExpTransfor)  is not needed, then implementing https://github.com/pyro-ppl/funsor/pull/365 for jax backend is not needed.

> Fritz Obermeyer: Yeah, let's just think about this more. A naive solution might simply define a custom hash method like
> ```py
> def hash_op(fn):
>     if isinstance(fn, backend_dist.transforms.Transform):
>         return id(fn)  # Hash Transform instances by id.
>     return hash(fn)
> ```
> that would at least allow us to delete existing custom to_funsor() implementations for singleton Transform s. But that would > still leak memory when we start using flows.

test/test_ops.py
```diff
@@ -0,0 +1,38 @@
...
+    except KeyError:
+        pytest.skip(f"missing distributions module for {get_backend()}")
+    return pytest.importorskip(module_name).dist
+
+
+def test_transform_op_cache(dist):
+    t = dist.transforms.PowerTransform(0.5)
+    W = WrappedTransformOp
+    assert W(t) is W(t)
+    assert W(t).inv is W(t).inv
```
eb8680, line 26: Looks like this test is failing on the jax backend because `_InverseTransform` isn't cached in NumPyro?

fritzo:
@fehiepsi 
> It seems to me that the default transform_to_funsor should take care of the job .

Yes, I think I better understand your point now. This PR now adds a default `to_transform(numpyro.distributions.Transform)` that should work on all transforms *that are not batched*. We would need to register additional patterns for batching to work. That should be easy in a subsequent PR: we can move the patterns from funsor/torch/distributions.py up to funsor/distribution.py and then register those patterns in funsor/jax/distributions.py

fehiepsi:
> will need to add a Transform.forward_shape() method in NumPyro

@fritzo This will resolve current issues at TransformedDistribution and init_strategy so I want to make a PR for it. But I am a bit worried if it will be enough for funsor, in complicated situations like normalization flows that Eli pointed out in our last discussions. Do we need batch shape or something else?

fritzo:
> Do we need batch shape or something else?

I believe `.batch_shape` is unnecessary for `TransformedDistribution`. I believe we need only a small interface to fully support normalizing flows and batched transforms inside `TransformedDistribution`:
- `Transform.domain.event_dim` and `Transform.codomain.event_dim` (done)
- `Transform.forward_shape()` and `.inverse_shape()` as in https://github.com/pytorch/pytorch/pull/50581
- `TransformedDistribution.event_shape` and `.batch_shape` as correctly computed in https://github.com/pytorch/pytorch/pull/50581

That would be sufficient metadata to implement `to_funsor(TransformedDistribution)`, but we still haven't implemented the generic conversion rule in Funsor. If we additionally want to construct flows with lazy inputs, we would need at least:
- `Distribution.infer_shapes()` as in https://github.com/pyro-ppl/pyro/pull/2739

funsor/ops/op.py
```diff
@@ -8,25 +8,66 @@
...
 
 
+class WeakPartial:
+    """
+    Like ``functools.partial(fn, arg)`` but weakly referencing ``arg``.
+    """
+    def __init__(self, fn, arg):
+        self.fn = fn
+        self.weak_arg = weakref.ref(arg)
+        functools.update_wrapper(self, fn)
```
fehiepsi, line 18: One day, I'll learn what `update_wrapper`, `weakref` do... >___<

fehiepsi (commented):
LGTM, the implementation fits my intuition (for now, I couldn't evaluate the caching and weakref details).

test/test_ops.py
```diff
@@ -0,0 +1,38 @@
...
+
+import pytest
+
+from funsor.distribution import BACKEND_TO_DISTRIBUTIONS_BACKEND
+from funsor.util import get_backend
+from funsor.ops import WrappedTransformOp
+
+
+@pytest.fixture
+def dist():
```
eb8680, line 14: This seems like a useful pattern, we may want to copy it in the distribution tests in a later PR.

test/test_distribution.py
```diff
@@ -1215,3 +1225,48 @@ def test_categorical_event_dim_conversion(batch_shape, event_shape):
         funsor.to_data(data, name_to_dim=name_to_dim))
     assert actual_log_prob.shape == expected_log_prob.shape
     assert_close(actual_log_prob, expected_log_prob)
+
+
+@xfail_if_not_implemented()
+@pytest.mark.parametrize("shape", [(), (4,), (3, 2)], ids=str)
+def test_power_transform(shape):
```
eb8680, line 1232: We should also test the other `funsor.distribution.Distribution` methods, especially sampling.

I believe you could achieve that by just adding the following test case to the list of test cases in `test/test_distribution_generic.py`:
```py
DistTestCase(
    "backend_dist.TransformedDistribution(backend_dist.Exponential(rate=case.rate), [backend_dist.transforms.PowerTransform(0.5),])",  # noqa: E501
    (("rate", f"rand({batch_shape})"),),
    funsor.Real,
    xfail_reason="backend not supported" if get_backend() != "torch" else "",
)
```

test/test_distribution.py
```diff
@@ -1215,3 +1225,48 @@ def test_categorical_event_dim_conversion(batch_shape, event_shape):
...
+    assert_close(actual_log_prob, expected_log_prob)
+
+
+@xfail_if_not_implemented()
+@pytest.mark.parametrize("shape", [(10,), (4, 3)], ids=str)
+@pytest.mark.parametrize("to_event", [
+    True,
+    xfail_param(False, reason="bug in to_funsor(TransformedDistribution)"),
+])
+def test_haar_transform(shape, to_event):
```
eb8680, line 1255: Ditto: try adding a `HaarTransform` test case to `test/test_distribution_generic.py` to test sampling, `to_funsor`/`to_data` conversion and other distribution methods?

funsor/ops/op.py
```diff
@@ -8,25 +8,66 @@
 from multipledispatch import Dispatcher
 
 
+class WeakPartial:
```
eb8680, line 11: nit: consider moving `WeakPartial` to `funsor.util`?

eb8680 (commented):
Code all looks good to me, but it would be good to add new `TransformedDistribution` test cases to our generic distribution tests to make sure sampling and other methods also behave correctly.  If getting those working turns out to be a hassle we can defer to a followup PR.

test/test_distribution.py
fritzo: Can you suggest a fix for this broken `test_generic_distribution_to_funsor()`:

<details>

```
$ FUNSOR_BACKEND=torch pytest -vx test/test_distribution_generic.py -k Haar --pdb
===================================================== test session starts ======================================================
platform darwin -- Python 3.7.0, pytest-6.1.2, py-1.9.0, pluggy-0.13.1 -- /Users/fobermey/opt/miniconda3/envs/pyro/bin/python
cachedir: .pytest_cache
benchmark: 3.2.3 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /Users/fobermey/github/pyro-ppl/funsor, configfile: setup.cfg
plugins: forked-1.2.0, nbval-0.9.6, xdist-2.1.0, benchmark-3.2.3
collected 2244 items / 2211 deselected / 33 selected

test/test_distribution_generic.py::test_generic_distribution_to_funsor[dist.TransformedDistribution( dist.Normal(loc=case.loc, scale=1.).to_event(1), dist.transforms.HaarTransform(dim=-1)) (('loc', 'rand(() + (3,))'),)] FAILED [  3%]
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> traceback >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

case = <test.test_distribution_generic.DistTestCase object at 0x7fb9b08c8cc0>

    @pytest.mark.parametrize("case", TEST_CASES, ids=str)
    def test_generic_distribution_to_funsor(case):

        HIGHER_ORDER_DISTS = [
            backend_dist.Independent,
            backend_dist.TransformedDistribution,
        ] + ([backend_dist.torch_distribution.ExpandedDistribution] if get_backend() == "torch"
             else [backend_dist.ExpandedDistribution])

        raw_dist = case.get_dist()
        expected_value_domain = case.expected_value_domain

        dim_to_name, name_to_dim = _default_dim_to_name(raw_dist.batch_shape)
        with interpretation(normalize_with_subs):
            funsor_dist = to_funsor(raw_dist, output=funsor.Real, dim_to_name=dim_to_name)
        assert funsor_dist.inputs["value"] == expected_value_domain

        while isinstance(funsor_dist, funsor.cnf.Contraction):
            funsor_dist = [term for term in funsor_dist.terms
>                          if isinstance(term, (funsor.distribution.Distribution, funsor.terms.Independent))][0]
E           IndexError: list index out of range

test/test_distribution_generic.py:590: IndexError
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> entering PDB >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB post_mortem (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> /Users/fobermey/github/pyro-ppl/funsor/test/test_distribution_generic.py(590)test_generic_distribution_to_funsor()
-> if isinstance(term, (funsor.distribution.Distribution, funsor.terms.Independent))][0]
(Pdb) print(funsor_dist.pretty())
Contraction(ops.nullop, ops.add,
 frozenset(),
 (Unary(ops.neg,
   Binary(ops.log_abs_det_jacobian,
   │Unary(ops._InverseTransform,
   │ Variable('value', Reals[3])),
   │Variable('value', Reals[3]))),
  Contraction(ops.add, ops.nullop,
   frozenset({Variable('_pyro_event_dim_-1_...
   (Normal(
   │ Tensor(
   │  torch.tensor([0.4366315007209778, 0.8093...
   │  (('_pyro_event_dim_-1__BOUND_1',
   │   │Bint[3, ],),),
   │  'real'),
   │ Tensor(
   │  torch.tensor([1.0, 1.0, 1.0], dtype=torc...
   │  (('_pyro_event_dim_-1__BOUND_1',
   │   │Bint[3, ],),),
   │  'real'),
   │ Binary(ops.GetitemOp(0),
   │  Unary(ops._InverseTransform,
   │   Variable('value', Reals[3])),
   │  Variable('_pyro_event_dim_-1__BOUND_1', ...
```

</details>

funsor/ops/op.py
fritzo: done

test/test_distribution.py
fritzo: done

test/test_distribution.py
eb8680: I think the underlying reason is that the way we use a lazy `funsor.terms.Independent` to convert `IndependentDistribution`s to funsors  (via the pattern `funsor.distribution.indepdist_to_funsor`) is sort of fiddly and affects pattern matching, especially when combined with transforms. Now that #402 is done we should try to represent them directly by `ops.expand`ing their parameters to the appropriate `event_shape`.  I'll see if I can get this working in a separate PR.

fritzo:
@eb8680 thanks for #443! This PR now passes locally and should be safe to merge.

eb8680 (approved):
LGTM

PR #428, opened 2021-01-18
Forward-backward algorithm as differentiation of the backward algorithm

ordabayevy:
This example shows how the `adjoint` module can be used to obtain forward-backward algorithm by differentiating the backward algorithm.


ordabayevy:
@eb8680 can you review this?

eb8680:
@ordabayevy you'll need to merge `master` to appease Travis. Can you also add a line for this example to the [`Makefile` so it runs under `make test`](https://github.com/pyro-ppl/funsor/blob/master/Makefile#L31)?

examples/forward_backward.py
```diff
@@ -0,0 +1,187 @@
...
+    Z_backward, betas = backward_algorithm(
+        sum_op, prod_op, init, trans, time, {"x_prev": "x_curr"}
+    )
+    assert_close(Z_forward, Z_backward)
+    gammas = []
+    for alpha, beta in zip(alphas, betas):
+        gamma = prod_op(alpha(**{"x_curr": "x_prev"}), beta)
+        gammas.append(gamma)
+
+    # forward-backward algorithm can be obtained by differentiating the backward algorithm
```
eb8680, line 158: I suppose this is true, but it seems like the point we'd like to make with this example is the same one made in reference [1], which is that the marginal probabilities `p(x[t] | Y)` or `p(x[t], x[t-1] | Y)` can be obtained directly by differentiating `Z_forward`, without ever constructing `beta[t]` ourselves - these are the quantities we actually care about, after all.

It would be easier to obtain these if you broke `trans` up into separate (log-)factors `p(x[t] | x[t-1])`, `p(y[t] | x[t])` in `main`.  Then you could write something like:
```py
with AdjointTape() as tape:
    log_Z = forward_algorithm(sum_op, prod_op, [trans, obs], time, {"x_prev": "x_curr"})  # don't return alphas

adjoint_log_marginals = tape.adjoint(sum_op, prod_op, log_Z, [trans, obs])
```
You could make `backward_algorithm` into `forward_backward_algorithm` that computes the same things by hand and compare the resulting marginals.

ordabayevy: I see. It makes more sense to me now. With the changes it will be much better example. Thanks!

examples/forward_backward.py
```diff
@@ -0,0 +1,194 @@
...
+        )
+    factors = [f.exp() for f in log_factors]
+
+    # Compute marginal probabilities using the forward-backward algorithm
+    marginals = forward_backward_algorithm(
+        ops.add, ops.mul, factors, {"x_prev": "x_curr"}
+    )
+    # Compute marginal probabilities using backpropagation
+    with AdjointTape() as tape:
+        log_Z = forward_algorithm(ops.add, ops.mul, factors, {"x_prev": "x_curr"})
```
ordabayevy, line 169: This works but I had to use `.log()` in `forward_algorithm` and register the adjoint of `LogOp`. Ideally I would like to run the forward algorithm with `log_factors` and use `ops.logaddexp` and `ops.add`. Then I think the derivative should work out as shown on line 130. However, at this moment it is not clear to me how `adjoint` module works with `ops.logaddexp` and `ops.add` and what are `adj_redop` and `adj_binop` are.

ordabayevy:
Blocked by #444

eb8680:
@ordabayevy can you fix merge conflicts and check the status of this example now that #445 has landed?

ordabayevy:
@eb8680 it is ready to be reviewed.

eb8680 (commented):
Since this is an example, I think in the interest of clarity it would be better to fix `sum_op` and `prod_op` to be `logaddexp` and `add` throughout and use infix syntax for `add` and `safesub`.  Looks good otherwise.

examples/forward_backward.py
```diff
@@ -0,0 +1,194 @@
...
+    [1] Jason Eisner (2016)
+        "Inside-Outside and Forward-Backward Algorithms Are Just Backprop
+        (Tutorial Paper)"
+        https://www.cs.jhu.edu/~jason/papers/eisner.spnlp16.pdf
+    [2] Zhifei Li and Jason Eisner (2009)
+        "First- and Second-Order Expectation Semirings
+        with Applications to Minimum-Risk Training on Translation Forests"
+        http://www.cs.jhu.edu/~zfli/pubs/semiring_translation_zhifei_emnlp09.pdf
+    """
+    funsor.set_backend("torch")
```
eb8680, line 144: Is this necessary?

ordabayevy: Good catch. Seems to work with numpy too.

eb8680 (approved):
Thanks for updating!

PR #430, opened 2021-01-19
Add Tuple funsor

eb8680:
pair-coded with @fritzo. Addresses #351.

### DIFF
--- a/funsor/domains.py
+++ b/funsor/domains.py
@@ -175,6 +175,33 @@ def bint(size):
     return Bint[size]
 
 
+class ProductDomain(Domain):
+
+    _type_cache = WeakValueDictionary()
+
+    def __getitem__(cls, arg_domains):
+        try:
+            return ProductDomain._type_cache[arg_domains]
+        except KeyError:
+            assert isinstance(arg_domains, tuple)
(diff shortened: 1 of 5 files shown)


funsor/domains.py
```diff
@@ -175,6 +175,37 @@ def bint(size):
...
+        return Product
+
+    @property
+    def shape(cls):
+        return (len(cls.__args__),)
+
+
+class Product(tuple, metaclass=ProductDomain):
+    """like typing.Tuple, but works with issubclass"""
+    __args__ = NotImplemented
```
fritzo, line 206: Does the `__args__` name have something to do with the `typing` module? I would expect Python-builtin-style naming to correspond to some Python concept, and to use `._args` otherwise.

fritzo (approved):
🎉

funsor/domains.py
eb8680: Yes, it's a convention I copied from `typing` whose purpose seems to be to signal that `__args__` are to be accessed through a `typing` introspection API (e.g. `typing_extensions.get_args(t)`) rather than directly, perhaps because the internals of `typing` change a lot between releases.  In a future PR we can add `get_origin()` and `get_args()` for `Domain`s.

PR #431, opened 2021-01-19
Add automatic registration of all unary and associative ops

fritzo:
Pair coded with @eb8680


fritzo:
@eb8680 thanks for tending the branches!

eb8680 (approved):

PR #432, opened 2021-01-19
Support IndependentConstraint in Distribution._infer_param_domain

eb8680:
Addresses #386. Blocking all other PRs.

This PR updates the logic in `Distribution._infer_param_domain` to handle the new `_IndependentConstraint` in NumPyro added in https://github.com/pyro-ppl/numpyro/pull/876, which should fix the CI failures in #430 and other PRs.

Tested:
- Exercised by existing tests that are currently failing, including `test_distributions_generic::test_generic_sample` for `CategoricalLogits`

### DIFF
--- a/funsor/distribution.py
+++ b/funsor/distribution.py
@@ -267,7 +267,7 @@ class Distribution(Funsor, metaclass=DistributionMeta):
         elif support_name == "Real":
             if name == "logits" and (
                     "probs" in cls.dist_class.arg_constraints
-                    and type(cls.dist_class.arg_constraints["probs"]).__name__ == "Simplex"):
+                    and type(cls.dist_class.arg_constraints["probs"]).__name__.lstrip("_") == "Simplex"):
                 output = Reals[raw_shape[-1 - event_dim:]]
             else:
                 output = Reals[raw_shape[len(raw_shape) - event_dim:]]
--- a/funsor/distribution.py
+++ b/funsor/distribution.py
@@ -249,23 +249,30 @@ class Distribution(Funsor, metaclass=DistributionMeta):
         # `infer_param_domain` methods.
         # Because NumPyro and Pyro have the same pattern, we use name check for simplicity.
         support_name = type(support).__name__
+
+        event_dim = 0
+        while support_name == "_IndependentConstraint":
+            event_dim += support.event_dim
+            support = support.base_constraint
+            support_name = type(support).__name__
+
         if support_name == "_Simplex":
-            output = Reals[raw_shape[-1]]
+            output = Reals[raw_shape[-1 - event_dim:]]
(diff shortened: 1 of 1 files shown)


fehiepsi (approved):
Thanks for the fix, Eli!

@fritzo I guess we can add
```
if isinstance(base_constraint, _IndependentConstraint):
    batch_ndims = base_constraint.batch_ndims + batch_ndims
    base_constraint = base_constraint.base_constraint
```
to simplify the `while` loop here. What do you think?

funsor/distribution.py
```diff
@@ -249,23 +249,30 @@ def _infer_param_domain(cls, name, raw_shape):
         # `infer_param_domain` methods.
         # Because NumPyro and Pyro have the same pattern, we use name check for simplicity.
         support_name = type(support).__name__
+
+        event_dim = 0
+        while support_name == "_IndependentConstraint":
+            event_dim += support.event_dim
```
fritzo, line 255: For now this should be corrected to
```py
event_dim += support.reinterpreted_batch_ndims
```
In the future (after https://github.com/pytorch/pytorch/pull/50547) we should eliminate this computation and simply define at the top
```py
event_dim = constraint.event_dim
```

fritzo (changes requested):

fritzo:
Note the constraint names will probably be made public in the future so we should do something like
```py
support_name = type(support).__name__.lstrip("_")
```
and reference names like `IndependentConstraint` and `Real`, so as to be forwards-compatible. We can do that in a subsequent PR or in this PR.

funsor/distribution.py
```diff
@@ -249,23 +249,30 @@ def _infer_param_domain(cls, name, raw_shape):
         # `infer_param_domain` methods.
         # Because NumPyro and Pyro have the same pattern, we use name check for simplicity.
         support_name = type(support).__name__
+
+        event_dim = 0
+        while support_name == "_IndependentConstraint":
```
fritzo, line 254: Can you change this to 
```py
support_name = type(support).__name__.lstrip("_")

event_dim = 0
while support_name == "IndependentConstraint":
    ...
```
so as to be compatible with all backends? Note in Pyro the class is named `IndependentConstraint`.

eb8680: We could also duplicate `_infer_param_domain` in each backend so that dispatch happens on constraint types rather than names.

fehiepsi: > and in a future PyTorch the class will be renamed to IndependentConstraint

Does it apply to other constraints as well?

fritzo: yes, see plans in https://github.com/pytorch/pytorch/issues/50616

fritzo (approved):

PR #433, opened 2021-01-20
Extract subtyping metaclass from FunsorMeta and domains

eb8680:
Addresses #351, #355.

This PR that attempts to isolate and reuse the parametric subtyping and type-caching functionality currently present in the metaclasses `FunsorMeta`, `ArrayType` and #430's `ProductDomain`, and to simplify their implementations.  This change should enable nested `Product`s and simplify the process of adding new `Product`/`tuple`-like `Domain`s (i.e. with the same variance as `Product`).

More broadly, this PR represents a minimal first attempt to bring both interpreter types and funsor types closer to the Python `typing` type system, thereby saving us work, reducing cognitive overhead for users who are familiar with `typing`, and  approaching compatibility with some of the wider tooling ecosystem built around `typing`.  It may turn out that these potential advantages are outweighed by design choices in `typing` or `multipledispatch`, or that bugs, edge cases or performance limit the feasibility of this direction in practice, but it's hard to know without trying.

I will also put up a different version of these changes that integrates with `typing` more directly in a separate PR.

Tested:
- Exercised by all existing tests

### DIFF
--- a/funsor/util.py
+++ b/funsor/util.py
@@ -4,8 +4,12 @@
 import functools
 import inspect
 import re
+import typing
+import weakref
 
 import numpy as np
+from multipledispatch.variadic import isvariadic
+
 
 _FUNSOR_BACKEND = "numpy"
 _JAX_LOADED = False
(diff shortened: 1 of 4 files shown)


funsor/util.py
```diff
@@ -230,3 +234,91 @@ def decorator(fn):
...
+    """
+    def __init__(cls, name, bases, dct):
+        super().__init__(name, bases, dct)
+        if not hasattr(cls, "__args__"):
+            cls.__args__ = ()
+        if cls.__args__:
+            base, = bases
+            cls.__origin__ = base
+        else:
+            cls._type_cache = weakref.WeakValueDictionary()
```
fritzo, line 251: could you add some comments here explaining when `._type_cache` is used, etc.

fritzo (commented):
Seems like a good idea; I'm interested in hearing details tomorrow.

eb8680:
Closing as obsolete.

PR #437, opened 2021-01-23
Add a sketch of a lazy Scatter term and some tests

eb8680:
Pair coded with @fritzo @ordabayevy @fehiepsi 

This is blocking the extension backward sample #438 to handle `MarkovProduct`, which is used in NumPyro.


eb8680:
Closing in favor of #445

PR #440, opened 2021-01-24
Prettify generic distribution tests

fritzo:
These are entirely aesthetic changes that make the tests more readable, auditable, and editable:
- wrap long lines
- replace `backend_dist` -> `dist` in test strings
- move `eval()` into a single helper
- avoid unnecessary destructured assignment
- patch to backport https://github.com/pyro-ppl/pyro/pull/2748


eb8680 (approved):
LGTM, thanks for cleaning these up!

PR #442, opened 2021-01-26
Add a @make_funsor factory for Funsor subclasses

fritzo:
pair coded with @eb8680 

Added a Funsor subclass factory `@make_funsor` with [dex-style notation](https://github.com/google-research/dex-lang/blob/main/examples/tutorial.dx) based on type hints.

## Tested
- [x] a few unit tests


eb8680 (approved):
🎉

PR #443, opened 2021-01-26
Directly convert IndependentDistribution to base distribution

eb8680:
Addresses #386. Should unblock #427.

This PR modifies `to_funsor` conversion of `IndependentDistribution`s to directly modify parameter inputs and outputs via `funsor.Lambda` rather than representing them with `funsor.terms.Independent`, which was breaking some pattern matching in #427.

Tested:
- Exercised by existing tests in `test_distribution_generic.py` that use `Independent`, and more directly in #427 
- Merged into #427 locally and confirmed that tests in #427 pass

### DIFF
--- a/funsor/distribution.py
+++ b/funsor/distribution.py
@@ -19,9 +19,9 @@ from funsor.cnf import Contraction, GaussianMixture
 from funsor.domains import Array, Real, Reals
 from funsor.gaussian import Gaussian
 from funsor.interpreter import gensym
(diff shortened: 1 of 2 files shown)


fritzo (approved):
LGTM once tests pass.

In the longer term will we deprecate `Independent` in favor of `Lambda`?

eb8680:
> In the longer term will we deprecate Independent in favor of Lambda?

I'm not sure, I guess it depends on how much it ends up being used in places that are closer to user code.

PR #447, opened 2021-01-31
Improve __str__; add ops //, %, <<, >>

fritzo:
## Summary
- Improves `Funsor.__repr__()`, `.__str__()` for quick debugging (for detailed structure, see `.pretty()`)
- Adds `ops.mod`, `ops.floordiv`, fixing the `Flatten21` example in test_factory.py
- Adds `ops.lshift`, `ops.rshift` (which Martin and I actually used just yesterday)
- Fixes a missing `find_domain()` computation in `Binary(Op, Number, Tensor)` and `Binary(Op, Tensor, Number)`

## Tested
- [x] un-xfailed the `Flatten21` test
- [x] added new ops to tensor tests

### DIFF
--- a/funsor/cnf.py
+++ b/funsor/cnf.py
@@ -82,17 +82,21 @@ class Contraction(Funsor):
         self.reduced_vars = reduced_vars
 
     def __repr__(self):
-        if self.op in _INFIX:
+        if self.bin_op in _INFIX:
+            bin_op = " " + _INFIX[self.bin_op] + " "
             return "{}.reduce({}, {})".format(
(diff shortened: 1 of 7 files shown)


eb8680 (approved):
Nice, are we missing support for any other builtin operators?

fritzo:
The only remaining [builtin ops](https://docs.python.org/3/library/operator.html) are non-numeric things like `x is y`, `x is not y` and `bool(x)`, and I believe we already have correct behavior for those.

It still remains to fix/define `find_domain()` for `<<`, `>>`, and a couple others, but I think we'll need to first resolve #448.

PR #450, opened 2021-02-02
Reinstate black-compatible isort

fritzo:
- Adds `isort .` to `make format`
- Adds `isort --check .` to `make lint`
- Adds isort section to setup.cfg
- Adds `isort>=5.0` to requirements
- Works around https://github.com/pytorch/pytorch/issues/47038

### DIFF
--- a/funsor/torch/__init__.py
+++ b/funsor/torch/__init__.py
@@ -4,18 +4,19 @@
 import torch
 from multipledispatch import dispatch
 
-import funsor.ops as ops
-import funsor.torch.distributions  # noqa: F401
-import funsor.torch.ops  # noqa: F401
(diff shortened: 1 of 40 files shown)


funsor/torch/__init__.py
```diff
@@ -4,13 +4,13 @@
 import torch
 from multipledispatch import dispatch
 
+import funsor.ops as ops
```
eb8680, line 7: I believe you can resolve the Travis error by adding a line below like `from funsor.ops import AssociativeOp` and changing `ops.AssociativeOp` to `AssociativeOp`.

eb8680 (approved):

PR #452, opened 2021-02-02
Move instrumentation into funsor.instrument module

fritzo:
This moves a bunch of instrumentation from `funsor.interpretations` to `funsor.instrument`, separating concerns.


eb8680 (approved):
LGTM

PR #453, opened 2021-02-03
Add Travis stages with Python 3.7-3.9

eb8680:


eb8680:
I've folded this into #451, since the changes there fix the test failure seen here for Python 3.9.

PR #456, opened 2021-02-04
Add helper to rewrite infix ops as local variables

fritzo:
This adds a helper to write infix ops as parametrized ops, which may improve readability functions parametrized by op or semiring. This crucially differs from reinterpretation because the nonstandard operator meanings are localized to the decorated function's source code; no subroutines are reinterpreted. This requires Python 3.9: it should not yet be used in main library code, but it will help with prototyping.

 For example the the following code
```py
@rewrite_ops({"+": "sum_op", "*": "prod_op"})
def product_rule(sum_op, prod_op, lhs, rhs, d):
    return d(lhs) * rhs + lhs * d(rhs)
```
will be rewritten as
```py
def product_rule(sum_op, prod_op, lhs, rhs, d):
    return sum_op(prod_op(d(lhs), rhs), prod_op(lhs, d(rhs)))
```

## Tested
- [x] unit tests (requires Python 3.9)


eb8680 (approved):
LGTM

Issue #411, opened 2020-12-22
Replace normalize interpretation with rewriting modulo a congruence


fritzo:
I would have more trust in Funsor's rewriting system if we (1) moved some complexity from the `normalize` interpretation to the pattern matching mechanism, and (2) simplified and codified the allowable control flow patterns whereby one interpretation can defer to / nest within another. This issue addresses (1).

Originally Funsor interpretations were naive and resulted in small reduction sequences, with depth linear in term length (I believe). Funsor's simple `eager` and `lazy` interpretations were dwarfed in complexity by the `multipledispatch` library on which they were built. But interpretations have became more complex, especially with the usage patterns in `normalize` and with interpretations calling sub-interpretations in increasingly nested ways. As we build more machinery on top of Funsor, pattern matching errors will increasingly surface to users, and increasingly often those users will be unaware that Funsor is being used at all.

This issue proposes to move some complexity of `normalize` into a smarter replacement for `multipledispatch`, as a separate library, possibly using `multipledispatch` or Matt's [unification](https://github.com/mrocklin/unification) library or even an SMT solver like Z3. I believe this would simplify the writing of new patterns, would allow for clearer error handling in case of missing patterns (which users will encounter often!), and will more precise debugging information for authors of new patterns, and hopefully will reduce complexity of large normalizations (since we can implement custom matching algorithms).

## Design

I believe the congruence should include laws for:
- associativity
- commutativity
- unit laws
- possibly distributivity
- possibly idempotence for some ops (e.g. max)

Initially these can be hard-coded; then we can refactor so the equational theory itself is declared in Funsor rather than in the rewriting library.

## Possibly relevant references
- Batchmair & Dershowitz (1989) "Completion for rewriting modulo a congruence" ([pdf](https://core.ac.uk/download/pdf/82630592.pdf))
- Viry (1995) "Rewriting modulo a rewrite system" ([pdf](https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.49.5233&rep=rep1&type=pdf))
- Rocha et al. (2014) "Rewriting modulo SMT and open systems analysis" ([pdf](https://core.ac.uk/reader/42708020))

PR #463, opened 2021-02-20
Add an Elbo interpretation implementing Jensen's inequality

fritzo:
pair coded with @eb8680 @fehiepsi @ordabayevy


fritzo:
@eb8680 Can we merge this with an xfailing test, since the pattern failure seems to lie elsewhere and will require significant work?

eb8680:
Yeah, no need to block this on #411

eb8680 (approved):

PR #466, opened 2021-02-20
Update readthedocs page

fehiepsi:


setup.py
```diff
@@ -39,15 +37,8 @@
...
     extras_require={
-        "torch": [
-            "pyro-ppl>=1.5.2",
-            "torch>=1.7.0",
-        ],
-        "jax": [
-            "numpyro>=0.2.4",
-            "jax>=0.1.57",
-            "jaxlib>=0.1.37",
-        ],
```
fehiepsi, line 50: @fritzo Do you have a config file for black? I run `make format` but it causes a lot of changes.

fritzo: Our only config file is [setup.cfg](https://github.com/pyro-ppl/funsor/blob/master/setup.cfg) but we do not configure black. Maybe you need to update isort, whose black support was added only in 5.0?

fehiepsi: I just updated the packages. I am not sure what is the default behavior but it seems to me the changes here look more consistent with what I think `black` will do. For example,
```
        params = {
            "eps_g": {},
            "eps_i": {},
        }
```
should be in 1 line
```
params = {"eps_g": {}, "eps_i": {}}
```
if it fits into an 88-length (black default line length) line

fritzo: Hmm that's weird. I especially don't understand why linting (`black --check .`) passed both before and after your black changes https://github.com/pyro-ppl/funsor/pull/466/commits/e7dc9d1db7637ed29c9ea8341a987360be607ccd Does `make lint` pass on your system?

fehiepsi: Yeah, you are right. `black --check .` passed now, and `make format` does not change files anymore. It is weird... I guess black supports two formatting rules for parentheses and based on "cache" or something like that, black will behave differently... Anyway, I will revert unrelated changes because `make format` seems to work now in my system.

fehiepsi:
@fritzo There are a few black changes remaining, hope that they are fine. I have incorporated (and tested) the following changes that I made to rtd page:
+ add the ability to add examples/tutorials to `index.rst` file
+ add the ability to generate thumbnail galleries (this is not necessary)
+ add colab links to rendered tutorials
+ each tutorial/example has a link that points to the source file (with correct version) in github.

I need to add `__version__` to `funsor.__init__.py` to generate correct links for local versions (not sure if it is needed for `funsor.pyro.ai`). We expose `__version__` in pyro and numpyro so it is fine to do it here I guess.

fritzo (approved):

fritzo:
Thanks @fehiepsi, this is great!

PR #468, opened 2021-02-23
Reduce stack depth of PrioritizedInterpretation; add docs

fritzo:
## Summary

- Reduces stack depth of `ProiritizedInterpretation`, as pair coded with @eb8680 in #463 
- Moves `MemoizeInterpretation` into interpretations.py (since it is a basic combinator and has no special patterns)
- Adds interpretations docstrings
- Fixes docs errors in funsor.ops and funsor.typing

## Tested
- built docs locally


funsor/interpretations.py
```diff
@@ -203,6 +256,46 @@ def register(cls, *args):
...
+        return value
+
+
+@contextmanager
+def memoize(cache=None):
+    """
+    Context manager wrapping :class:`Memoize` and yielding the ``cache`` dict.
+    """
+    base_interpretation = get_interpretation()
+    with Memoize(base_interpretation, cache) as interp:
```
eb8680, line 295: This seems like a reoccurring pattern (also in `substitute`), should we abstract it into an `Interpretation`?

eb8680 (approved):
LGTM

funsor/interpretations.py
fritzo: I think we can wait for a couple more examples. The only reason this isn't directly in `Memoize` is because `Memoize.__enter__()` returns self whereas this needs to return the cache dict.

PR #478, opened 2021-03-04
Bump Pyro and PyTorch dependencies and reinstate VAE

### LINKED ISSUE #474: Restore VAE example smoke test to CI
In #473 we disabled the `examples/vae.py` smoke test due to a bug in `torchvision`.  This issue is a reminder to reenable the test again when releases of PyTorch and `torchvision` that include a fix become available.

eb8680:
Resolves #474

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,7 +51,7 @@ jobs:
         python -m pip install --upgrade pip
         # Keep track of pyro-api master branch
         pip install https://github.com/pyro-ppl/pyro-api/archive/master.zip
-        pip install torch==1.8.0+cpu torchvision==0.9.0+cpu -f https://download.pytorch.org/whl/test/cpu/torch_test.html
+        pip install torch==1.8.0+cpu torchvision==0.9.0+cpu -f https://download.pytorch.org/whl/torch_stable.html
         # Keep track of Pyro master branch
         pip install https://github.com/pyro-ppl/pyro/archive/master.zip
         pip install .[test,torch]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,7 +51,7 @@ jobs:
         python -m pip install --upgrade pip
         # Keep track of pyro-api master branch
         pip install https://github.com/pyro-ppl/pyro-api/archive/master.zip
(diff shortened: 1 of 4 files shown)


fritzo (approved):

fritzo:
Thanks for updating this @eb8680 !

fehiepsi (approved):

.github/workflows/ci.yml
```diff
@@ -51,7 +51,7 @@ jobs:
         python -m pip install --upgrade pip
         # Keep track of pyro-api master branch
         pip install https://github.com/pyro-ppl/pyro-api/archive/master.zip
-        pip install torch==1.7.0+cpu torchvision==0.8.1+cpu -f https://download.pytorch.org/whl/torch_stable.html
+        pip install torch==1.8.0+cpu torchvision==0.9.0+cpu -f https://download.pytorch.org/whl/test/cpu/torch_test.html
```
fritzo, line 54: these should be stable builds

fritzo (changes requested):
There have been some nontrivial distribution shape changes in PyTorch 1.8 We may need to fix some of the distribution shape conversion.

.github/workflows/ci.yml
eb8680: Tests keep failing with stable builds.

fritzo:
It looks like you might need to `pip install requests`?

eb8680:
> It looks like you might need to pip install requests?

But should we be doing that manually or should it be installed as a dependency of `torchvision`?

eb8680:
> We may need to fix some of the distribution shape conversion

The tests in `test_distribution.py` and `test_distribution_generic.py` seem to be passing - are those tests too weak to catch these changes?

.github/workflows/ci.yml
eb8680: Reverted to stable

fritzo (approved):
LGTM, thanks for figuring this out

fritzo:
> should we be [installing requests] manually

I think you probably did the right thing in adding `requests` as a test dependency. In an ideal world we could rely on packages declaring all their dependencies, but honestly the dark corners of Pyro have a few undeclared dependencies (biopython, lap, gpytorch), so I can't criticize.

> are [our existing distribution] tests too weak to catch these changes?

hmm, I guess if they pass, then everything is ok 🤷

PR #482, opened 2021-03-06
Add ReductionOp, ops.mean, ops.std, ops.var

ordabayevy:


fritzo:
Could you register implementations in funsor/jax/ops.py the following test to test/test_terms.py:

<details>

```diff
diff --git a/test/test_terms.py b/test/test_terms.py
index e495e3e..88ed47a 100644
--- a/test/test_terms.py
+++ b/test/test_terms.py
@@ -287,6 +287,29 @@ def test_unary(symbol, data):
     check_funsor(actual, {}, Array[dtype, ()], expected_data)


+@pytest.mark.parametrize("event_shape", [(4,), (3, 2)], ids=str)
+@pytest.mark.parametrize(
+    "name",
+    [
+        "all",
+        "any",
+        "logsumexp",
+        "max",
+        "mean",
+        "min",
+        "prod",
+        "std",
+        "sum",
+        "var",
+    ],
+)
+def test_reduce_event(name, event_shape):
+    dtype = 2 if name in ("any", "all") else "real"
+    x = random_tensor(OrderedDict(i=Bint[5]), output=Array[dtype, event_shape])
+    actual = getattr(x, name)()
+    check_funsor(actual, x.inputs, Array[dtype, ()])
+
+
 BINARY_OPS = [
     "+",
     "-",
```

</details>

ordabayevy:
Is there a need to allow `ddof` option for `ops.std` and `ops.var`? If yes, how is it passed to `Unary`? Similar to `shape` option in `.reshape`?

```py
    def reshape(self, shape):
        return Unary(ops.ReshapeOp(shape), self)
```

fritzo:
> Is there a need to allow `ddof` option

I'm not sure what the `ddof` option is, but it would be reasonable to create, in addition to these default `mean`,`var`,`std` ops, more flexible op classes `MeanOp`, `VarOp`, and `StdOp` to take non-default arguments `(axis,keepdims)` (in numpy parlance) or `(dim,keepdim)` (in pytorch parlance). You're correct that these could follow `ReshapeOp` or `GetitemOp` by using `CachedOpMeta`. However I think those ops would best be separate and complementary to the default ops, and we would also want to provide them for `sum` etc.:
```py
class Funsor(...):
    ...
    def mean(self, axis=None, keepdims=False):
        if axis is None and keepdims is False:
            op = ops.mean  # default version from this PR
        else:
            op = ops.MeanOp(axis, keepdims)  # fancy version from a future PR
        return Unary(op, self)
```

funsor/jax/ops.py
```diff
@@ -147,6 +147,11 @@ def _log(x):
     return np.log(x)
 
 
+@ops.mean.register(array, (int, type(None)))
+def _mean(x, dim):
```
fritzo, line 151: Note there's no way to specify `dim` or `ddof` if you use the simple op factory `mean = make_op(np.mean)`. If you want fancier behavior you'd need to define a
```py
class MeanOpMeta(CachedOpMeta):
    ...
class MeanOp(Op, metaclass=MeanOpMeta):
    def __init__(self, axis=None, keepdims=False):
        ...
    ...
mean = MeanOp()  # default behavior
```

funsor/tensor.py
```diff
@@ -772,23 +767,27 @@ def eager_reshape_tensor(op, arg):
     return Tensor(data, arg.inputs, arg.dtype)
 
 
-@eager.register(Unary, ops.SumOp, Tensor)
-def eager_sum_tensor(op, arg):
+@eager.register(Unary, ops.ReductionOp, Tensor)
+def eager_reduction_tensor(op, arg):
+    if not arg.output.shape:
+        return arg
```
fritzo, line 773: This won't work for `argmin`,`argmax`.  Are those two really `ReductionOp`s?  Maybe we should call them `ArgreductionOp`s or sth?

test/test_tensor.py
```diff
@@ -1317,23 +1317,23 @@ def test_sum(event_shape):
     for dim in DIMS:
         assert_close(Tensor(ops.sum(data, dim)), ops.sum(Tensor(data), dim))
         assert_close(Tensor(ops.sum(data, dim=dim)), ops.sum(Tensor(data), dim=dim))
-    for keepdims in KEEPDIMS:
+    for keepdim in KEEPDIMS:
```
fritzo, line 1320: Could you copy in my stronger tests from #503, and I'll close that PR?

fritzo (commented):
Looks great!  I'm still uncertain of `argmin`, `argmax`...

test/test_tensor.py
```diff
@@ -1345,14 +1345,14 @@ def test_sum_batch(batch_shape, event_shape):
     DIMS = [None, 0, 1, 2, -1, -2, -3, (0, 2)]
     KEEPDIMS = [False, True]
 
-    def raw_sum(x, dim=None, keepdims=False, batch_ndims=len(batch_shape)):
+    def raw_sum(x, dim=None, keepdim=False, batch_ndims=len(batch_shape)):
```
fritzo, line 1348: Can you parameterize this and the above test on op?
```py
@pytest.mark.parametrize(
    "op",
    [ops.any, ops.all, ops.amin, ops.amax, ops.sum, ops.prod, ops.mean, ops.std, ops.var],
)
```

ordabayevy:
> I'm still uncertain of `argmin`, `argmax`...

There were some rough edges and `argmin` and `argmax` are definitely one of them. Probably should be in a separate `ArgreductionOp` as you suggested.

fritzo (approved):
🎉

PR #483, opened 2021-03-06
Clean up examples and tests after black

fritzo:
This cleans up after `black` expanded some code in examples in tests:
- Replaces `OrderedDict([("i", Real)])` -> `OrderedDict(i=Real)` as allowed in Python 3.6
- Leverages `Funsor.reduce()` convenience handling to simplify example `reduced_vars`

### DIFF
--- a/test/test_tensor.py
+++ b/test/test_tensor.py
@@ -129,15 +129,7 @@ def test_indexing():
 
 def test_advanced_indexing_shape():
     I, J, M, N = 4, 4, 2, 3
-    x = Tensor(
-        randn((I, J)),
-        OrderedDict(
-            [
-                ("i", Bint[I]),
-                ("j", Bint[J]),
-            ]
-        ),
-    )
+    x = Tensor(randn((I, J)), OrderedDict(i=Bint[I], j=Bint[J]))
     m = Tensor(numeric_array([2, 3]), OrderedDict([("m", Bint[M])]), I)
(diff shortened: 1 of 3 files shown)


test/test_tensor.py
```diff
@@ -1030,41 +960,13 @@ def test_tensor_stack(n, shape, dim):
...
-            [
-                ("i", Bint[2]),
-                ("k", Bint[4]),
-            ]
-        ),
-        output,
-    )
+    x = random_tensor(OrderedDict(i=Bint[2]), output)
+    y = random_tensor(OrderedDict(j=Bint[3]), output)
+    z = random_tensor(OrderedDict(i=Bint[2], k=Bint[4]), output)
```
fehiepsi, line 965: I like this one-line pattern when the expression is short. But when we add a comma `,` before the closing bracket `]` of a list (or a tuple), somehow black will expand it into multiple lines. A solution is to remove those extra `comma` for short expressions. What do you think? I can make a quick PR for it (using a hack: running `black` master with `--line-length 1000` and  `--skip-magic-trailing-comma` option to remove the final comma, then run `black` again with default options) so you can see what's the difference.

fehiepsi (approved):

test/test_tensor.py
fehiepsi: Nevermind, the trick I mentioned above works well for blocks without "comments". For those with comments, the comments are moved to unexpected positions... ==

PR #484, opened 2021-03-06
Unexpand short expressions with black

fehiepsi:
This PR is just used to illustrate the difference between two behaviors that black allows. Pros vs cons of this change

### pros:
+ many short expressions are wrapped into 1 line, hence we have `249 additions and 1,108 deletions` (half of them belong to `test/test_sum_product.py`)

### cons:
+ some comments are moved to unexpected positions (hence causes some lint errors), so we need to fix them by hand (this is not a big problem)
+ some multi-line short expressions are intended, so we need to fix them by hand by adding `,` again before the closing brackets (this can be a big problem because the pattern depends on personal preference and this change kills that preference)


examples/eeg_slds.py
```diff
@@ -47,15 +47,15 @@ def download_data():
...
+        num_components,
+        hidden_dim,
+        obs_dim,
+        fine_transition_matrix=True,
+        fine_transition_noise=False,
+        fine_observation_matrix=False,
+        fine_observation_noise=False,
         moment_matching_lag=1,
-    ):  # controls the expense of the moment matching approximation
+    ):  # the number of switching states K  # the dimension of the continuous latent space  # the dimension of the continuous outputs  # controls whether the transition matrix depends on s_t  # controls whether the transition noise depends on s_t  # controls whether the observation matrix depends on s_t  # controls whether the observation noise depends on s_t  # controls the expense of the moment matching approximation
```
fritzo, line 58: Move to arg comments to a new docstring

funsor/minipyro.py
```diff
@@ -286,10 +286,10 @@ def sample(name, fn, obs=None, infer=None):
         "fn": fn,
         "args": (),
         "value": obs,
-        "cond_indep_stack": {},  # maps dim to CondIndepStackFrame
+        "cond_indep_stack": {},
         "output": fn.output,
         "infer": {} if infer is None else infer,
-    }
+    }  # maps dim to CondIndepStackFrame
```
fritzo, line 292: fix comment

funsor/minipyro.py
```diff
@@ -347,9 +347,9 @@ def fn(init_value, constraint):
         "fn": fn,
         "args": (init_value, constraint),
         "value": None,
-        "cond_indep_stack": cond_indep_stack,  # maps dim to CondIndepStackFrame
+        "cond_indep_stack": cond_indep_stack,
         "output": output,
-    }
+    }  # maps dim to CondIndepStackFrame
```
fritzo, line 352: fix comment

test/test_factory.py
```diff
@@ -19,9 +19,7 @@
 def test_lambda_lambda():
     @make_funsor
     def LambdaLambda(
-        i: Bound,
-        j: Bound,
-        x: Funsor,
+        i: Bound, j: Bound, x: Funsor
```
fritzo, line 22: i find it much easier to read `@make_funsor` functions with vertical argument lists. can we revert changes to this file?

funsor/distribution.py
```diff
@@ -160,9 +160,11 @@ def _get_raw_dist(self):
         """
         value_name = [
             name
-            for name, domain in self.value.inputs.items()  # TODO is this right?
+            for name, domain in self.value.inputs.items()
             if domain == self.value.output
-        ][0]
+        ][
+            0
+        ]  # TODO is this right?
```
fritzo, line 167: rewrite or move the TODO to preserve meaning

fritzo (changes requested):
Can you explain in the PR description how you triggered these changes? Was it twiddling commas, or did you change a config file?

Most of these changes look good. I'd prefer vertical lists in a few places if possible, e.g. `__all__`, `@make_funsor` functions, and global config dicts.

fehiepsi:
@fritzo I used the following combos
```
pip install git+https://github.com/psf/black.git
black -l 1000 -C .
black .
```
I have reverted the behaviors at the comment positions, `__all__`, some global dict, and `make_funsor`. The remaining changes look clean to me. Thanks!

fritzo (approved):
Thanks for cleaning this up!

Issue #446, opened 2021-01-29
Redesign adjoint / transpose interface and algorithms


fritzo:
## Reading list

- https://popl21.sigplan.org/details/lafi-2021-papers/9/Decomposing-reverse-mode-automatic-differentiation
- [The Simple Essence of Automatic Differentiation](https://arxiv.org/pdf/1804.00746.pdf)

ordabayevy:
Just throwing out an idea here. Another perspective on two different semi-rings (ops.add, ops.mul) and (ops.logaddexp, ops.add) is to view the latter as operations lifted by a `Log` Functor. In Haskell notation:

```hs
instance Functor Log where
    fmap f :: Log a -> Log b
    fmap f (Log a) = Log (f a)

...

fmap (*) (Log (x, y)) = Log (x*y) = Log x + Log y
fmap (+)  = logaddexp
```

These two different types could be represented by something like `Factor` and `Logfactor` funsors which can be used for pattern matching.


As a side note, can Funsor be used to represent partially applied functions?

fritzo:
@ordabayevy yes I think your `Log` functor relates the semiring pair `((add,mul), (logaddexp,add))` and the pair `((max,mul),(max,add))`. However no functor relates e.g. the pair `((add,mul), (max,add))`.

Indeed `Funsor`s can represent functions, and funsor substitution represents partially applied functions.

PR #488, opened 2021-03-11
Use Approximate in adjoint

eb8680:
pair coded with @fritzo 

Addresses #446

### DIFF
--- a/funsor/jax/ops.py
+++ b/funsor/jax/ops.py
@@ -64,7 +64,7 @@ def _any(x, dim):
 
 @ops.astype.register(array, str)
 def _astype(x, dtype):
-    return x.astype(dtype)
+    return x.astype(np.result_type(dtype))
 
 
 @ops.cat.register(int, [array])
--- a/funsor/testing.py
+++ b/funsor/testing.py
@@ -89,25 +89,24 @@ def assert_close(actual, expected, atol=1e-6, rtol=1e-6):
     if is_array(actual):
         assert is_array(expected), msg
     elif isinstance(actual, Tensor) and is_array(actual.data):
-        assert isinstance(expected, Tensor) and is_array(expected.data)
+        assert isinstance(expected, Tensor) and is_array(expected.data), msg
     elif (
         isinstance(actual, Contraction)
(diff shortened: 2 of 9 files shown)


fritzo (approved):

fritzo:
I think we can fix tests by simply
```diff
- data1 = ops.astype(data1, "int64")
+ data1 = ops.astype(data1, "int32" if get_backend() == "jax" else "int64")
```

fehiepsi (approved):

funsor/jax/ops.py
```diff
@@ -64,7 +64,7 @@ def _any(x, dim):
 
 @ops.astype.register(array, str)
 def _astype(x, dtype):
-    return x.astype(dtype)
+    return x.astype(np.result_type(dtype))
```
fehiepsi, line 67: `result_type` can be used to convert x64 to x32 dtypes. In JAX, using `x.astype("int64")` in x32 mode will raise a warning.

funsor/testing.py
```diff
@@ -96,6 +96,10 @@ def assert_close(actual, expected, atol=1e-6, rtol=1e-6):
         and is_array(actual.terms[0].data)
     ):
         assert isinstance(expected, Contraction) and is_array(expected.terms[0].data)
+    elif isinstance(actual, Contraction) and isinstance(actual.terms[0], Delta):
+        assert isinstance(expected, Contraction) and isinstance(
+            expected.terms[0], Delta
+        )
```
fehiepsi, line 102: I'm not sure how to resolve this issue, where input `numpy.ndarray` is compared to `DeviceArray` but we can't assert they have the same type. This is not important I guess.

fritzo: Hmm, I think we should add some sort of check that values agree here, otherwise we'll have no confidence in the tests, right?

also nit: all assertions here should end with `, msg`

fritzo: Ah, I see the values are checked later in this function. LGTM.

fritzo (commented):
LGTM, @fehiepsi thanks for fixing the JAX issue. (I've pushed a couple nits)

Issue #489, opened 2021-03-13
Support for non-funsor arguments to ops


fritzo:
This issue tracks changes needed to support non-funsor args as in @ordabayevy's #482, e.g.
```py
ops.sum(Tensor(data), -1, keepdims=True)
# -op-dispatch-> Unary(ops.SumOp(-1, True), Tensor(data))
# -eager-> Tensor(ops.SumOp(-1,True)(data))  # TODO fix batch dimension handling
```

## Plan
Note during the first two refactorings, we may want to try to decouple dispatch tables from the `Op` hierarchy (but that would be a bigger refactoring, possibly beyond the scope of this issue).
- [x] #491 Refactor `Op` so that the dispatch table is in `Op` subclasses, not in instances.
- [x] #490 Demonstrate parametrized ops via `ops.sum(data, dim=None, keepdims=False)`
- [x] #482 Fix batch dimension handling in `Tensor.eager_unary()`
- [x] #482 add more ops, e.g. a richer `ops.sum`, `ops.mean`, etc.

eb8680:
Can this be closed now that #482 is merged?

fritzo:
Yes, thanks for your help @ordabayevy !

PR #490, opened 2021-03-13
Support ops.sum(data, dim=None, keepdims=False)

fritzo:
Addresses #489 
pair coded with @eb8680 @ordabayevy @fehiepsi 

This demonstrates the new parametrized op syntax from #491 .  The recipe is:
1. add new `*args, *kwargs` to your op in `funsor.ops.array`
2. add backend support in `funsor.torch.ops` and `funsor.jax.ops`
3. implement `find_domain(op, ...)` in `funsor.domains`
4. work around batch dimensions in `funsor.tensor`
5. add some tests

## Tested
- [x] unit test for various op arg,kwarg syntax
- [x] unit test for batched tensors

### DIFF
--- a/funsor/tensor.py
+++ b/funsor/tensor.py
@@ -773,7 +773,7 @@ def eager_reshape_tensor(op, arg):
 
 
 @eager.register(Unary, ops.SumOp, Tensor)
-def eager_reshape_tensor(op, arg):
+def eager_sum_tensor(op, arg):
     if not arg.inputs:
         return Tensor(op(arg.data), arg.inputs, arg.dtype)
 
--- a/funsor/tensor.py
+++ b/funsor/tensor.py
@@ -772,6 +772,25 @@ def eager_reshape_tensor(op, arg):
     return Tensor(data, arg.inputs, arg.dtype)
 
 
+@eager.register(Unary, ops.SumOp, Tensor)
+def eager_reshape_tensor(op, arg):
+    if not arg.inputs:
+        return Tensor(op(arg.data), arg.inputs, arg.dtype)
+
+    # Work around batch inputs.
+    dim = op.defaults.get("dim", None)
+    keepdims = op.defaults.get("keepdims", False)
+    ndims = len(arg.output.shape)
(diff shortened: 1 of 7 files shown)


fritzo:
@ordabayevy I hope this can serve as a template for your #482 .  Feel free to refactor after this PR merges, in case you'd like to reuse the logic e.g. in `find_domain` or `eager.register`. We may even be able to create a subclass `ReductionOp(UnaryOp)` and register `ops.sum`, `ops.mean`, etc. in a single pattern.

funsor/tensor.py
```diff
@@ -772,6 +772,25 @@ def eager_reshape_tensor(op, arg):
     return Tensor(data, arg.inputs, arg.dtype)
 
 
+@eager.register(Unary, ops.SumOp, Tensor)
+def eager_reshape_tensor(op, arg):
+    if not arg.inputs:
+        return Tensor(op(arg.data), arg.inputs, arg.dtype)
+
+    # Work around batch inputs.
```
eb8680, line 780: This logic converting dimensions into batch-aware dimensions seems useful and general-purpose enough that some version of it should maybe live in an `Op` method or something - we don't want to have to write this from scratch in each new op with nontrivial shape semantics.

funsor/tensor.py
```diff
@@ -772,6 +772,25 @@ def eager_reshape_tensor(op, arg):
     return Tensor(data, arg.inputs, arg.dtype)
 
 
+@eager.register(Unary, ops.SumOp, Tensor)
+def eager_reshape_tensor(op, arg):
```
eb8680, line 776: nit: `eager_sum_tensor`

funsor/tensor.py
fritzo: I agree this is useful, and I'm hoping @ordabayevy will generalize this in #482 once we have more than one use case.  This file seems like the right place for that general logic since `funsor.ops` should be agnostic to `inputs` and `domains` etc.

eb8680 (approved):

PR #491, opened 2021-03-15
Refactor ops to allow non-funsor parameters

fritzo:
Addresses #489 
partly pair coded with @eb8680 

This major op refactoring aims to unblock @ordabayevy in #477 and #482 by allowing ops to have arbitrary *arg,**kwarg parameters.  Changes include:
- `nullop` is renamed to `null` for consistency (`NullOp` remains unchanged).
- Dispatch tables are moved from op instances to op subclasses.
- `@make_op` has moved to a classmethod `@MyOpClass.make` e.g. `@BinaryOp.make`
- Each concrete op subclass has a fixed `.arity` attribute.
- Op patterns dispatch on only the first `[:arity]`-many args (and patterns are validated to ensure this).
- Ops are new weakly cached via `WeakValueDict`.
- `.subclass_register()` handlers have a different inferface, now including `cls`.
- There is now a single shared `WrappedTransformOp` class, and its instances each have an op.
- `ops.WrappedTransformOp(fn)` must now be written as a kwarg `ops.WrappedTransformOp(fn=fn)`
- `is_numeric_array()` is demoted from an `Op` to a `@singledispatch`.
- `ops.stack` interface has changed to `ops.stack(parts, dim)` as in NumPy and PyTorch.
- `funsor.tensor.stack` is replaced by `ops.stack`.
- `ops.einsum` interface has changed to `ops.einsum(operands, equation)` where `operands` is a list or tuple.
- `Finitary` funsor interface has changed from variously-many args to a single tuple arg.

## Tested
- refactoring is covered by existing tests


funsor/adjoint.py
```diff
@@ -49,15 +49,7 @@ def interpret(self, cls, *args):
...
-        lazy_args = [
-            self._eager_to_lazy.get(
-                id(arg)
-                if ops.is_numeric_array(arg) or not isinstance(arg, Hashable)
-                else arg,
-                arg,
-            )
-            for arg in args
-        ]
+        lazy_args = [self._eager_to_lazy.get(arg, arg) for arg in args]
```
fritzo, line 52: this was causing recursion errors because `.get()` calls `.__eq__` which calls `ops.equal` which creates a funsor etc. i'm unsure how to make this safer

.github/workflows/ci.yml
```diff
@@ -86,4 +86,4 @@ jobs:
         pip freeze
     - name: Run test
       run: |
-        CI=1 FUNSOR_BACKEND=jax make test
+        CI=1 JAX_ENABLE_X64=1 FUNSOR_BACKEND=jax make test
```
fritzo, line 89: @fehiepsi does this seem ok to you? I'm working around errors of the form
```
UserWarning: Explicitly requested dtype float64 requested in full is not available, and will be truncated to dtype float32. To enable more dtypes, set the jax_enable_x64 configuration option or the JAX_ENABLE_X64 shell environment variable. See https://github.com/google/jax#current-gotchas for more.
```

fritzo:
@eb8680 finally got the tests to pass 😓

eb8680 (approved):

fehiepsi:
For some reason, I am getting
```
E       AttributeError: 'DirichletMultinomial' object has no attribute 'data'
```
in NumPyro `test/test_handlers.py::test_collapse_beta_binomial`. The instance `DirichletMultinomial` has the value
```python
DirichletMultinomial(concentration=[1.5 0.5], total_count=10.0, value=Finitary(stack, (Tensor(7.0, OrderedDict(), 'real'), Number(3.0))))
```

.github/workflows/ci.yml
fehiepsi: I can address it in a follow-up PR.

fritzo:
@eb8680, @fehiepsi's issue with `DirichletMultinomial` sounds related to the [lazy test we disabled](https://github.com/pyro-ppl/funsor/pull/491/files#diff-4347c85e3f7c7b6325a851b81efa03efb7bfc4c500dbcfac315fd6b8ac007c91L660).  Do you think we should try to get that working again?

fehiepsi:
Yeah, enabling it raises two errors
+ `NotImplementedError: inhomogeneous total_count is not supported` (only with `torch` backend)
+ `NotImplementedError: TODO support converting Indep(Transform)`
but it seems to me `refactoring of stack` is more relevant.

Issue #493, opened 2021-03-15
Track issues with adjoint


ordabayevy:
This is to track issues related to using adjoint.

- [ ] `test_adjoint.py::test_sequential_sum_product_adjoint`:  `xfail_param(MarkovProduct, reason="mysteriously doubles adjoint values?")`

`MarkovProduct` is evaluated again at `actual.reduce(sum_op)` line:

```py
    with AdjointTape() as actual_tape:
        actual = impl(sum_op, prod_op, trans, time, {"prev": "curr"})
        actual = actual.reduce(sum_op)
```

It might be due to " `.reduce` binds variables, which triggers alpha-renaming and reinterprets `actual`?" from slack discussion with @eb8680 

- [ ] `MarkovProduct` is not expanded by `lazy` (or any other Funsor not defined in `adjoint_ops`). A temporary solution (#492 ) might be to define an adjoint for `MarkovProduct`. Still work in progress.

- [ ] `TraceEnum_ELBO` uses `AdjointTape` to calculate expectations. Under `eager` evaluation `nan` gradients appear for parameters in `test_enum_funsor.py::test_elbo_plate_plate`.

PR #494, opened 2021-03-17
Autodiff implementation (experimental)

ordabayevy:
This is an implementation of autodiff. The goal is to address issues in computing expectations in `TraceEnum_ELBO` and `TraceMarkovEnum_ELBO` (#493). As of now it seems to fix `nan` gradients under `eager` interpretation in `TraceEnum_ELBO`.

Gradients are calculated by transposing linearized functions.

1. **Linearize.** Variables that need to be linearized are replaced by primal- tangent tuple `JVP(primal, tangent)` and then pattern matched to propagate tangents, e.g.:

```py
JVP(x, dx) + JVP(y, dy) = JVP(x+y, dx+dy)
JVP(x, dx) * JVP(y, dy) = JVP(x*y, ydx + xdy)
JVP(x, dx) * y = JVP(x*y, ydx)
```

Out tangent is a linear function of in tangents. `JVP` is used for `(add,mul)` semiring and `LJVP` is used for `(logaddexp,add)` semiring.

2. **Transpose of a linear function.** Transpose is simply inverting the order of execution and transposing matrices, in this case swapping more primitive operations `.reduce(sum_op, "i")` and `.expand("i")` (broadcasting does this automatically).


funsor/terms.py
```diff
@@ -994,6 +1001,46 @@ def die_binary(op, lhs, rhs):
     raise NotImplementedError(f"Missing pattern for {repr(expr)}")
 
 
+class Expand(Funsor):
```
fritzo, line 1004: Hmm I'd like to better understand the need for this.

We've been trying to preserve the [extensionality](https://en.wikipedia.org/wiki/Extensionality) property in Funsor, which states that: if under every grounding substitution `subs` a pair of funsors `f,g` satisfy `f(**subs) == g(**subs)`, then it should be permissible for an optimizer to replace funsor `f` with funsor `g` in any expression.  IIUC this `Expand` funsor would break extensionality because `f.expand(...)` behaves as `f` under every grounding substitution.

PR #495, opened 2021-03-17
Extract A-normal form computation from stack_reinterpret

eb8680:
This PR refactors `funsor.interpreter.stack_reinterpret` into two functions `anf` and `stack_reinterpret`, where `anf` transforms a Funsor expression into a topologically ordered `OrderedDict` of subexpressions and `stack_reinterpret` `interpret()`s the subexpressions in order, using the `OrderedDict` to avoid evaluating expressions twice.

Exposing this representation of Funsor expressions should eventually make it easier to implement transformations like `adjoint` or `solve` that traverse expressions in reverse without resorting to a tape, or that do not need to traverse the entire expression (like alpha-renaming).

I would also like to refactor `recursion_reinterpret` to use `anf` to ensure that traversal order is always correct.

Tested: pure refactoring exercised by existing tests with `FUNSOR_USE_TCO=1`.

### DIFF
--- a/funsor/interpreter.py
+++ b/funsor/interpreter.py
@@ -6,7 +6,7 @@ import os
 import re
 import types
 import warnings
-from collections import OrderedDict
+from collections import OrderedDict, deque
 from functools import singledispatch
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


funsor/interpreter.py
```diff
@@ -216,67 +215,49 @@ def gensym(x=None):
...
-    parent_to_children = OrderedDict()
-    child_to_parents = OrderedDict()
+def anf(x):
+    stack = [x]
+    child_to_parents, children_counts = {}, {}
+    leaves = []
     while stack:
-        h_name, h = stack.pop(0)
-        parent_to_children[h_name] = []
+        h = stack.pop(0)
```
fritzo, line 223: It would be cheaper to use a [deque](https://docs.python.org/3/library/collections.html#collections.deque) for stack and leaves.

fritzo (approved):
Nice refactoring!

PR #499, opened 2021-03-18
Share more code between reinterpret implementations

eb8680:
This PR refactors `recursion_reinterpret` to use the same utilities (`is_atom` and `children`) as `stack_reinterpret` and simplifies its implementation by removing a `singledispatch`, which should also slightly simplify stack traces.

Tested: pure refactoring, exercised everywhere `reinterpret` is called


fritzo (approved):
💯 These simplifications look great.

PR #500, opened 2021-03-18
Try to use canonical dtype as much as possible in JAX backend

fehiepsi:
as title


fritzo (approved):
Thanks for fixing this @fehiepsi !

PR #501, opened 2021-03-18
Add test to catch eager expand issue in jax backend

fehiepsi:
### TODO
- [x] isolate the reason and ~make the test passes~


fehiepsi:
It seems that this is expected. In `Tensor.unscaled_sample`, we always apply a [dice factor](https://github.com/pyro-ppl/funsor/blob/master/funsor/tensor.py#L394) in `jax` backend; while in `torch` backend, we only want apply it if `logits.requires_grad` is True.

cc @fritzo @eb8680 in case this is not expected

PR #504, opened 2021-03-19
Add a generic op factory make_op

fritzo:
Addresses #351
pair coded with @eb8680 @ordabayevy @fehiepsi 

Implements a `@make_op` decorator, automating e.g. `@BinaryOp.make` + `@find_domain.register` + `@eager.register`.

## Tested
- three unit tests of different domains

### DIFF
--- a/funsor/op_factory.py
+++ b/funsor/op_factory.py
@@ -70,7 +70,7 @@ def make_op(fn):
 
     # Register an eager funsor rule.
     # TODO generalize to more funsor types, ideally to Funsor itself.
-    pattern = [funsor_cls, op_cls] + [(Number, Tuple, Tensor)] * arity
+    pattern = [funsor_cls, type(op)] + [(Number, Tuple, Tensor)] * arity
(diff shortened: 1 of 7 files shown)


funsor/op_factory.py
```diff
@@ -0,0 +1,76 @@
...
+            for arg, hint in zip(args, hints):
+                if hint is not None:
+                    assert deep_issubclass(arg, hint)
+        if isinstance(output_type, Dependent):
+            return output_type(**dict(zip(parameters, args)))
+        return output_type
+
+    # Register an eager funsor rule.
+    # TODO generalize to more funsor types, ideally to Funsor itself.
+    pattern = [funsor_cls, op_cls] + [(Number, Tuple, Tensor)] * arity
```
eb8680, line 73: This needs to be `pattern = [funsor_cls, type(op)] + ...`. That might also fix the ambiguity errors from using `Funsor` in the pattern.

eb8680 (approved):

PR #505, opened 2021-03-20
Use new ops to reduce multiple dims in eager_reduce

ordabayevy:
This simplifies `eager_reduce` by using new `ops` (#482) that can reduce over multiple dims.


fritzo (approved):
Thanks, this is great!

PR #506, opened 2021-03-21
Allow stack Numbers

fehiepsi:
Fixes https://github.com/pyro-ppl/numpyro/issues/964

After https://github.com/pyro-ppl/funsor/pull/491, we are no longer able to stack Numbers. This PR relaxes this restriction because `numpy`, `jax` backends can perform stack over Python scalars

```python
import funsor; funsor.set_backend("jax")
from funsor import ops, Number
ops.stack((Number(2.), Number(3)))
```

### DIFF
--- a/funsor/ops/array.py
+++ b/funsor/ops/array.py
@@ -409,7 +409,7 @@ def stack(parts, dim=0):
     raise NotImplementedError
 
 
-stack.register(arraylist)(np.stack)
+stack.register(typing.Tuple[typing.Union[array + (int, float)], ...])(np.stack)
 
 
 @UnaryOp.make
--- a/funsor/terms.py
+++ b/funsor/terms.py
@@ -1874,6 +1874,7 @@ def symbolic(*signature):
             return _symbolic(inputs, output, fn)
     # Usage: @symbolic(Real, Reals[3], Bint[3])
     output = None
+    # FIXME: what is inputs?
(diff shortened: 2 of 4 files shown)


funsor/terms.py
```diff
@@ -1874,6 +1874,7 @@ def xpyi(x: Real, y: Reals[3], i: Bint[3]):
             return _symbolic(inputs, output, fn)
     # Usage: @symbolic(Real, Reals[3], Bint[3])
     output = None
+    # FIXME: what is inputs?
```
fritzo, line 1877: good catch, i'll look into a fix...

funsor/ops/array.py
```diff
@@ -409,7 +409,7 @@ def stack(parts, dim=0):
     raise NotImplementedError
 
 
-stack.register(arraylist)(np.stack)
+stack.register(typing.Tuple[typing.Union[array + (int, float)], ...])(np.stack)
```
fritzo, line 412: Hmm, I wonder if there's a way to make this backend-agnostic, maybe
```py
stack.register(typing.Tuple[typing.Union[array], ...])(np.stack)
stack.register(typing.Tuple[typing.Union[int, float], ...])(make_array)
```
Anyway, I'll merge now since this is high priority, and we can support `backend==torch` later.

fritzo (approved):
Thanks for fixing this, and sorry we broke it!

PR #507, opened 2021-03-21
Support nn.Modules in @function, make_funsor, and make_op

fritzo:
This adds support for `torch.nn.Module` instances in `@function`, `make_funsor()`, and `make_op`.

This also fixes a couple unrelated docs errors:
- addes docstring to examples/forward_backward.py
- changes scope of Op.arity cvar

## Tested
- [x] refactored VAE example to test `@function` (this example will later be changed to use `make_op()`)
- [x] unit test for `make_funsor()`
- [x] unit test for `make_op()`


test/examples/test_bart.py
```diff
@@ -31,11 +32,10 @@ def bounded_exp(x, bound):
...
 
-@funsor.function(
-    Reals[2 * num_origins * num_destins],
-    (Reals[num_origins, num_destins, 2], Reals[num_origins, num_destins]),
-)
-def unpack_gate_rate(gate_rate):
+@funsor.function
+def unpack_gate_rate(
+    gate_rate: Reals[2 * num_origins * num_destins],
+) -> typing.Tuple[Reals[num_origins, num_destins, 2], Reals[num_origins, num_destins]]:
```
eb8680, line 38: Does this syntax already work, or does this `Tuple` need to be a `funsor.domains.Product`?

test/test_factory.py
```diff
@@ -172,6 +173,27 @@ def _(loc, scale, value):
...
 
 
+@pytest.mark.skipif(get_backend() != "torch", reason="requires nn.Module")
+def test_nn_module():
+    import torch
+
+    class Matmul(torch.nn.Module):
+        def forward(
+            self,
+            x: Funsor,
```
eb8680, line 183: Is this a pattern (Funsor terms written as `nn.Module`s) we want to support? An alternative would be to raise an error when passing `Module` instances to `make_funsor` and request that they be converted to `Op`s instead.

test/test_factory.py
```diff
@@ -172,6 +173,27 @@ def _(loc, scale, value):
...
+            y: Funsor,
+            i: Bound,
+        ) -> Fresh[lambda x: x]:
```
eb8680, line 186: Now that we have a `Dependent` I guess in a separate PR we should update the syntax of `make_funsor` to use `Dependent` rather than `Fresh` for return types in `make_funsor`.

test/test_op_factory.py
```diff
@@ -119,3 +120,40 @@ def matmul(
...
+
+
+@pytest.mark.skipif(get_backend() != "torch", reason="requires nn.Module")
+@pytest.mark.parametrize("batch_shape", [(), (10,)], ids=str)
+def test_nn_module(batch_shape):
+    import torch
+
+    class Matmul(torch.nn.Module):
+        def forward(
+            self, x, y
```
eb8680, line 132: One potential problem with converting `Module`s to `Op`s is that preexisting `Module`s may be especially likely to have type signatures that are not compatible with `make_op`. For example, someone might have written `Matmul` as
```py
def forward(self, x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
    ...
```
Should we do any preprocessing of the signature to either convert it to a Funsor-compatible form, raise an error, or raise a warning reminding the user to register a `find_domain` implementation for the new op?

eb8680 (commented):
Code looks good, I just have some design questions.

test/test_factory.py
fritzo: I don't see any reason to not support `Funsor` subclasses based on `nn.Module` s.  Forcing users to make an op would force them into the world of positional tensors.  But I haven't thought this through entirely.

test/examples/test_bart.py
fritzo: The tests pass 🤷

test/test_factory.py
fritzo: Sounds reasonable. I guess we can make `Fresh` subclass `Dependent` when introducing fresh variables.

test/test_op_factory.py
fritzo: Hopefully things just break with `torch.Tensor` signatures, and users can wrap their modules with funsor compatible types.  That's out of scope for this little Sunday morning PR though.

eb8680 (approved):

PR #509, opened 2021-03-27
Named tensor notation tutorial Part 1

ordabayevy:
This is part 1 in the series of tutorials based on [Named Tensor Notation](https://namedtensor.github.io/). Any suggestions & corrections are very welcome.

[Link to the tutorial](https://nbviewer.jupyter.org/github/ordabayevy/funsor/blob/named-tensor-i/tutorials/named_tensor_notation_i.ipynb)

PR #511, opened 2021-03-29
Fix printing of Tensors

fritzo:
Addresses ugly printing in #509 

I'd like to merge this before #509 so the tutorial looks nice.

## Tested
- [x] unit test for domains
- [x] unit test for tensors

### DIFF
--- a/funsor/tensor.py
+++ b/funsor/tensor.py
@@ -151,7 +151,7 @@ class Tensor(Funsor, metaclass=TensorMeta):
 
     @ignore_jit_warnings()
     def __repr__(self):
-        data = str(self.data).replace("\n", "\n       ")
+        data = repr(self.data).replace("\n", "\n       ")
         inputs = dict.__repr__(self.inputs)
         if self.dtype != "real":
             return "Tensor({}, {}, {})".format(data, inputs, repr(self.dtype))
--- a/funsor/tensor.py
+++ b/funsor/tensor.py
@@ -102,7 +102,7 @@ class TensorMeta(FunsorMeta):
     def __call__(cls, data, inputs=None, dtype="real"):
         if inputs is None:
             inputs = tuple()
-        elif isinstance(inputs, OrderedDict):
+        elif isinstance(inputs, dict):
             inputs = tuple(inputs.items())
         # XXX: memoize tests fail for np.generic because those scalar values are hashable?
         # it seems that there is no harm with the conversion generic -> ndarray here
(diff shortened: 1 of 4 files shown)


funsor/tensor.py
```diff
@@ -122,14 +122,14 @@ class Tensor(Funsor, metaclass=TensorMeta):
...
-        x = Tensor(data, OrderedDict([("i", Bint[5]), ("j", Bint[4])]))
+        x = Tensor(data, {"i": Bint[5], "j": Bint[4]})
         assert x.output == Reals[3, 2]
 
     Operators like ``matmul`` and ``.sum()`` operate only on the output shape,
     and will not change the named inputs.
 
     :param numeric_array data: A PyTorch tensor or NumPy ndarray.
-    :param OrderedDict inputs: An optional mapping from input name (str) to
+    :param dict inputs: An optional mapping from input name (str) to
```
fritzo, line 132: this is safe because we support only CPython 3.6+

ordabayevy:
Somewhat related to this, when Tensors are printed the first row is misaligned because of the additional `Tensor(` string (see output of cell 6 in #509). Do you think that should also be fixed @fritzo?

funsor/tensor.py
```diff
@@ -151,19 +151,21 @@ def __init__(self, data, inputs=None, dtype="real"):
...
     @ignore_jit_warnings()
     def __repr__(self):
-        if self.output != "real":
-            return "Tensor({}, {}, {})".format(self.data, self.inputs, repr(self.dtype))
+        inputs = dict.__repr__(self.inputs)
+        if self.dtype != "real":
+            return "Tensor({}, {}, {})".format(self.data, inputs, repr(self.dtype))
         elif self.inputs:
-            return "Tensor({}, {})".format(self.data, self.inputs)
+            return "Tensor({}, {})".format(self.data, inputs)
```
fehiepsi, line 158: WDYT of stripping `tensor(...)` here? So we get `Tensor([1, 2, 3])` instead of `Tensor(tensor([1, 2, 3]))`.

fritzo: Hmm I'd prefer for now to be more verbose so that the result can be evaluated back to the original.

In the future I think we could elide the `tensor(...)` if we made `TensorMeta` promote `list` objects via `ops.make_array()` or something (dispatching to `np.array` or `torch.tensor`), but that seems out of scope for this PR. Feel free to add an `ops.make_array()` and change this, though.

fritzo:
> Tensors are printed the first row is misaligned

Good point, I've fixed this.

funsor/tensor.py
fehiepsi: Makes sense to me. Thanks for explaining.

fehiepsi: (by the way, asking for an opinion is not a request - so no need to put a thumb down)

fehiepsi (approved):
LGTM. Going back to the first time playing with funsor, I found `OrderedDict` is a bit ugly. But for now, when getting more familiar with it, I think I will miss OrderedDict representation...

PR #516, opened 2021-04-02
Adam interpretation

fritzo:
group coded with @eb8680 @fehiepsi @ordabayevy 

Implements an `Adam` interpretation of `ops.min` and `ops.max`.

## Tested
- [x] added an example
- [x] added include check to `make lint` (which had failed on an earlier commit)


fritzo:
@fehiepsi do I need to do anything to get this to appear in docs?

fehiepsi:
Yes, you'll need to add the example to the same index file.

eb8680 (approved):

PR #517, opened 2021-04-03
Fix a scatter eager_subs issue

fehiepsi:
Pair-coded with @eb8680 @fritzo @ordabayevy 

This bug is detected while running a test for infer_discrete with NumPyro scan

### DIFF
--- a/funsor/terms.py
+++ b/funsor/terms.py
@@ -1222,8 +1222,7 @@ class Scatter(Funsor):
         subs = OrderedDict(subs)
         new_subs = []
         for name, sub in self.subs:
-            if name in subs:
-                assert isinstance(subs[name], Variable)
+            if name in subs and isinstance(subs[name], Variable):
                 new_subs.append((subs[name].name, sub))
             else:
                 new_subs.append((name, sub))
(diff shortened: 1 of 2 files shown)


funsor/terms.py
```diff
@@ -1218,6 +1218,17 @@ def _alpha_convert(self, alpha_subs):
         reduced_vars = frozenset(alpha_subs.get(var.name, var) for var in reduced_vars)
         return op, subs, source, reduced_vars
 
+    def eager_subs(self, subs):
+        subs = OrderedDict(subs)
+        new_subs = []
+        for name, sub in self.subs:
+            if name in subs:
+                assert isinstance(subs[name], Variable)
```
fehiepsi, line 1226: Remove this assertation seem to resolve the Slice substitution issue

(there is another issue happens in `test_scan_hmm` that `@_get_support_value.register(funsor.cnf.Contraction)` couldn't extract the `delta_terms` - but I guess that issue is not related to this)

eb8680: Hmm, I guess what's happening is that there can be `bound` variables of `Scatter` that share names with the `fresh` variables, and we need to implement a better `_alpha_convert` method that properly handles this case. Rather than just removing the assertion, does changing this to
```py
if name in subs and isinstance(subs[name], Variable):
    new_subs.append((subs[name].name, sub))
```
fix the downstream error with `Slice`?  It's still not correct in general, but it might be enough to unblock you.

fehiepsi: Yes, it does resolving the Slice issue (but the test still fails by the `_get_support_value` issue mentioned above).

eb8680 (approved):

Issue #481, opened 2021-03-06
What is the type hint of variables bound inside (e.g., `ax` in Softmax)


ordabayevy:
What should be the type hint for `ax` in this `Softmax` function?
1. `Funsor` works but then then `ax` cannot be a `str`
2. `Value[str]` can only be a `str` and not `Variable`
3. `Bound` works both as `str` and `Variable` but `ax` is not actually bound

```py
@make_funsor
def Softmax(
    x: Funsor,
    ax: ? # Funsor or Bound or Value[str]
) -> Fresh[lambda x: x]:
    return x.exp() / x.exp().reduce(ops.add, ax)
```

fritzo:
You'll need `ax: Bound` and `ax2: Fresh[lambda ax: ax]`. As we discussed yesterday, `@make_funsor` currently does not support binding-and-return of a dimension with the same name, so for now you'll need to rename the dimension. I believe @eb8680 is planning to address that as part of alpha-renaming work.
```py
@make_funsor
def Softmax(
    x: Funsor,
    ax: Bound,
    ax2: Fresh[lambda ax: ax],
) -> Fresh[lambda x: x]:
    x = x(**{ax: ax2})  # rename to work around alpha renaming limitations
    y = x - x.reduce(ops.logaddexp, ax2)
    return y.exp()
```

PR #518, opened 2021-04-04
BindReturn type hint for make_funsor

ordabayevy:
Addresses #481.

`BindReturn` type hint is used for binding and returning a variable. For example:

```py
@make_funsor
def Unroll(
    x: Has[{"ax"}],  # noqa: F821
    ax: BindReturn[lambda ax, k: Bint[ax.size - k + 1]],
    k: Value[int],
    kernel: Fresh[lambda k: Bint[k]],
) -> Fresh[lambda x: x]:
    return x(**{ax.name: ax + kernel})

x = random_tensor(OrderedDict(a=Bint[5]))
y = Unroll(x, "a", 2, "kernel")
check_funsor(y, {"a": Bint[5 - 2 + 1], "kernel": Bint[2]}, Real)
```

or

```py
@make_funsor
def Softmax(
    x: Has[{"ax"}],  # noqa: F821
    ax: BindReturn[lambda ax: ax],
) -> Fresh[lambda x: x]:
    y = x - x.reduce(ops.logaddexp, ax)
    return y.exp()

x = random_tensor(OrderedDict(a=Bint[3], b=Bint[4]))
y = Softmax(x, "a")
check_funsor(y, {"a": Bint[3], "b": Bint[4]}, Real)
```

### DIFF
--- a/funsor/factory.py
+++ b/funsor/factory.py
@@ -67,10 +67,10 @@ class Bound:
     pass
 
 
-class BindReturn(Bound):
+class BindReturn:
     """
     Type hint for :func:`make_funsor` decorated functions. This provides hints
-    for bind-return variables (names).
+    for variables (names) that are bound and returned.
     """
 
     pass
@@ -161,6 +161,7 @@ def make_funsor(fn):
 
     - Funsor inputs are typed :class:`~funsor.terms.Funsor`.
     - Bound variable inputs (names) are typed :class:`Bound`.
+    - Bind and Return variable inputs (names) are typed :class:`BindReturn`.
     - Fresh variable inputs (names) are typed :class:`Fresh` together with
       lambda to compute the dependent domain.
     - Ground value inputs (e.g. Python ints) are typed :class:`Value` together with
@@ -186,7 +187,9 @@ def make_funsor(fn):
     """
     input_types = typing.get_type_hints(as_callable(fn))
     for name, hint in input_types.items():
-        if not (hint in (Funsor, Bound, BindReturn) or isinstance(hint, (Fresh, Value, Has))):
+        if not (
+            hint in (Funsor, Bound, BindReturn) or isinstance(hint, (Fresh, Value, Has))
+        ):
             raise TypeError(f"Invalid type hint {name}: {hint}")
     output_type = input_types.pop("return")
     hints = tuple(input_types.values())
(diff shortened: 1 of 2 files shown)


eb8680:
Thanks for adding this! WDYT about making `BindReturn` the default behavior of `Fresh`? That would be consistent with default behavior in existing terms like `Cat`, `Stack` and `Independent` and avoid growing the number of special `make_funsor` type annotations. The conservative alternative is having `make_funsor` raise an error when a `Fresh` variable appears in the inputs of another argument.

ordabayevy:
> WDYT about making BindReturn the default behavior of Fresh? 

Do you mean that in the example below `Fresh` type hint would be smart to make `ax` both `bound` and `fresh` and make `kernel` only `fresh`? 

```py
@make_funsor
def Unroll(
    x: Has[{"ax"}],  # noqa: F821
    ax: Fresh[lambda ax, k: Bint[ax.size - k + 1]],
    k: Value[int],
    kernel: Fresh[lambda k: Bint[k]],
) -> Fresh[lambda x: x]:
    return x(**{ax.name: ax + kernel})
```

eb8680:
> Do you mean that in the example below Fresh type hint would be smart to make ax both bound and fresh and make kernel only fresh?

Yes, exactly. 

On a related note, we should also start using `funsor.domains.Dependent` rather than `Fresh` for annotating return types of `make_funsor`, but that's for another PR.

ordabayevy:
> Yes, exactly.

I like the idea. I'll make the changes then.

eb8680:
Sorry for taking so long to review this (especially since I suggested you try it in the first place). I am still not sure how to go about fixing alpha-conversion as a whole in a way that remains compatible with cons-hashing, so I had put off thinking about details.

I think for the behavior implemented in this PR to be safe and correct by construction in general, we would need to *eagerly* alpha-mangle the arguments to a `make_funsor` term with a bound `Fresh` variable *before* evaluating that term with a rewrite rule to guarantee that there are no collisions between the fresh variable and the implicitly bound variable inside the rule body. In fact, we could always do this for any Funsor; the fact that we only alpha-convert in `reflect` is an (important) optimization.

We could write a simple decorator for rewrite rules to perform this extra step:
```py
def bind_args(term):

    def binding_wrapper(rule):
        def wrapped_rule(*args):
            mangled_args = reflect.interpret(term, *args)._ast_values
            return rule(*mangled_args)
        return functools.wraps(rule)(wrapped_rule)

    return binding_wrapper
```
To illustrate the use of `bind_args`, in your `Softmax` example, we could separate the term definition and the default reduction rule and manually apply `bind_args` to the rule:
```py
@make_funsor
def Softmax(
    x: Has[{"ax"}],  # noqa: F821
    ax: Fresh[lambda ax: ax],
) -> Fresh[lambda x: x]:
    return None

@eager.register(Softmax, Tensor, Variable)
@bind_args(Softmax)
def _eager_softmax(x, ax):
    y = x - x.reduce(ops.logaddexp, ax)
    return y.exp()
```
Of course, this is less ergonomic than the original syntax, so I could imagine folding `bind_args` into `make_funsor` or even into interpretations, although this would come at a considerable computational cost with the current implementation of alpha-conversion and is not necessary for this particular example.

PR #519, opened 2021-04-04
Reshape using Variable in Named Tensor tutorial Part 1

ordabayevy:
I want to change reshaping in the tutorial to use `Variable` instead of `Tensor`. (Using `Tensor` feels more like indexing rather than reshaping.)

Use

```py
layer = Variable("layer", Bint[9])

A_layer = A(height=layer // Number(3, 4), width=layer % Number(3, 4))
```

instead of

```py
layer = Tensor(tensor([0, 1, 2, 3, 4, 5, 6, 7, 8]), dtype=9)["layer"]

A(height=layer // Number(3, 4), width=layer % Number(3, 4))
```

Also add reshaping back `layer->(height,width)` example:

```py
height = Variable("height", Bint[3])
width = Variable("width", Bint[3])

A_layer(layer=height * Number(3, 4) + width % Number(3, 4))
```

### DIFF
--- a/tutorials/named_tensor_notation_i.ipynb
+++ b/tutorials/named_tensor_notation_i.ipynb
@@ -42,7 +42,8 @@
     "\n",
     "import funsor\n",
     "import funsor.ops as ops\n",
-    "from funsor import Number, Tensor\n",
+    "from funsor import Number, Tensor, Variable\n",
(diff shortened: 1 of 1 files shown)


eb8680 (approved):

ordabayevy:
Thanks @eb8680 !

PR #520, opened 2021-04-04
Talbot's method for numerical inversion of the Laplace transform

ordabayevy:
This is the implementation of Talbot's method for numerical inversion of the Laplace transform. It is adapted from Tucker McClure's [MATLAB implementation](https://www.mathworks.com/matlabcentral/fileexchange/39035-numerical-inverse-laplace-transform). Example is the [n-step kinetic model](https://www.sciencedirect.com/science/article/pii/S0006349503746487) used for the analysis of helicase catalyzed DNA unwinding experiments.

Does it look like an interesting example to add to funsor library?


funsor/torch/ops.py
```diff
@@ -332,7 +332,7 @@ def _pow(x, y):
 
 @ops.reciprocal.register(torch.Tensor)
 def _reciprocal(x):
-    result = x.reciprocal().clamp(max=torch.finfo(x.dtype).max)
+    result = x.reciprocal()  # .clamp(max=torch.finfo(x.dtype).max)
```
ordabayevy, line 335: `.clamp` doesn't work with `torch.complex64` data type.

funsor/inverse_laplace.py
```diff
@@ -0,0 +1,80 @@
...
+    Tucker McClure (2021). Numerical Inverse Laplace Transform
+    (https://www.mathworks.com/matlabcentral/fileexchange/39035-numerical-inverse-laplace-transform),
+    MATLAB Central File Exchange. Retrieved April 4, 2021.
+
+    :param num_steps: number of terms to sum for each t.
+    """
+
+    def __init__(self, num_steps):
+        super().__init__("talbot")
+        self.N = num_steps
```
eb8680, line 48: nit: just name this `self.num_steps`

funsor/inverse_laplace.py
```diff
@@ -0,0 +1,80 @@
...
+            - 1j / (pi / self.N * k).tan()
+        ) * delta[1:].exp()
+
+        delta = Tensor(delta)["num_steps"]
+        gamma = Tensor(gamma)["num_steps"]
+        ilt = 0.4 / t * (gamma * F(**{s.name: delta / t})).reduce(ops.add, "num_steps")
+
+        return Tensor(ilt.data.real, ilt.inputs)
+    else:
+        raise NotImplementedError(f"Unsupported backend {get_backend()}")
```
eb8680, line 75: From #516 and this PR, it seems like this check is a useful pattern we could abstract into a helper context manager/decorator `backends_supported(*backends)` that takes a sequence of backends and raises `NotImplementedError` if the current backend is not among them:
```py
@Talbot.register(...)
@backends_supported("torch")
def talbot(...):
    k = torch.arange(...)
    ...
    return Tensor(...)
```
Maybe add that in a followup PR?

funsor/inverse_laplace.py
```diff
@@ -0,0 +1,80 @@
...
+
+from . import ops
+
+
+@make_funsor
+def InverseLaplace(
+    F: Has[{"s"}], t: Funsor, s: Bound  # noqa: F821
+) -> Fresh[lambda F: F]:
+    """
+    Inverse Laplace transform of function F.
```
eb8680, line 20: Say explicitly that there is no closed-form solution for arbitrary `F`, so we must resort to numerical approximations which we store in a new interpretation `Talbot`.

funsor/inverse_laplace.py
```diff
@@ -0,0 +1,80 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+from math import pi
```
eb8680, line 4: nit: use `math.pi` rather than `pi`

eb8680 (commented):
Nice example! I think it would make more sense to put the stuff in `inverse_laplace.py` into the example script. That way we can highlight it as a simple, self-contained demonstration of defining a new term, a new interpretation and a new approximate pattern.

funsor/inverse_laplace.py
ordabayevy: > Maybe add that in a followup PR?

Sounds good.

examples/talbot.py
```diff
@@ -0,0 +1,165 @@
...
+            * k
+            * (1 + 1 / (math.pi / self.num_steps * k).tan() ** 2)
+            - 1j / (math.pi / self.num_steps * k).tan()
+        ) * delta[1:].exp()
+
+        delta = Tensor(delta)["num_steps"]
+        gamma = Tensor(gamma)["num_steps"]
+        ilt = 0.4 / t * (gamma * F(**{s.name: delta / t})).reduce(ops.add, "num_steps")
+
+        return Tensor(ilt.data.real, ilt.inputs)
```
ordabayevy, line 94: I feel like there should be some kind of support/interface for complex data type. Complex number can come up in many numerical approximation methods.

eb8680 (approved):

ordabayevy:
Thanks for reviewing @eb8680 !

PR #522, opened 2021-04-09
backends_supported helper function

ordabayevy:
Follow up to https://github.com/pyro-ppl/funsor/pull/520#discussion_r606908196

Helper decorator/context manager that raises error if `get_backend()` not in `backends_supported`.

### DIFF
--- a/funsor/adam.py
+++ b/funsor/adam.py
@@ -3,6 +3,7 @@
 
 import funsor.ops as ops
 from funsor.domains import RealsType
+from funsor.factory import backends_supported
 from funsor.interpretations import StatefulInterpretation
 from funsor.tensor import Tensor
 from funsor.terms import Funsor, Reduce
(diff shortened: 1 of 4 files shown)


fritzo (changes requested):
Hmm, we really do want dispatch here, rather than error. Those `NotImplementedError`s mean not-yet-implemented.  The decorator proposed in this PR would close off the path to implementing them.

ordabayevy:
The idea is to add new supported backends to the `backends_supported` args as  they are implemented:

```py
@backends_supported("torch", "jax", ...)
...
```

That shouldn't block new implementations I believe. This should just delegate the logic of checking available backends to `backends_supported`.

funsor/adam.py
```diff
@@ -35,41 +35,38 @@ def param(self, name, domain=None):
...
-            import torch
+        import torch
 
-            return Tensor(torch.randn(domain.shape, requires_grad=True))
-        raise NotImplementedError(f"Unsupported backend {get_backend()}")
+        return Tensor(torch.randn(domain.shape, requires_grad=True))
 
 
 @Adam.register(Reduce, ops.MinOp, Funsor, frozenset)
+@backends_supported("torch")
```
fritzo, line 48: Can you please revert this file?  I really do want `if - else - ` logic here.

PR #523, opened 2021-04-09
Reimplement substitute() using anf()

fritzo:
Blocking #414 (for performance reasons)
group coded with @eb8680 @ordabayevy @fehiepsi

### DIFF
--- a/funsor/interpreter.py
+++ b/funsor/interpreter.py
@@ -6,7 +6,7 @@ import os
 import re
 import types
 import warnings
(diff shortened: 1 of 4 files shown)


fritzo:
Looks like this is failing at
```sh
pytest -x test/pyro/test_hmm.py::test_gaussian_mrf_log_prob
```

fritzo (commented):
Thanks for fixing @eb8680!

eb8680 (approved):

PR #527, opened 2021-04-13
Clarify meaning of reduced_vars in Scatter docs

fritzo:
pair coded with @eb8680


fehiepsi (approved):
Makes sense to me. It is quite tricky to perform `scatter_add` properly so even making a correct behavior already amazed me.

Issue #533, opened 2021-04-21
Delayed param in minipyro


ordabayevy:
This issue proposes to create delayed `pyro.param` by `log_joint` and then use `funsor.adam.Adam` to optimize parameters. This would allow to write optimization part in `minipyro` in a way that is more consistent with funsor style (if there is such a thing) and also  make it backend-agnostic (with more supported backends in `funsor.adam.Adam`). Support for constrained variables (#502) might be useful here.

```py
class log_joint(Messenger):
    ...
    def process_message(self, msg):
        if msg["type"] == "param":
            msg["value"] = funsor.Variable(msg["name"], msg["output"])

class SVI:
    ...
    def run(self, *args, **kwargs):
        ...
        loss = ...
        with funsor.montecarlo.MonteCarlo():
            with funsor.adam.Adam(**options):
                loss.reduce(ops.min)
```

(I have tried this out for a simple guide where `pyro.param` doesn't have any constraints and not nested in `pyro.plate` and it seems to work.)

fritzo:
This is great idea! It's a near literal translation of "(stochastic) (gradient descent)".

PR #534, opened 2021-04-23
Delayed param

ordabayevy:
Addresses #533

### DIFF
--- a/funsor/minipyro.py
+++ b/funsor/minipyro.py
@@ -24,6 +24,8 @@ from pyro.distributions import validation_enabled
 from pyro.optim.clipped_adam import ClippedAdam as _ClippedAdam
 
 import funsor
+from funsor.adam import Adam
+import funsor.ops as ops
 
 
 # Funsor repreresents distributions in a fundamentally different way from
@@ -234,6 +236,10 @@ class log_joint(Messenger):
             if msg["value"] is None:
                 # Create a delayed sample.
                 msg["value"] = funsor.Variable(msg["name"], msg["fn"].output)
+        elif msg["type"] == "param":
+            if msg["value"] is None:
+                # Create a delayed sample.
+                msg["value"] = funsor.Variable(msg["name"], msg["output"])
 
     def postprocess_message(self, msg):
         if msg["type"] == "sample":
@@ -388,8 +394,8 @@ class PyroOptim(object):
 
 
 # We wrap some commonly used PyTorch optimizers.
-class Adam(PyroOptim):
-    TorchOptimizer = torch.optim.Adam
+#  class Adam(PyroOptim):
+#      TorchOptimizer = torch.optim.Adam
 
 
 class ClippedAdam(PyroOptim):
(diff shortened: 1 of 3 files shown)


funsor/minipyro.py
```diff
@@ -21,9 +21,9 @@
 
 import torch
 from pyro.distributions import validation_enabled
-from pyro.optim.clipped_adam import ClippedAdam as _ClippedAdam
 
 import funsor
+from funsor.adam import Adam  # noqa: F401
```
ordabayevy, line 26: for pyroapi compatibility

funsor/minipyro.py
```diff
@@ -306,11 +330,13 @@ def param(
...
-        if event_dim is None:
-            event_dim = init_value.dim()
-        output = funsor.Reals[init_value.shape[init_value.dim() - event_dim :]]
+    # infer output
+    value = init_value
+    if value is None:
+        value, _ = PARAM_STORE[name]
+    if event_dim is None:
+        event_dim = value.dim()
+    output = funsor.Reals[value.shape[value.dim() - event_dim :]]
```
ordabayevy, line 339: infer `output` when `pyro.param` was already defined elsewhere

funsor/minipyro.py
```diff
@@ -408,24 +400,33 @@ def __init__(self, model, guide, optim, loss):
...
+            name = name.replace("_unconstrained", "")
+            value = value.data
+            old_value, constraint = PARAM_STORE[name]
+            if old_value is not value:
+                old_value.data.copy_(value)
+        return result
+
+    def step(self, *args, **kwargs):
+        self.optim.num_steps = 1
+        return self.run(*args, **kwargs)
```
ordabayevy, line 429: for compatibility with `SVI` interface

funsor/minipyro.py
```diff
@@ -408,24 +400,33 @@ def __init__(self, model, guide, optim, loss):
...
-        funsor.to_data(loss).backward()
-        # Grab all the parameters from the trace.
-        params = [site["value"].data.unconstrained() for site in param_capture.values()]
-        # Take a step w.r.t. each parameter in params.
-        self.optim(params)
-        # Zero out the gradients so that they don't accumulate.
-        for p in params:
-            p.grad = torch.zeros_like(p.grad)
-        return loss.item()
+        with funsor.terms.lazy:
```
ordabayevy, line 407: `lazy` interpretation is needed here to make sure that `funsor.Integrate` is not eagerly expanded in `Expectation`

ordabayevy:
`examples/minipyro.py` is currently failing with jit. My guess is that jitting needs to be baked into `funsor.adam.Adam`, I will think more about this.

ordabayevy:
Am I right that when using `funsor.adam.Adam` the function that needs to be jit traced is the loss function below (`Subs` funsor) @fritzo @eb8680 ? If yes, then it first needs to be converted to  a function with positional arguments?

```py
step_loss = loss(**{k: v[...] for k, v in params.items()}).data
```

funsor/minipyro.py
fritzo: Hmm, let's think about alternative workarounds... One issue here is that the Adam optimizer statistics would not be persisted across svi steps.

One option is simply to change pyroapi's SVI interface to look for either `.run()` or if missing fall back to `.step()`.

fritzo:
> ...failing with jit. My guess is that jitting needs to be baked into `funsor.adam.Adam`

I think you're right, but let's discuss.  That's a little different from Pyro where jit is baked into ELBO subclasses.

PR #537, opened 2021-05-04
`apply_optimizer` bug for terms in `sakka_bilmes_product`

ordabayevy:
There are terms in `sarkka_bilmes_product` that are incorrectly evaluated by `apply_optimizer` (#536). The test in this PR replicates one such expression.


ordabayevy:
Closing this PR since I believe the root cause of the issue here is related to alpha-conversion edge cases reported in #414 .

PR #543, opened 2021-07-16
ProvenanceTensor

ordabayevy:
This is an implementation of Provenance Tracking (https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.361.7132&rep=rep1&type=pdf) in Pytorch. The main idea is that provenance of the output tensor is the union of provenances of input tensors.

Tests:

- [x] unary and binary ops
- [x] simple indexing and advanced indexing with Vindex


ordabayevy:
@eb8680 is this along the lines what you were suggesting? Is this new tensor type is supposed to be wrapped by `funsor.Tensor`?

funsor/torch/metadata.py
```diff
@@ -0,0 +1,24 @@
...
+        self._metadata = metadata
+
+    def __repr__(self):
+        return "Metadata:\n{}\n\ndata:\n{}".format(self._metadata, self._t)
+
+    def __torch_function__(self, func, types, args=(), kwargs=None):
+        if kwargs is None:
+            kwargs = {}
+        meta = frozenset().union(
+            *tuple(a._metadata for a in args if hasattr(a, "_metadata"))
```
ordabayevy, line 20: Provenance of the output tensor is the union of provenances of input tensors.

funsor/constant.py
```diff
@@ -0,0 +1,128 @@
...
+    """
+
+    def __call__(cls, const_inputs, arg):
+        if isinstance(const_inputs, dict):
+            const_inputs = tuple(const_inputs.items())
+
+        return super(ConstantMeta, cls).__call__(const_inputs, arg)
+
+
+class Constant(Funsor, metaclass=ConstantMeta):
```
eb8680, line 36: Interesting! It would probably be easiest for us to go over this PR and https://github.com/pyro-ppl/pyro/pull/2893 over Zoom, but one thing that would help me beforehand is if you could add a docstring here explaining how `Constant` behaves differently from `Delta` wrt `Reduce/Contraction/Integrate`

funsor/torch/provenance.py
```diff
@@ -0,0 +1,48 @@
...
+        if isinstance(data, ProvenanceTensor):
+            provenance |= data._provenance
+            data = data._t
+        self._t = data
+        self._provenance = provenance
+
+    def __repr__(self):
+        return "Provenance:\n{}\nTensor:\n{}".format(self._provenance, self._t)
+
+    def __torch_function__(self, func, types, args=(), kwargs=None):
```
eb8680, line 26: @ordabayevy now that you've had a chance to play around with `__torch_function__`, I'm curious about whether you think we should add a `Funsor.__torch_function__` method and attempt to use it in Pyro. I opened #546 to discuss.

funsor/torch/provenance.py
```diff
@@ -0,0 +1,65 @@
...
+        if kwargs is None:
+            kwargs = {}
+        # collect provenance information from args
+        provenance = frozenset()
+        # extract ProvenanceTensor._t data from args
+        _args = []
+        for arg in args:
```
eb8680, line 39: This logic is a bit convoluted. Maybe it could be simplified with some of the helpers in [`torch.overrides`](https://pytorch.org/docs/stable/torch.overrides.html)?

eb8680 (approved):
Implementation seems reasonable, and nicely separated from the rest of the code.

funsor/torch/provenance.py
ordabayevy: > the helpers in torch.overrides?

That might be  useful, I look more into `torch.overrides` functionality.

PR #544, opened 2021-07-21
Fix nan gradients in TraceEnum_ELBO when evaluated eagerly

ordabayevy:
Fixes nan gradients in `TraceEnum_ELBO` when evaluated eagerly (mentioned in https://github.com/pyro-ppl/funsor/issues/493).

The problem seems to be that `reflect.interpret` in line 62 leads to substitutions that have `adjoint` as a `base_interpretation` which in turn leads to some extra expressions being added to the `tape`. Here I fix it by changing interpretation to `_old_interpretation`.

This also solves `test_adjoint.py::test_sequential_sum_product_adjoint`:  `xfail_param(MarkovProduct, reason="mysteriously doubles adjoint values?")` (blocked by #545 ).

PR #545, opened 2021-07-30
Avoid trivial subs

ordabayevy:
This proposes to avoid any trivial `subs` (do not call `interpret` if all `subs` are trivial).

Trivial subs can arise for example in eager `MarkovProduct` with `step_names = {"prev": "prev", "curr": "curr"}`:

```py
def eager_markov_product(sum_op, prod_op, trans, time, step, step_names):
    ...
    return Subs(result, step_names)
``` 

which can pollute `AdjointTape.tape` under `adjoint` interpretation (#493 #544).

### DIFF
--- a/funsor/terms.py
+++ b/funsor/terms.py
@@ -873,8 +873,12 @@ class SubsMeta(FunsorMeta):
 
     def __call__(cls, arg, subs):
         subs = tuple(
-            (k, to_funsor(v, arg.inputs[k])) for k, v in subs if k in arg.inputs
+            (k, to_funsor(v, arg.inputs[k]))
+            for k, v in subs
+            if k in arg.inputs and k is not v
         )
+        if not subs:
+            return arg
         return super().__call__(arg, subs)
 
 


eb8680:
I don't think `*Meta.__call__` methods are the right place for this kind of simplification logic (these methods should do nothing except fill default values for optional arguments and possibly call `to_funsor` where appropriate), and it really shouldn't be necessary regardless - it sounds like the issue, as is so often unfortunately the case, is with the logic in `funsor.adjoint`, or perhaps with alpha-renaming. Can you distill the underlying issue into a failing test comparing a correct expression and a very simple incorrect one generated by `adjoint`?

Pragmatically, however, I'm OK with merging this if you can do that and explain how this fix would unblock what you're actually working on.

funsor/terms.py
```diff
@@ -873,8 +873,12 @@ class SubsMeta(FunsorMeta):
 
     def __call__(cls, arg, subs):
         subs = tuple(
-            (k, to_funsor(v, arg.inputs[k])) for k, v in subs if k in arg.inputs
+            (k, to_funsor(v, arg.inputs[k]))
+            for k, v in subs
+            if k in arg.inputs and k is not v
```
eb8680, line 878: What if `v` is already a `Variable`?

ordabayevy:
> Can you distill the underlying issue into a failing test comparing a correct expression and a very simple incorrect one generated by adjoint?

I discovered this issue by examining `test_adjoint.py::test_sequential_sum_product_adjoint: xfail_param(MarkovProduct, reason="mysteriously doubles adjoint values?")`.

In that test under `adjoint` interpretation `eager_markov_product` calls `sequential_sum_product` (no issues there) and then calls `Subs(results, step_names)` where `step_names = {"prev": "prev", "curr": "curr"}`. Since the names in `step_names` are the identical, `Subs(result, step_names)` returns the same `result`, however, it also gets appended to the `AdjointTape.tape` and I think that doubles adjoint values.

Simple solution can be to check `step_names` and call `Subs(result, step_names)` only if names are not identical. I moved that logic to `SubsMeta.__call__` thinking that it might help to avoid similar issues in the future.

```py
def eager_markov_product(sum_op, prod_op, trans, time, step, step_names):
    if step:
        result = sequential_sum_product(sum_op, prod_op, trans, time, dict(step))
    ...
    return Subs(result, step_names)
```

ordabayevy:
> Can you distill the underlying issue into a failing test comparing a correct expression and a very simple incorrect one generated by adjoint?

Yes, I believe this can be boiled down to much simpler failing test with just couple of lines of code than `test_adjoint.py::test_sequential_sum_product_adjoint: xfail_param(MarkovProduct, reason="mysteriously doubles adjoint values?")`. Should I add such a test?

test/test_adjoint.py
```diff
@@ -259,3 +259,19 @@ def test_sequential_sum_product_adjoint(
...
+@pytest.mark.parametrize(
+    "use_subs", [False, xfail_param(True, reason="doubled adjoint value")]
+)
+def test_subs_adjoint(use_subs):
+    x = random_tensor(OrderedDict(i=Bint[3]))
+
+    with AdjointTape() as tape:
+        y = 2 * x
+        if use_subs:
+            y = y(i="i")
```
ordabayevy, line 273: This substitution shouldn't change adjoint value.

test/test_adjoint.py
```diff
@@ -259,3 +259,32 @@ def test_sequential_sum_product_adjoint(
...
+    with AdjointTape() as tape:
+        y = 2 * x
+        if test == "same":
+            y = y(i="i")
+        elif test == "empty":
+            y = y()
+        elif test == "other":
+            y = y(j=0)
+        elif test == "reduce":
+            y = funsor.terms.Reduce(ops.add, y, frozenset())
```
ordabayevy, line 286: In fact, it seems that any atomic op that doesn't change the funsor doubles the adjoint value.

PR #549, opened 2021-09-17
Add recipes.forward_filter_backward_rsample()

fritzo:
Addresses https://github.com/pyro-ppl/pyro/pull/2929

This implements a multi-sample `forward_filter_backward_rsample()` for use in Pyro's `AutoGaussian` guide.

## Changes

- Modifying the semantics of `MonteCarlo`. These changes preserve semantics of single-sample `MonteCarlo()` but change semantics of `MonteCarlo(particles=Bint[n])` from mean-reducing over `particles` to introducing a new batch dimension over `particles`.
- Changes semantics of `Funsor.sample()` to avoid scaling by numel(sampled_inputs). Correspondingly `.unscaled_sample()` is renamed to `._sample()`.
- Supporting `.reduce(ops.mean, ...)`. This breaks from `.reduce(op, ...)` supporting only associative ops, but this does seem like the cleanest syntax to support `ReductionOp`s over discrete input variables, which will be an important pattern now that the `1/numel` scaling os no longer performed by `.sample()`.
- Adding a new `funsor.recipes` module with high-level algorithms intended for use in both Pyro and NumPyro. The idea so to  maximize test sharing of these recipes by testing all backends in the funsor repo.
- Adding a `batch_vars` arg to `AdjointTape` to support batched backward sample (this might be simplified by #548)
- Factoring out a new `forward_sample()` function from `adjoint()`.

## Tested
- [x] updated existing tests
- [x] add new tests
- [x] ensure this fixes https://github.com/pyro-ppl/pyro/pull/2929

### DIFF
--- a/funsor/terms.py
+++ b/funsor/terms.py
@@ -3,7 +3,6 @@
 
 import functools
 import itertools
-import math
 import numbers
 import typing
 import warnings
@@ -376,12 +375,13 @@ class Funsor(object, metaclass=FunsorMeta):
         Reduce along all or a subset of inputs.
 
         :param op: A reduction operation.
-        :type op: ~funsor.ops.AssociativeOp
+        :type op: ~funsor.ops.AssociativeOp or ~funsor.ops.ReductionOp
         :param reduced_vars: An optional input name or set of names to reduce.
             If unspecified, all inputs will be reduced.
         :type reduced_vars: str, Variable, or set or frozenset thereof.
         """
-        assert isinstance(op, AssociativeOp)
+        assert isinstance(op, (AssociativeOp, ops.ReductionOp))
+
         # Eagerly convert reduced_vars to appropriate things.
         if reduced_vars is None:
             # Empty reduced_vars means "reduce over everything".
@@ -389,6 +389,22 @@ class Funsor(object, metaclass=FunsorMeta):
         else:
             reduced_vars = _convert_reduced_vars(reduced_vars, self.inputs)
         assert isinstance(reduced_vars, frozenset), reduced_vars
+
+        # Attempt to convert ReductionOp to AssociativeOp.
+        if isinstance(op, ops.ReductionOp):
+            if isinstance(op, ops.MeanOp):
+                reduced_vars &= self.input_vars
+                sizes = [v.output.num_elements for v in reduced_vars]
+                scale = 1.0 / reduce(ops.mul, sizes, 1.0)
+                return self.reduce(ops.add, reduced_vars) * scale
+            if isinstance(op, ops.VarOp):
+                diff = self - self.reduce(ops.mean, reduced_vars)
+                return (diff * diff).reduce(ops.mean, reduced_vars)
+            if isinstance(op, ops.StdOp):
+                return self.reduce(ops.var, reduced_vars).sqrt()
+            raise NotImplementedError(f"Unsupported reduction op: {op}")
+        assert isinstance(op, AssociativeOp)
+
         if not reduced_vars:
             return self
         return Reduce(op, self, reduced_vars)
@@ -436,8 +452,7 @@ class Funsor(object, metaclass=FunsorMeta):
             exact = (x.exp() * integrand).reduce(ops.add)
             approx = (y.exp() * integrand).reduce(ops.add)
 
-        If ``sample_inputs`` is provided, this creates a batch of samples
-        scaled samples.
+        If ``sample_inputs`` is provided, this creates a batch of samples.
 
         :param sampled_vars: A set of input variables to sample.
         :type sampled_vars: str, Variable, or set or frozenset thereof.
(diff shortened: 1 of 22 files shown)


eb8680:
All of the relevant downstream tests in Numpyro pass under these changes (on my local machine, at least), but the changes to the behavior of `.sample()` seem to have broken a bunch of downstream tests in `pyro.contrib.funsor`.

Since we don't pin Pyro `dev` to Funsor `master`, one way to proceed would be for me to merge this and then put up a Pyro pull request with fixes that we can merge on the next Funsor or Pyro release. Does that sound reasonable?

funsor/terms.py
```diff
@@ -457,19 +474,12 @@ def sample(self, sampled_vars, sample_inputs=None, rng_key=None):
...
-                if var in result.inputs and var not in self.inputs:
-                    log_scale -= math.log(domain.dtype)
-            if log_scale != 0:
-                result += log_scale
         return result
 
-    def unscaled_sample(self, sampled_vars, sample_inputs, rng_key=None):
+    def _sample(self, sampled_vars, sample_inputs, rng_key=None):
         """
         Internal method to draw an unscaled sample.
```
eb8680, line 484: Update this docstring to clarify the new difference between `sample` and `_sample`?

funsor/terms.py
```diff
@@ -376,19 +375,37 @@ def reduce(self, op, reduced_vars=None):
...
         # Eagerly convert reduced_vars to appropriate things.
         if reduced_vars is None:
             # Empty reduced_vars means "reduce over everything".
             reduced_vars = frozenset(Variable(k, v) for k, v in self.inputs.items())
         else:
             reduced_vars = _convert_reduced_vars(reduced_vars, self.inputs)
         assert isinstance(reduced_vars, frozenset), reduced_vars
+
+        # Attempt to convert ReductionOp to AssociativeOp.
+        if isinstance(op, ops.ReductionOp):
```
eb8680, line 394: Do we want this standardization logic to run under all interpretations?

test/test_recipes.py
```diff
@@ -0,0 +1,331 @@
...
+            diff12 = diff1_ * diff_2
+            moments[name1, name2] = diff12.reduce(ops.mean, reduced_vars)
+
+    return moments
+
+
+def check_ffbr(factors, eliminate, plates, actual_samples, actual_log_prob):
+    """
+    This can be seen as performing naive tensor variable elimination by
+    breaking all plates and creating a single flat joint distribution.
```
eb8680, line 44: If this actually turns out to be true, at some point we should move this logic into `funsor.sum_product` and test it against existing implementations since it's somewhat complex for testing logic and it is used to check correctness of what will presumably end up as one of our most important code paths if/when `AutoGaussian` is finished.

eb8680 (commented):
Looks good per Zoom reviews, just a few nits.

funsor/terms.py
fritzo: yes, I see `.reduce()` as lightweight syntax around actual syntax including `Reduce`. Note it is nonsensical to create a lazy `Reduce` of a `ReductionOp`, the way we have defined `Reduce`.

test/test_recipes.py
fritzo: SGTM, it seems reasonable to eventually factor this out as a `naive_sum_product` or sth. Before we refactor, we should fix the intractable tests below, as that will involve more lambdas etc.

fritzo:
> one way to proceed would be for me to merge this and then put up a Pyro pull request with fixes

I think we can safely make the following sequence of changes:
1. merge this PR
2. fix Pyro dev and pin to a particular Funsor commit
3. ...add AutoGaussian features via many commits to Funsor and Pyro...
4. release Funsor
5. pin Pyro to the Funsor release (required for Pyro releases)
6. release Pyro

eb8680:
That sounds reasonable, provided the fixes necessary in `pyro.contrib.funsor` aren't too onerous - I don't want to block `AutoGaussian` on that if it can be avoided

test/test_approximations.py
```diff
@@ -110,7 +134,9 @@ def test_gaussian_linear(approximate):
         q1 = m1.approximate(ops.logaddexp, guide, "x")
         q2 = m2.approximate(ops.logaddexp, guide, "x")
     actual = q1 + s * q2
-    assert_close(actual, expected)
+
+    if approximate not in (monte_carlo, monte_carlo_10):
```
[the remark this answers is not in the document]
fritzo: oh right, I was going to strengthen these tests once we fixed adjoint, will do...

test/test_approximations.py
```diff
@@ -87,7 +108,9 @@ def test_tensor_linear(approximate):
...
```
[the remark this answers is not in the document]
fritzo: I've just removed these test as I don't understand what they do.

test/test_approximations.py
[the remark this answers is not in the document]
fritzo: removed.

eb8680 (approved):

PR #555, opened 2021-09-25
Add ops.getslice for complex indexing by int,slice,None,Ellipsis

fritzo:
Addresses #556
Blocking #553 

This adds `ops.getslice` to perform fancy slice operations like `x[..., 0, ::2, None, :]`, in particular supporting nontrivial slicing, and supporting unsqueezing via None. These fancier slicing ops will be needed for symbolic Gaussians #556. Note this now distinguishes between `ops.getitem` which supports "advanced indexing" where funsors can appear in the index, from `ops.getslice` which supports only indexing of ground values. Numpy distinguishes semantics of these two types of indexing, and it is much easier to fully implement ground indexing (which I believe this PR does for `ops.getitem`).

## Tested
- [x] unit tests
- [x] pass existing tests

### DIFF
--- a/funsor/terms.py
+++ b/funsor/terms.py
@@ -34,7 +34,7 @@ from funsor.interpretations import (
 )
 from funsor.interpreter import PatternMissingError, interpret
 from funsor.ops import AssociativeOp, GetitemOp, Op
-from funsor.ops.builtin import parse_ellipsis
+from funsor.ops.builtin import normalize_ellipsis, parse_ellipsis
 from funsor.syntax import INFIX_OPERATORS, PREFIX_OPERATORS
 from funsor.typing import GenericTypeMeta, Variadic, deep_type, get_args, get_origin
 from funsor.util import getargspec, lazy_property, pretty, quote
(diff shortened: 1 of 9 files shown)


funsor/terms.py
```diff
@@ -1754,6 +1758,20 @@ def eager_getitem_lambda(op, lhs, rhs):
...
 
+@eager.register(Unary, ops.GetsliceOp, Lambda)
+def eager_getslice_lambda(op, x):
+    index = normalize_ellipsis(op.defaults["index"], len(x.shape))
+    head, tail = index[0], index[1:]
+    expr = x.expr
+    if head != slice(None):
+        expr = expr(**{x.var.name: head})
+        if x.var.name not in expr.inputs:
+            return expr
```
eb8680, line 1769: What if `tail` is not empty here?

eb8680 (commented):
Code looks good, but can you add a couple of tests for `Lambda` with multiple-value indices to exercise `eager_getslice_lambda`?

funsor/terms.py
fritzo: thanks, fixed

fritzo:
ok, I've added some simple tests for multiple value indices

eb8680 (approved):

Issue #560, opened 2021-09-30
`NotImplementedError` when using `apply_optimizer`


ordabayevy:
This term below comes up in `test_pyroapi_funsor`:

```py
from funsor.cnf import Contraction
from funsor.tensor import Tensor
import torch
import funsor.ops as ops
from funsor import Bint, Real
from funsor.terms import Unary, Binary, Variable, Number, lazy, to_data
from funsor.constant import Constant
from funsor.delta import Delta
import funsor
funsor.set_backend("torch")

with lazy:
    term = Contraction(ops.add, ops.mul,
     frozenset({Variable('plate_inner__BOUND_78', Bint[2]), Variable('x__BOUND_77', Real), Variable('plate_outer__BOUND_79', Bint[3])}),  # noqa
     (Unary(ops.exp,
       Contraction(ops.null, ops.add,
        frozenset(),
        (Delta(
          (('x__BOUND_77',
            (Tensor(
              torch.tensor([-1.5227684840130555, 0.38168390241818895, -1.0276085758313882], dtype=torch.float64),  # noqa
              (('plate_outer__BOUND_79',
                Bint[3],),),
              'real'),
             Number(0.0),),),)),
         Constant(
          (('plate_inner__BOUND_78',
            Bint[2],),),
          Tensor(
           torch.tensor([0.0, 0.0, 0.0], dtype=torch.float64),
           (('plate_outer__BOUND_79',
             Bint[3],),),
           'real')),))),
      Unary(ops.all,
       Binary(ops.eq,
        Tensor(
         torch.tensor([-1.5227684840130555, 0.38168390241818895, -1.0276085758313882], dtype=torch.float64),  # noqa
         (('plate_outer__BOUND_79',
           Bint[3],),),
         'real'),
        Tensor(
         torch.tensor([-1.5227684840130555, 0.38168390241818895, -1.0276085758313882], dtype=torch.float64),  # noqa
         (('plate_outer__BOUND_79',
           Bint[3],),),
         'real'))),
      Tensor(
       torch.tensor([[-3.304130052277938, -0.9255234395261538, -1.5122103473560844], [-3.217490519312117, -1.2663745664889694, -1.2900109994655682]],
dtype=torch.float64),  # noqa
       (('plate_inner__BOUND_78',
         Bint[2],),
        ('plate_outer__BOUND_79',
         Bint[3],),),
       'real'),))

x = to_data(funsor.optimizer.apply_optimizer(term))
```

Error message:
```
[...]

PR #562, opened 2021-10-04
`find_domain` for `astype`

ordabayevy:
Addresses #560

### DIFF
--- a/test/test_cnf.py
+++ b/test/test_cnf.py
@@ -22,7 +22,7 @@ from funsor.interpretations import eager, normalize, reflect
 from funsor.interpreter import reinterpret
 from funsor.optimizer import apply_optimizer
 from funsor.tensor import Tensor
-from funsor.terms import Number, Variable
+from funsor.terms import Number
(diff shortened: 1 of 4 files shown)


eb8680 (changes requested):
Can you add a test showing this does indeed resolve #560?

test/test_cnf.py
```diff
@@ -130,3 +134,21 @@ def test_eager_contract_tensor_tensor(
...
+    integrand = Constant((("x", Real),), data)
+    with reflect:
+        term = Integrate(log_measure, integrand, "x")
+        result = term.reduce(ops.add, {"i", "j"})
+
+    with eager:
+        expected = reinterpret(result)
+
+    actual = apply_optimizer(result)
+    assert_close(actual, expected)
```
ordabayevy, line 154: This is as much as I could simplify the test that would still fail.

eb8680 (approved):
Looks good

ordabayevy:
Thanks, @eb8680 !

PR #563, opened 2021-10-04
Replace the deprecated index_update

fehiepsi:
`index_update` is deprecated in favor of `src.at(indices).set(desc)`. See [its docs](https://jax.readthedocs.io/en/latest/_autosummary/jax.ops.index_update.html#jax.ops.index_update)


fritzo:
do we need to bump the jax version in this PR?

fehiepsi:
Yes, the minimal version in setup.py does not have this feature yet. So I just bumped the version up a bit.

eb8680 (approved):
Thanks for fixing this!

PR #579, opened 2021-11-03
Support ops with namedtuple output in ProvenanceTensor

fritzo:
This adds support for pytorch ops that return a namedtuple, e.g. `torch.triangular_solve`.

## Tested
- [x] added a regression test


fritzo:
@fehiepsi any idea why the jax tests are flaky? I would have thought that by fixing `np.random.seed(0)` the jax tests would be deterministic 😕

eb8680 (approved):

PR #583, opened 2022-02-23
Update black

fritzo:
1. `pip install --update black && make format`

Issue #559, opened 2021-09-30
Gaussian funsor variable elimination


fritzo:
Addresses https://github.com/pyro-ppl/pyro/pull/2929
See [design doc](https://docs.google.com/document/d/1MVfG0po7M5UB2mNM2In6iKKq_OYNpa_I-FnJncUgXPI)

This issue tracks changes needed to efficiently perform variable elimination in Gaussian graphical models with plates. While `funsor.sum_product.sum_product()` is a partial solution, we'd like to generalize to a complete solution.

## Tasks
- [ ] Handle collider variables where a latent variable outside a plate depends on an upstream latent variable inside a plate, thereby coupling the upstream variables via moralization. Currently such problems cannot even be specified in the plated-einsum DSL.
    **Temporary workaround:** Globally break all plates out of which any arrow leads; equivalent to `.to_event()`.
- [ ] Handle complete bipartite graphs resulting from the RBM motif (`x_i --> y_ij <-- z_j`). Currently `sum_product()` and the TVE algorithm give up in this case with "intractable!".
    **Temporary workaround:** no known workaround
- [ ] Introduce a new Funsor `ConditionalGaussian(info_vec, precision, conditional, inputs)` representing the batched conditional distribution of the rightmost real input variable, conditioned on other real input variables. This could be (i) a new Funsor in addition to `Gaussian`, (ii) a replacement or generalization of `Gaussian`, or (iii) a special case of `Gaussian` where the input `info_vec` and `precision` are structured (requires #556). This may allow cheaper linear algebra.
    **Temporary Workaround:** naively scatter the three parameters `(info_vec, precision, conditional)` into a dense `Gaussian`.  This can be much more computationally expensive.

fritzo:
@eb8680 it looks like `AutoGaussian(pyrocov_model)` runs out of GPU memory in constructing a low-rank  matrices `precision = sqrt @ sqrt.T`. One possible solution is to use a sqrt(precision) representation in funsor's `Gaussian`. I guess the crux is whether we can implement cheap [Gaussian tensordot](https://github.com/pyro-ppl/funsor/issues/238) without materializing intermediate low-rank precision matrices. @fehiepsi already worked out most of the sqrt representation in [Pyro PR #2019](https://github.com/pyro-ppl/pyro/pull/2019), where `ops.add` becomes mere concatenation.

[...]

PR #584, opened 2022-02-24
Extend partial_sum_product() to cases tractable for Gaussians

fritzo:
Addresses #559 
~~Blocked by #583~~
pair coded with @eb8680 

## Tested
- added tests of intractable cases
- added tests for unreduced sum variables in plates
- added tests for batching, i.e. unreduced plate variables


eb8680 (approved):
I think this is OK to merge, since it's backwards compatible and raises conservative errors. Figuring out what should happen to uneliminated variables or plates can be left to future PRs.

PR #588, opened 2022-03-17
Fix provenance tensor to work with PyTorch 1.11

fritzo:
Note that [`__init__` is automatically called after `__new__`](https://stackoverflow.com/questions/674304/why-is-init-always-called-after-new). I've already made this change in Pyro's fork of `ProvenanceTensor` in https://github.com/pyro-ppl/pyro/pull/3045


ordabayevy (approved):
I can't recall now why I added a code on line 18 but as long as tests pass and it works as expected :+1:

ordabayevy:
> Note that __init__ is automatically called after __new__.

I see.

PR #593, opened 2022-04-11
Provenance funsor

ordabayevy:
Provenance tracking implementation in funsor. Here the provenance is the set of `(name, (point, log_density))` tuples of RV samples that were in the history of the computations of the tracked term. Substitution results in the multiplication by the density:

```py
>>> x = Provenance(value, {("x", (point, dice_factor)})
>>> logp_x = Provenance(logp_value, {("x", (point, dice_factor)})
>>> logp_x(x=point)
logp_value * dice_factor.exp()
```

For example, when integrated wrt to a Delta term the cost term gets multiplied by dice factors that were in the history of the cost term:
```py
>>> delta_x = Delta("x", (point, dice_factor))
>>> Integrate(delta_x, logp_x, {"x"})
logp_value * dice_factor.exp()
```

The `Provenance` funsor is intended to be used in https://github.com/pyro-ppl/pyro/pull/2893 as a wrapper for `ProvenanceTensor`.


ordabayevy:
Tests are failing due to the new jax version #594

PR #598, opened 2022-10-25
Include LICENSE.md in sdist tarball

xhochy:
This ensures that the license is shipped in the sdist tarball on PyPI.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,7 @@
+[metadata]
+# This includes the license file(s) in the wheel.
+license_files = LICENSE.md
+
 [flake8]
 max-line-length = 120
 exclude = docs/src, build, dist
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+include LICENSE.md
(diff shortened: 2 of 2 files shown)


jakirkham:
Think we can add [this to `setup.cfg`]( https://github.com/pypa/sampleproject/blob/d4ee05fdc03e848ed6e7065d8fe8e833a3c8c0b2/setup.cfg#L1-L4 )

setup.cfg
```diff
@@ -1,3 +1,7 @@
+[metadata]
+# This includes the license file(s) in the wheel.
+license_files = LICENSE.md
```
jakirkham, line 3: This should also work for `sdist`s

https://github.com/pypa/setuptools/pull/1536

fritzo:
Hi @xhochy thanks for the fix.  Could you merge in master to get the fix from #599 that's causing the CI failure?

xhochy:
> Hi @xhochy thanks for the fix. Could you merge in master to get the fix from #599 that's causing the CI failure? (I'm not sure how to do that as a maintainer from the github web ui)

Done.

fritzo (approved):

PR #599, opened 2022-10-28
Update __torch_function__ idiom

fritzo:
Blocking #598 

This unblocks CI by fixing to the newer PyTorch idiom of declaring `.__torch_function__()` methods as `@classmethod`.


ordabayevy (approved):
LGTM, thanks for the fix!

PR #601, opened 2022-12-17
Add fixes for jax 0.4

fehiepsi:
Currently, JAX releases 0.4, which drops support for 3.8. So we need to bump version in CI to be able to test the latest version. In addition, additional treatment for jax array is needed because after jax 0.4, DeviceArray becomes jax.Array (i.e. jax.numpy.ndarray) and tracer is a subclass of jax Arrray.

Also attempt to fix docs builds which are failing recently

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,7 +67,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.8]
+        python-version: [3.7,3.8]
     env:
       CI: 1
       FUNSOR_BACKEND: jax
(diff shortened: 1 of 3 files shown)


fritzo (approved):
LGTM.  I'm surprised this PR didn't include a `jax>=0.4` version pin. Should we add one?

Feel free to hit the merge button either wat :+1:

fehiepsi:
Hi Fritz, thanks for reviewing. The change works for both jax 0.4 and jax < 0.4. We have two separate jobs for jax backend on CI: python 3.8 with latest jax and python 3.7 with jax < 0.4.

PR #606, opened 2023-02-06
Scale factors across plate dims in `partial_sum_product`

ordabayevy:
One of the features not supported by `TraceEnum_ELBO` is that you cannot subsample a local variable when it depends on a global variable that is enumerated in the model because it requires a common scale:

```py
@config_enumerate
def model(data):
    # Global variables.
    locs = torch.tensor([1., 10.])
    assignment = pyro.sample('assignment', dist.Categorical(torch.ones(2)))
    with pyro.plate('data', len(data), subsample_size=2) as ind:  # cannot subsample here
        # Local variables.
        pyro.sample('obs', dist.Normal(locs[assignment], 1.), obs=data[ind])

def guide(data):
    pass
``` 

This has been asked on the forum as well: https://forum.pyro.ai/t/enumeration-and-subsampling-expected-all-enumerated-sample-sites-to-share-common-poutine-scale/4938

A solution I'm proposing here is to perform plate-wise scaling inside the `partial_sum_product` by passing the `plate_to_scale` dictionary. Then whenever a plate is reduced we can scale the factor:

```py
# inside partial_sum_product
f = f.reduce(prod_op, reduced_plates)
f_scales = [plate_to_scale[plate] for plate in reduced_plates if plate in plate_to_scale]
if f_scales:
    scale = reduce(ops.mul, f_scales)
    f = pow_op(f, scale)
```

- Added tests

### DIFF
--- a/funsor/sum_product.py
+++ b/funsor/sum_product.py
@@ -4,7 +4,7 @@
 import re
 from collections import OrderedDict, defaultdict
 from functools import reduce
-from math import gcd, prod
+from math import gcd
 
 import funsor
 import funsor.ops as ops
@@ -209,7 +209,7 @@ def partial_sum_product(
     eliminate=frozenset(),
     plates=frozenset(),
     pedantic=False,
-    scales={},
+    plate_to_scale={},
 ):
     """
     Performs partial sum-product contraction of a collection of factors.
@@ -223,9 +223,9 @@ def partial_sum_product(
     assert all(isinstance(f, Funsor) for f in factors)
     assert isinstance(eliminate, frozenset)
     assert isinstance(plates, frozenset)
-    assert isinstance(scales, dict)
+    assert isinstance(plate_to_scale, dict)
 
-    if scales:
+    if plate_to_scale:
         if sum_op is ops.logaddexp and prod_op is ops.add:
             pow_op = ops.mul
         elif sum_op is ops.add and prod_op is ops.mul:
@@ -272,7 +272,7 @@ def partial_sum_product(
             remaining_sum_vars = sum_vars.intersection(f.inputs)
             if not remaining_sum_vars:
                 f = f.reduce(prod_op, leaf & eliminate)
-                f_scales = [scales[plate] for plate in leaf & eliminate if plate in scales]
+                f_scales = [plate_to_scale[plate] for plate in leaf & eliminate if plate in plate_to_scale]
                 if f_scales:
                     scale = reduce(ops.mul, f_scales)
                     f = pow_op(f, scale)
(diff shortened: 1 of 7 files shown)


ordabayevy:
@fritzo @eb8680 can you have a look at this? Do changes to `partial_sum_product` look reasonable to you?

ordabayevy:
@eb8680 I opened a PR in NumPyro (https://github.com/pyro-ppl/numpyro/pull/1572) where I tried to expand the math and check whether expectations match for a full- and a mini-batch log-likelihoods. Can you verify that the math is correct?

funsor/sum_product.py
```diff
@@ -217,6 +223,15 @@ def partial_sum_product(
...
     assert isinstance(plates, frozenset)
+    assert isinstance(plate_to_scale, dict)
+
+    if plate_to_scale:
+        if sum_op is ops.logaddexp and prod_op is ops.add:
+            pow_op = ops.mul
+        elif sum_op is ops.add and prod_op is ops.mul:
+            pow_op = ops.pow
+        else:
+            raise ValueError("should not be here!")
```
fritzo, line 234: This should be NotImplementedError

fritzo, line 234: Move this out to a PROD_TO_POWER dict or similar, see ops

funsor/sum_product.py
```diff
@@ -203,7 +203,13 @@ def partial_unroll(factors, eliminate=frozenset(), plate_to_step=dict()):
 
 
 def partial_sum_product(
-    sum_op, prod_op, factors, eliminate=frozenset(), plates=frozenset(), pedantic=False
+    sum_op,
+    prod_op,
```
fritzo, line 207: Add an optional power_op arg

funsor/sum_product.py
```diff
@@ -256,7 +271,16 @@ def partial_sum_product(
...
-                results.append(f.reduce(prod_op, leaf & eliminate))
+                f = f.reduce(prod_op, leaf & eliminate)
+                f_scales = [
+                    plate_to_scale[plate]
+                    for plate in leaf & eliminate
+                    if plate in plate_to_scale
+                ]
+                if f_scales:
+                    scale = reduce(ops.mul, f_scales)
+                    f = pow_op(f, scale)
```
fritzo, line 282: Could we wrap this in an `if plate_to_scale:` guard to improve readability?

funsor/sum_product.py
```diff
@@ -306,6 +330,14 @@ def partial_sum_product(
                 reduced_plates = leaf - new_plates
                 assert reduced_plates.issubset(eliminate)
                 f = f.reduce(prod_op, reduced_plates)
+                f_scales = [
```
fritzo, line 333: ditto: `if plate_to_scale: ...`

funsor/sum_product.py
```diff
@@ -571,15 +603,23 @@ def modified_partial_sum_product(
...
 
 def sum_product(
-    sum_op, prod_op, factors, eliminate=frozenset(), plates=frozenset(), pedantic=False
+    sum_op,
+    prod_op,
+    factors,
+    eliminate=frozenset(),
+    plates=frozenset(),
+    pedantic=False,
+    plate_to_scale={},
```
fritzo, line 612: default to None, maybe comment on datatype

fritzo (approved):

ordabayevy:
@fritzo can you review the changes please? Also I believe the rule for jax tests needs to be changed in the Settings to use python 3.9 instead of 3.8. Jax doesn't support 3.8 anymore.

fritzo (approved):
LGTM, thanks for addressing my comments.

PR #611, opened 2026-04-16
Fix funsor for JAX 0.10 / NumPy 2.x

juanitorduz:
JAX 0.10 removed the deprecated `a`, `a_min`, `a_max` kwargs from `jnp.clip` (NumPy 2.x alignment). Switch every `np.clip` / `jnp.clip` call to positional arguments so the code is robust to any future signature change as well.

Also fix `from jax.config import config` in the test conftest, which no longer exists in JAX 0.10.

See https://github.com/pyro-ppl/numpyro/pull/2173 and https://github.com/jax-ml/jax/releases/tag/jax-v0.10.0

@fehiepsi could yo please take a look 🙏  ?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.11, 3.12, 3.13, 3.14]
     env:
       CI: 1
     steps:
@@ -38,7 +38,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7]
+        python-version: [3.13]
     env:
       CI: 1
       FUNSOR_BACKEND: torch
@@ -67,7 +67,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.9]
+        python-version: [3.13]
     env:
       CI: 1
       FUNSOR_BACKEND: jax
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,7 +84,7 @@ jobs:
         pip install https://github.com/pyro-ppl/pyro-api/archive/master.zip
         pip install jax<0.10
         # Keep track of NumPyro master branch
-        pip install https://github.com/saitcakmak/numpyro/archive/master.zip
+        pip install https://github.com/pyro-ppl/numpyro/archive/master.zip
         pip install .[test]
         pip freeze
     - name: Run test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,7 +82,7 @@ jobs:
         python -m pip install --upgrade pip
         # Keep track of pyro-api master branch
         pip install https://github.com/pyro-ppl/pyro-api/archive/master.zip
-        pip install jax<0.10
+        pip install "jax<0.10"
         # Keep track of NumPyro master branch
         pip install https://github.com/pyro-ppl/numpyro/archive/master.zip
         pip install .[test]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.11, 3.12, 3.13, 3.14]
+        python-version: [3.8, 3.9]
     env:
       CI: 1
     steps:
@@ -38,7 +38,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.13]
+        python-version: [3.9]
     env:
       CI: 1
       FUNSOR_BACKEND: torch
@@ -67,7 +67,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.13]
+        python-version: [3.9]
     env:
       CI: 1
       FUNSOR_BACKEND: jax
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.8, 3.9]
+        python-version: ["3.11"]
     env:
       CI: 1
     steps:
@@ -38,7 +38,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.10]
+        python-version: ["3.11"]
     env:
       CI: 1
       FUNSOR_BACKEND: torch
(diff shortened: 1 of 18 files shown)


funsor/jax/__init__.py
```diff
@@ -3,7 +3,7 @@
 
 import re
 
-from jax.core import Tracer
+from jax.core import Tracer  # singledispatch needs the concrete type; jax.Array isn't a base of Tracer
```
fehiepsi: Pls revert

funsor/jax/ops.py
```diff
@@ -8,7 +8,7 @@
 import jax.random
 import numpy as onp
 from jax import lax
-from jax.core import Tracer
+from jax.core import Tracer  # multipledispatch needs the concrete type; jax.Array isn't a base of Tracer
```
fehiepsi: Pls revert

fehiepsi (approved):
Thanks @juanitorduz !

juanitorduz:
Reverted the comments. It would be good to go @fehiepsi 

Do we need a funsor release for the downstream packages?

funsor/jax/__init__.py
juanitorduz: done!

funsor/jax/ops.py
juanitorduz: done!

juanitorduz:
@fehiepsi this ci has very old Python versions. Shall I bump them in this PR or another one?

fehiepsi:
Could you help update the version in this PR?

juanitorduz:
> a913061

Sure, I just did it: https://github.com/pyro-ppl/funsor/pull/611/commits/a91306190201978838e009cb4b5f1b8ca05cd940 

@fehiepsi can you please approve the workflow to see if it worked :) ?

fehiepsi:
I think torch support is outdated. Could you just simply remove python 3.7 support or so to pass CI?

juanitorduz:
Done in https://github.com/pyro-ppl/funsor/pull/611#:~:text=Verified-,80d70ed,-Merge%20info

@fehiepsi you can push directly to the branch or somehow allow the CI to run when I push, because otherwise the debugging is very slow 😅

.github/workflows/ci.yml
fehiepsi: could you update `python-version: [3.13]` above (maybe to 3.10?) and `torch==1.9.0+cpu torchvision==0.10.0+cpu` (maybe to 2.4.1) to some dedicated version? `pip install torch==2.4.1 torchvision==0.19.1 torchaudio==2.4.1 --index-url https://download.pytorch.org/whl/cpu
`

fehiepsi:
I guess running the tests locally is much faster than relying on CI.

juanitorduz:
"ERROR: Package 'numpyro' requires a different Python: 3.9.25 not in '>=3.11'" 😅

.github/workflows/ci.yml
```diff
@@ -53,7 +53,7 @@ jobs:
         python -m pip install --upgrade pip
         # Keep track of pyro-api master branch
         pip install https://github.com/pyro-ppl/pyro-api/archive/master.zip
-        pip install torch==1.9.0+cpu torchvision==0.10.0+cpu -f https://download.pytorch.org/whl/torch_stable.html
+        pip install torch==1.11.0 torchvision==0.20.0 -f https://download.pytorch.org/whl/torch_stable.html
```
juanitorduz: @fehiepsi this works locally

juanitorduz (commented):
tested locally (installation), for torch.

I will do more experiments locally

juanitorduz:
@fehiepsi I worked out it locally and I think https://github.com/pyro-ppl/funsor/pull/611/commits/d66db6d69b7a66fac43c4e142efa7536806208db solved the problem. 

The issue: we are in a chicken and egg situation: to support jax>=0.10 we need to install numpyro from the branch that solved the jax>=10.0 issues https://github.com/pyro-ppl/numpyro/pull/2173

I am open to suggestions on how to proceed :)

setup.py
```diff
@@ -79,9 +84,7 @@
         "License :: OSI Approved :: Apache Software License",
         "Operating System :: POSIX :: Linux",
         "Operating System :: MacOS :: MacOS X",
-        "Programming Language :: Python :: 3.7",
-        "Programming Language :: Python :: 3.8",
-        "Programming Language :: Python :: 3.9",
-        "Programming Language :: Python :: 3.10",  # for jax but not torch
+        "Programming Language :: Python :: 3.10",
+        "Programming Language :: Python :: 3.11",
```
Zethson, line 88: Why not also support 3.12, 3.13, 3.14?

3.10 and 3.11 are already old and not even supported by spec-o anymore: https://scientific-python.org/specs/spec-0000/

setup.py
```diff
@@ -38,30 +38,35 @@
...
+            "pytest>=7",
+            "pytest-xdist>=3",
             "requests",
             "scipy",
-            "torchvision>=0.12.0",
         ],
         "dev": [
-            "black",
-            "flake8",
+            "black<24",
```
Zethson: Using ruff would be even better but black or flake8 pins like this are off.

setup.py
juanitorduz: agree, but at this point we want to make minimal changes (I could take this one in another PR)

setup.py
juanitorduz: indeed! But again, as we wanna merge this one soon, I will do that later

fehiepsi:
I think you can pin jax version in CI. We can have flexible version in setup.py but a bit strict in CI.

.github/workflows/ci.yml
```diff
@@ -82,8 +82,8 @@ jobs:
         python -m pip install --upgrade pip
         # Keep track of pyro-api master branch
         pip install https://github.com/pyro-ppl/pyro-api/archive/master.zip
-        # Keep track of NumPyro master branch
-        pip install https://github.com/pyro-ppl/numpyro/archive/master.zip
+        # Track numpyro PR #2173 (xla_pmap_p import fix for JAX>=0.10) until it merges into master
+        pip install https://github.com/saitcakmak/numpyro/archive/fix-xla-pmap-p-import.zip
```
fehiepsi: Let's pin jax release instead.

juanitorduz: @fehiepsi done! can you approve the workflows again :)

setup.py
fehiepsi: It's fine to add 3.12 3.13 3.14 i guess.

setup.py
fehiepsi: Would black behavior change between python versions? If not let's just run the format and relax the version here (also for flake8)

juanitorduz: I think this pin was unnecessary. Reverted

juanitorduz:
@fehiepsi  There are some torch tests failing because the dataset from https://d2hg8soec8ck9v.cloudfront.net/datasets/prep_seal_data.csv can not be reached

image

so I commented them out from the Makefile in https://github.com/pyro-ppl/funsor/pull/611/commits/6f4a6113a8b86084b7d443043d52e4f82bae5840 . We can create an issue and try to fix it later. WDYT?

juanitorduz:
Other comment: Tha JAX tests take ages: https://github.com/pyro-ppl/funsor/actions/runs/25123322581/job/73756052232 🫠

juanitorduz:
Thank you @fehiepsi !

PR #615, opened 2026-07-10
Switch packaging and CI to uv

ordabayevy:
## Summary

- Move package metadata and supported tool configuration from `setup.py`/`setup.cfg` to `pyproject.toml`.
- Add uv-managed dependency resolution via `uv.lock` and update local install docs/Makefile to use `uv`.
- Update CI and Read the Docs configuration to install from the new pyproject metadata.

## Test plan

- `uv lock`
- `uv build`
- `uv run --extra test python -m pytest test/test_import.py`
- `uv run --no-sync python -m pytest test/test_import.py`


pyproject.toml
```diff
@@ -0,0 +1,118 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+[build-system]
+requires = ["setuptools>=77", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "funsor"
+version = "0.4.7"
```
fritzo, line 10: Is there a way to add a comment `mirrored in funsor/__init__.py`?

PR #619, opened 2026-07-31
Move test/docs/dev deps to groups and simplify CI installs.

ordabayevy:
## Summary
- Move `test`, `docs`, and `dev` from package extras into PEP 735 `[dependency-groups]`, keeping only `torch`/`jax` as optional extras.
- Point local `make install` and GitHub Actions at `pip install ... --group ...` (requires pip ≥ 25.1).
- Drop CI installs from pyro/numpyro/pyro-api master and special torch/jax pins; resolve backends from the declared extras instead.

## Test plan
- [ ] Default CI job passes with `pip install . --group test`
- [ ] Torch CI job passes with `pip install .[torch] --group test`
- [ ] Jax CI job passes with `pip install .[jax] --group test`
- [ ] Locally: `pip install -e . --group dev` still installs docs + test tooling
- [ ] Confirm ReadTheDocs is unchanged (`docs/requirements.txt` still used)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,9 +25,7 @@ jobs:
       run: |
         sudo apt install -y pandoc
         python -m pip install --upgrade pip
-        # Keep track of pyro-api master branch
-        pip install https://github.com/pyro-ppl/pyro-api/archive/master.zip
-        pip install .[test]
+        pip install . --group test
         pip freeze
     - name: Run test
       run: make test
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -51,12 +49,9 @@ jobs:
...
         python -m pip install --upgrade pip
-        # Keep track of pyro-api master branch
-        pip install https://github.com/pyro-ppl/pyro-api/archive/master.zip
-        pip install torch==2.4.1 torchvision==0.19.1 --index-url https://download.pytorch.org/whl/cpu
-        # Keep track of Pyro master branch
-        pip install https://github.com/pyro-ppl/pyro/archive/master.zip
-        pip install .[test,torch]
+        # TODO: remove once pyro-ppl releases the Uniform arg_constraints fix
+        # (https://github.com/pyro-ppl/pyro/pull/3453).
+        pip install .[torch] --group test "torch<=2.6"
```
fritzo, line 54: oh man we really need to release pyro, it's been over two years 😫

fritzo (commented):
LGTM thanks for maintaining!

fritzo (approved):

PR #620, opened 2026-08-01
Convert Makefile and CI to uv

ordabayevy:
## Summary
- Switch Makefile and GitHub Actions from pip to `uv sync` / `uv run`, without tracking a `uv.lock`.
- Rename the former `test` dependency group to `dev` (uv's default); keep `docs` separate.
- CI uses `astral-sh/setup-uv` (with matrix Python) and `make install`, then `make test`; torch/jax extras are added inexactly from the Makefile.
- Carry temporary `torch<=2.6` / `jax<0.10` constraints under `[tool.uv]`, bump `actions/checkout` to v7, and run `format` as part of `make test`.

## Test plan
- [ ] Default CI: `make install` then `make test`
- [ ] Torch CI: `FUNSOR_BACKEND=torch make test` after `make install`
- [ ] Jax CI: `FUNSOR_BACKEND=jax make test` after `make install`
- [ ] Local: `make install`, `make lint`, `make format`
- [ ] Optional: `make docs` (docs-only sync + `uv run --no-sync`)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,17 +14,14 @@ docs: FORCE
 	$(MAKE) -C docs html SPHINXBUILD="$(UV_RUN) sphinx-build"
 
 lint: FORCE
-	uv sync --inexact
 	$(UV_RUN) ruff check --fix .
 	$(UV_RUN) python scripts/update_headers.py --check
 	$(UV_RUN) python test/test_import.py
 
 license: FORCE
-	uv sync --inexact
 	$(UV_RUN) python scripts/update_headers.py
 
 format: license FORCE
-	uv sync --inexact
 	$(UV_RUN) ruff format .
 
 test: lint FORCE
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -3,62 +3,67 @@
...
 license: FORCE
-	python scripts/update_headers.py
+	uv run python scripts/update_headers.py
 
 format: license FORCE
-	ruff format .
+	uv run ruff format .
 
-test: lint FORCE
+test: FORCE
```
fritzo, line 22: This seems quite intrusive. These tests are useful insofar as they often fail, whereupon users will `pytest ...` an individual test. It seems like extra friction to `uv run pytest ...`. Can instead activate an environment or something so that `python` and `ruff` and `pytest` all refer to a uv environment?

fritzo (commented):
What's the motivation for switching to uv in CI? While I use uv locally, I have so far avoided dependency on uv in CI, aiming for a minimal supply chain. Requiring uv in the Makefile imposes an opinion on contributors. What does it gain us as maintainers?

ordabayevy:
Agree with your concerns, I was also hesitant about opening this PR. Opened one last PR to modernize ReadTheDocs https://github.com/pyro-ppl/funsor/pull/621 and then we can cut a release?

